PR #41, opened 2017-03-24
Adrian

Adrian-St:
Branch passes tests and is ready to merge


app/controllers/exercises_controller.rb
```diff
@@ -16,6 +16,13 @@ def index
   # GET /exercises/1
   # GET /exercises/1.json
   def show
+
+    exercise_relation = ExerciseRelation.find_by(clone_id: @exercise.id)
+    if exercise_relation
+      @exercise_relation = exercise_relation
+      @exercise_relation.origin = exercise_relation.origin
+      @exercise_relation.clone = exercise_relation.clone
```
tstaubitz: Are these two lines really necessary? shouldn't origin and clone already be copied with the object?

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 db:
     image: postgres
     ports:
-     - "5432:5432"
+     - "54320:5432"
```
Adrian-St: on purpose, because port 5432 is used by postgres

app/controllers/exercises_controller.rb
```diff
@@ -25,23 +32,27 @@ def show
...
     @exercise.descriptions << Description.new
+    @form_action
   end
 
+
   def duplicate
-    exercise = Exercise.find(params[:id])
+    exercise_origin = Exercise.find(params[:id])
     @exercise = Exercise.new
-    @exercise.title = exercise.title
```
tstaubitz: yes it's not

PR #51, opened 2017-10-02
As/style

Adrian-St:
Changed Layout and Comments


app/models/comment.rb
```diff
@@ -11,4 +11,22 @@ def self.search(search)
       all
     end
   end
+
+  def created
```
tstaubitz: There is a gem that handles this (more or less correct) also internationalized. It's called timeago or so.
Please use this instead.

app/models/test.rb
```diff
@@ -10,4 +10,37 @@ def content
       return ''
     end
   end
+
+  def name
+    if exercise_file
```
tstaubitz: please rewrite:
name = ''
name = exercise_file.name if exercise_file

app/models/test.rb
```diff
@@ -10,4 +10,37 @@ def content
...
+      return exercise_file.name
+    else
+      return ''
+    end
+  end
+
+  def path
```
tstaubitz: same here

app/models/test.rb
```diff
@@ -10,4 +10,37 @@ def content
...
+    if exercise_file
+      return exercise_file.path
+    else
+      return ''
+    end
+  end
+
+  def file_type_id
+    if exercise_file
```
tstaubitz: and here

app/models/test.rb
```diff
@@ -10,4 +10,37 @@ def content
...
+      return exercise_file.file_type_id
+    else
+      return ''
+    end
+  end
+
+  def file_type
+    if exercise_file
+      return exercise_file.file_type
```
tstaubitz: and here

PR #238, opened 2020-01-21
Ace editor variable highlighting

kkoehn:
Added automatic switching of editor mode based on selected language.
Added `editor_mode` field to FileType (just like CO), we'll have to fill those - or replant the corresponding seeds.
fixes #33 
fixes #235 

To fix the dynamic reloading of the editor-modes I added a gem called `ace-rails-ap`.
Note that assets:precompile seems to take some time now

PR #239, opened 2020-01-22
Upgrade to rails 6

kkoehn:
Upgrade Rails
Have not seen any breaking changes

requires 
ruby_version     = ">= 2.5.0"
rubygems_version = ">= 1.8.11"


MrSerth:
I checked the requirements on our server, we fulfill these ;).
So there should be nothing to do besides triggering a rollout once we merged.

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# Be sure to restart your server when you modify this file.
+#
+# This file contains migration options to ease your Rails 6.0 upgrade.
```
MrSerth: Do we need to continue working on some of these default settings (for future Rails upgrades) or is everything done?

db/schema.rb
```diff
@@ -173,6 +173,7 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.string "file_extension"
+    t.string "editor_mode"
```
MrSerth: We might merge #238 first

config/initializers/new_framework_defaults_6_0.rb
kkoehn: You are right, we shouldn't need this file anymore.

MrSerth: 👍

PR #253, opened 2020-02-01
Add label validation, make category optional

kkoehn:
Labels can be added to exercises through a select2, which creates them, if they don't exist. There is also LabelCategory to sort the further, but it isn't used anywhere right now.

Categories in the seeds are:
```
Languages
Level
Target Group
```
There were problems while dynamically creating Labels, because no LabelCategory was given. This is optional now.
I also added a validation for the name of the label.

I'd also propose to remove LabelCategory completely  if you have no objections


MrSerth:
That's fine for me, let's get rid of the LabelCategories. There is no LabelCategory on production and we don't have a concept to introduce these right now.

MrSerth:
Looks good. What about all the `test` files? I know, we run Rspec, so these are neither run nor used?

kkoehn:
I also stumbled over the old `/test/` folder and got rid of it as well...
It looks like a relic of the past...
Please wait with the merge though, I'm seeing some irregularities when creating labels...

kkoehn:
> Please wait with the merge though, I'm seeing some irregularities when creating labels...
Nvm - it works. I mistook labels for groups.

PR #263, opened 2020-02-11
Add membership validations

kkoehn:
Adds validations to prevent duplicate memberships

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -21,7 +21,7 @@ class Group < ApplicationRecord
     user.in_group?(self, as: 'admin')
   end
 
-  def full_member?(user)
+  def confirmed_member?(user)
     user.in_group?(self) unless user.in_group?(self, as: 'pending')
   end
 
@@ -47,7 +47,7 @@ class Group < ApplicationRecord
     User.in_group(self).as(:admin)
   end
 
-  def full_members
+  def confirmed_members
     User.in_group(self).as(:member) | User.in_group(self).as(:admin)
   end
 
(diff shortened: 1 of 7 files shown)


kkoehn:
You are right. The problem is, that groupify uses a Membership with an empty type to signal general membership, if you have additional roles, you get an additional membership with the corresponding type. I think I got it down in a custom uniqueness validation now.

app/models/group.rb
```diff
@@ -17,31 +17,40 @@ def self.create_with_admin(params, user)
...
+  def full_member?(user)
     user.in_group?(self) unless user.in_group?(self, as: 'pending')
   end
 
+  def member?(user)
+    user.in_group?(self, as: 'member')
+  end
+
   def make_admin(user)
+    users.delete(user)
```
MrSerth: Do we have a way to remove admin permissions from a member?

kkoehn: The there is currently no way to "unadmin" an admin in a group.

app/models/group.rb
```diff
@@ -17,31 +17,40 @@ def self.create_with_admin(params, user)
...
-  def admins
-    User.in_group(self).as(:admin)
-  end
-
   def admin?(user)
     user.in_group?(self, as: 'admin')
   end
 
-  def member?(user)
+  def full_member?(user)
```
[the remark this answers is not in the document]
kkoehn: It's good that we don't have to deal with cache invalidation.

`active_member` suggests that there are `inactive` ones as well. I like `confirmed_member` though.

app/models/group.rb
MrSerth: Alright :)

PR #266, opened 2020-02-15
AccountLink sharing and permissions

kkoehn:
Big points:
- Enable usage of shared AccountLinks
- Authorize users before using an AccountLink
- Add overview for sharings per account_link (in user profile)

Small points:
- use AJAX for the managing of account link sharings (account_link_users)


app/views/groups/show.html.slim
```diff
@@ -17,21 +17,13 @@
...
-                        = link_to remove_account_link_from_member_group_path(@group, user: admin, account_link: account_link), method: 'post', class: 'dropdown-item remove_account_link' do
-                          i.fa.fa-check
-                          = ' '
-                          = account_link.name
-                      - else
-                        = link_to add_account_link_to_member_group_path(@group, user: admin, account_link: account_link), method: 'post', class: 'dropdown-item add_account_link' do
-                          i.fa.fa-close
-                          = ' '
-                          = account_link.name
+                      = render partial: 'share_button', locals: {shared_user: admin, account_link: account_link}
```
MrSerth: I would prefer a more unique name than `share_button`, maybe something like `share_account_link_button`?

kkoehn:
@MrSerth fixed your reviewpoints and some more bugs that snuck in there after testing

Issue #241, opened 2020-01-24
Check database inconsistencies and model validations


MrSerth:
While working on #224, I noticed a few strange things in our database. I propose to perform these two steps (am I missing more?):

- [x] delete / fix invalid data from the database
- [x] ensure we have model validations / and proper dependents settings (such as `dependent: :destroy`) to prevent these errors in the future

I found these relations which might be problematic:

- [x] Account link users: Are these still used? How (through groups?)?
- [x] Carts without a user (empty carts are okay, I assume?)
- [x] Collections without any user
- [x] Description without exercise
- [x] Exercise authors: Are these still used?
- [x] Exercise files without exercise
- [x] (Exercise group accesses?)
- [x] Exercise lables without exercise
- [x] Messages with no sender status / recipient status
- [x] Relations without exercise relations
- [x] Tests without exercise

Some of them might be okay, I was just unsure about these.

kkoehn:
> * Account links: Are these still used? How?
We use AccountLink to save the data that is necessary to connect to other APIs (mainly CO right now)
> * Carts without a user (empty carts are okay, I assume?)
AFAIK one user should have one cart. Carts without a user should not exist. Users without a cart should get one.
> * Collections without any user
Not quite sure, but doesn't seem to make sense

MrSerth:
Thanks for your quick reply, I did not finish my issue yet when sending so that I edited it a few moments later.

MrSerth:
> > * Account links: Are these still used? How?
> 
> We use AccountLink to save the data that is necessary to connect to other APIs (mainly CO right now)

For sure. I intended to ask Account Link users as I though that now every user holds a unique account link.

> 
> > * Carts without a user (empty carts are okay, I assume?)
> 
> AFAIK one user should have one cart. Carts without a user should not exist. Users without a cart should get one.

Does every user has one cart at all times or is that dynamic, so that the cart is only created once I open the page / add something to my cart?

> 
> > * Collections without any user
> 
> Not quite sure, but doesn't seem to make sense

I agree.

kkoehn:
@MrSerth updated my reply

kkoehn:
On the matter of AccountLinkUser: I'm not sure. I have seen that multiple times now and yes I think it should work through groups somehow. Have not tested that yet

MrSerth:
### General reply:

Thanks. Maybe @tstaubitz knows more?
[...]

PR #284, opened 2020-03-16
Reverse ExerciseFile <> Test relation

kkoehn:
Derived from #241 

- Fixes the relation between ExerciseFile and Test
- Small fixes
- Changes to reference validations

@MrSerth please check, that every Test has an exercise_file_id, the migration will fail otherwise:
`
Test.all.select { |t| t.exercise_file_id.nil? }
`
should be empty


MrSerth:
I just checked the database and performed the corrections necessary (there was one test without an `exercise_file_id`).

MrSerth:
The rollout finished without error.

Issue #250, opened 2020-01-30
Check interoperability with other files


MrSerth:
We designed CodeHarbor to work with ProFormA in order to allow easier file exchange. Let's test that with [these scripts](https://github.com/ProFormA/examples/tree/master/proforma_v2.0.x/ostfalia/task), they should follow the current standard and were create with this [Moodle plugin](https://moodle.org/plugins/qtype_proforma).

kkoehn:
The provided files are referencing `xmlns="urn:proforma:v2.0"` the most recent `.xsd` is `v2.0.1` and won't allow any other version.
We should think about a way to handle such a situation in the future, I'm currently not sure what the best approach would be. 

I'll manually edit the files to fix the version and test those with our importer for now.

kkoehn:
List of problems during import:
- No feedbackmessage and testing-framework (we use our own namespace here). It's not mandatory, therefore an easy fix
  - Testingframework is in there, but under its own namespace, will check that.
- In Proforma Tests can have an arbitrary amount of files. We need them to have exactly one file. Not sure how to approach this. The Test structure would have to be reworked and might be incompatible with the structure in CO. - I will just import the files separately for now, but ignore tests without files. 
- We use the `internal-description` of exercise files to store the `role` - this will probably cause problems down the line.

MrSerth:
Sorry for my late reply. I just took the time to check the current status and tried the import myself.

- feedback-message
  - The feedback-message is mandatory when importing to CodeOcean, right? Can we ensure to provide a general error message such as "Your programm does not meet the test requirements. Please check the output for more information"?
- testing-framework
  - You tackled that with the PR as far as I can see? Sure, manual editing is required to get the best results
- number of test files
  - In CodeOcean, I would also import each file separately. The test file referenced for execution would be a `teacher_defined_test` and others a `hidden` `regular_file`. Could we project this to CodeHarbor? I also noticed with the `reverse.zip` example that the test case was not properly imported. Is this due to the two files included?
- empty test files
  - I am fine ignoring those.
- internal-description
  - Yes, I guess, it's not the right place to put our roles. How is the `task-file-type` used?
- Main File
  - Yes, we need to add one manually. Other grades don't require these but CO does.

Additionally, I stumbled accross those issues:

- HTML entities
[...]

PR #288, opened 2020-03-19
Fix interoperability with external files

kkoehn:
depends on #284 

fixes some issues that appeared with #250


MrSerth:
So far, your changes look good. :)

PR #289, opened 2020-03-20
Fix syntax highlighting tests

### LINKED ISSUE #256: Enable code highlighting for all files with same style
Extracted from #238 

It would be nice if all files, including test files, would be displayed with syntax highlighting. As a bonus, I would appreciate the same styling for all the files 😏.

kkoehn:
fixes #256


MrSerth:
The test ran completed successfully on Travis CI, even though the status is not shown here. I'll merge anyway.

PR #293, opened 2020-03-23
Collection features

kkoehn:
- add leave feature for Collections
- remove instant deletion of Collection
- automatically delete Collection when everyone has left (with extra warning)


MrSerth:
Thanks, it's working like a charm.
I also deployed it to production a few minutes ago.

PR #309, opened 2020-04-08
Rails admin show valid in index views

kkoehn:
This adds a `valid` column for all models with validations in RailsAdmin's index pages.

We can now also sort the column, if we want to have different one on the first page (only programmatically) 
The big downside of this is, that the columns are now hardcoded and need to be adjusted, if anything changes on the models.

TODO
- [ ] Make sure before merge, that every Description has a Exercise
- [ ] Add valid to filter


db/migrate/20200404103206_add_dates_to_descriptions.rb
```diff
@@ -0,0 +1,5 @@
+class AddDatesToDescriptions < ActiveRecord::Migration[6.0]
+  def change
+    add_timestamps :descriptions, null: true #TODO better default than null?
```
MrSerth, line 3: Could we use any associated timestamp for migrations, e.g. the timestamp of the exercise?

MrSerth (approved):

kkoehn:
@MrSerth could you make sure, that we don't have any dangling Descriptions (without exercise) in the database?

I originally wanted to add the ability to filter for valid?, but couldn't get it to work, so I'll postpone it for now.

Also: could we ignore codeclimate in this instance? I'm not even sure it's possible to spec this.

MrSerth:
I checked the exercise descriptions and deleted two (test) descriptions without any exercise.

I am also fine ignoring CodeClimate here as these are super-admin only changes.

MrSerth:
That is really helpful! I deployed the newest version (without problems) and went through all entries to fix any invalid ones. Hence, the database should be in a valid state as of now.

PR #368, opened 2020-08-20
Remove user groups

kkoehn:
Remove an unused model


MrSerth (approved):
Good catch! I didn't notice it was unused, but there is no production data and I was unable to find any other references in the code than the ones you want to remove.

PR #371, opened 2020-08-24
Enable new cops by default

kkoehn:
Fix issues introduced by new cops


MrSerth (approved):
Using `$stdout` instead of `STDOUT` is new to me, happy to learn something new 👍

PR #387, opened 2020-09-12
Bump proforma multi schema

kkoehn:
- bump proforma gem version
- adjustments to incorporate the new change in the gem
  - use of validator
  - extra (any) node in test-configuration, which currently only supports unit-tests
  - handle tests with more then one file (first one will be added to the test, rest will be added as regular-file to the exercises, if a unittest tag is supplied, the file in "entry-file" will be taken (falls back to first file))
    - unused files will be "visible: no"
- reference implementations will have the visibility set according to their own setting (it was always "delayed" before)


MrSerth (approved):
Thanks!

I like the handling of tests with multiple files, this is a good way to go in my opinion. 

One thing I noticed is that the "delayed" setting for the reference implementation can no longer be set (just "yes" or "no"). As we don't use this for now, I would say it is okay and better than using "delayed" as default.

PR #422, opened 2020-11-03
add handling for missing filetype

kkoehn:
Adds handling of missing filetype (mainly during proforma import)

### DIFF
--- a/app/models/exercise_file.rb
+++ b/app/models/exercise_file.rb
@@ -26,7 +26,7 @@ class ExerciseFile < ApplicationRecord
 
   def full_file_name=(full_file_name)
     extension = File.extname(full_file_name)
-    file_type = FileType.find_by(file_extension: extension)
+    file_type = file_type_by_extension(extension)
     raise I18n.t('models.exercise_file.errors.no_filetype', extension: extension) if file_type.nil?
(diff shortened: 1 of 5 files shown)


app/models/exercise_file.rb
```diff
@@ -55,4 +55,10 @@ def parse_text_data
     self.content = Paperclip.io_adapters.for(attachment.instance.attachment).read
     self.attachment = nil
   end
+
+  def file_type_by_extension(extension)
+    FileType.find_or_create_by(file_extension: extension) do |filetype|
```
MrSerth, line 60: Should we use `file_type` here as well? Similar to CodeOcean? And should be similar to `spec/factories/file_type.rb`

kkoehn: I used `filetype` to avoid a name collision - the method is in exercise_file which has a `belongs_to :file_type`

kkoehn: In the factory  I tried to stay coherent

MrSerth: Ah, right. Thanks - I didn't see that in the collapsed GitHub view.
I agree with you; we should keep it coherent.

PR #427, opened 2020-11-08
Active storage

kkoehn:
Switch from paperclip to activestorage
Migration of old not necessary
- [x] ImportFileCache - zip_file
- [x] User - avatar
- [x] ExerciseFile - attachment


MrSerth (approved):
I've first taken the time to read the ActiveStorage introduction and then found your code quite logically. It reads much better and cleaner than having Paperclip and you also used some image manipulation features to serve variants to the images. And the tests are easier to read by having files in the repo instead of escaped base64 strings.

Thanks, also for the clean up along the way.

As there are many changes, do we want to do the deployment together?

PR #433, opened 2020-11-16
remove unused helper and small cleanup

kkoehn:


MrSerth (approved):
`@form_action` is the return value and was always `nil`, right? Don't know why ... 🤷

PR #436, opened 2020-11-17
Migrate exercise model to task

kkoehn:


db/migrate/20201106180002_create_model_solutions.rb
```diff
@@ -0,0 +1,12 @@
+class CreateModelSolutions < ActiveRecord::Migration[6.0]
+  def change
+    create_table :model_solutions do |t|
+      t.string "description"
+      t.string "internal_description"
+      t.string "xml_id"
```
MrSerth, line 6: auto generate for new model_solutions?

db/migrate/20201127153334_add_proforma_columns_to_test.rb
```diff
@@ -0,0 +1,12 @@
+class AddProformaColumnsToTest < ActiveRecord::Migration[6.0]
+  def change
+    add_column :tests, :title, :string
```
MrSerth, line 3: remove old test columns

db/migrate/20210219161631_create_task_authors.rb
```diff
@@ -0,0 +1,11 @@
+class CreateTaskAuthors < ActiveRecord::Migration[6.0]
+  def change
```
MrSerth, line 2: delete if model is deleted

app/services/task_file_service/ace_mode_by_filename.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module TaskFileService
+  class AceModeByFilename < ServiceBase
```
MrSerth, line 4: If the JS function can be used, we would prefer that over the Rails service.

MrSerth (approved):

app/views/tasks/_model_solution_form.html.slim
```diff
@@ -6,7 +6,7 @@
   = model_solution.text_field :internal_description, class: 'form-control'
 .field-element
   = model_solution.label :xml_id, t('tasks.tests.xml_id'), class: 'form-label'
-  = model_solution.text_field :xml_id, class: 'form-control'
+  = model_solution.text_field :xml_id, class: 'form-control', value: test.object.xml_id.presence || '__nested_field_for_replace_with_index__'
```
MrSerth, line 9: That is using a [magic string](https://github.com/ncri/nested_form_fields#index-replacement-string), right? Looks good, but wasn't aware of this feature :)

MrSerth (approved):
Pretty nice! 👏  Thanks for all the effort you invest in making the model and whole codebase cleaner. I really appreciate it!

PR #502, opened 2021-03-17
Add Ransack gem and use it to filter and sort tasks

kkoehn:
Adds ransack gem to replace manual search query building

Note: I'm not really satisfied with how I search for own tasks (I use the visibility scope and an optional user argument. The scope should also support private tasks in the future (not implemented for tasks yet)). Ideas are welcome

### DIFF
--- a/app/assets/javascripts/tasks_index.coffee
+++ b/app/assets/javascripts/tasks_index.coffee
@@ -61,31 +61,36 @@ initializeFilter =->
 
 
   $(".change-hidden-field").click ->
+    console.log('asdf')
     $('#option').val(this.id)
-    $(".index").submit()
+    $('#task_search').submit()
   $('#'+$('#option').val()).addClass('selected')
   intializeAdvancedFilter()
 
 intializeAdvancedFilter =->
-  if $('#dropdownWindowActive')
-    if $('#dropdownWindowActive').val() == "true"
-      $('.dropdown-content').show()
+  $dropdownWindowActive = $('#dropdownWindowActive')
+  $dropdownContent = $('.dropdown-content')
+  if $dropdownWindowActive
+    if $dropdownWindowActive.val() == "true"
+      $dropdownContent.show()
 
-  $('#advanced').click ->
-    $('.dropdown-content').toggle()
-    search = $('#search')
-    if $('#dropdownWindowActive').val() == "true"
-      $(search).css("border-bottom-left-radius", "4px")
-      $("#advanced").css("border-bottom-right-radius", "4px")
-      $('#dropdownWindowActive').val(false)
-      $('#drop').removeClass('fa-caret-up').addClass('fa-caret-down')
+  $advanced = $('#advanced')
+  $drop = $('#drop')
+  $advanced.click ->
+    $dropdownContent.toggle()
+    $search = $('#search')
+    if $dropdownWindowActive.val() == "true"
+      $search.css("border-bottom-left-radius", "4px")
+      $advanced.css("border-bottom-right-radius", "4px")
+      $dropdownWindowActive.val(false)
+      $drop.removeClass('fa-caret-up').addClass('fa-caret-down')
     else
-      $(search).css("border-bottom-left-radius","0px")
-      $("#advanced").css("border-bottom-right-radius", "0px")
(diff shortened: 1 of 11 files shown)


app/controllers/tasks_controller.rb
```diff
@@ -61,54 +57,29 @@ def destroy
 
   private
 
+  def set_search
+    search = params[:search]
+    @created_before_days = search[:created_before_days] if search.is_a?(ActionController::Parameters)
+    @option = params[:option]
```
MrSerth, line 63: Reading the code, I was unable to identify the meaning behind the "option". It only got clear when executing the code. What about something like "visibility"? Or "scope"? Both are not ideal but maybe better than just "option"?

app/controllers/tasks_controller.rb
```diff
@@ -61,54 +57,29 @@ def destroy
...
+    @dropdown = params[:dropdownWindowActive]
```
MrSerth, line 64: I was wondering whether we should use some better describing name here as well. What about "advancedFilterActive", "advancedFilterVisible", "advancedFilterUsed", ...?

This would separate the variable naming from the implementation detail.

app/assets/javascripts/tasks_index.coffee
```diff
@@ -56,36 +56,40 @@ addAnimatedSliding =->
...
-    search = $('#search')
-    if $('#dropdownWindowActive').val() == "true"
-      $(search).css("border-bottom-left-radius", "4px")
-      $("#advanced").css("border-bottom-right-radius", "4px")
-      $('#dropdownWindowActive').val(false)
-      $('#drop').removeClass('fa-caret-up').addClass('fa-caret-down')
+  $dropdownWindowActive = $('#dropdownWindowActive')
+  $dropdownContent = $('.dropdown-content')
+  if $dropdownWindowActive
+    if $dropdownWindowActive.val() == "true"
```
MrSerth, line 73: You only moved that, I know. But shouldn't we use a safe comparison operator `===` here (and a few lines below)?

app/models/task.rb
```diff
@@ -22,64 +22,12 @@ class Task < ApplicationRecord
...
-    end
-  }
-  scope :mine, lambda { |user|
-    if user.nil?
-      where(nil)
-    else
-      where(user_id: user.id)
-    end
-  }
+  scope :visibility, ->(visibility, user = nil) { {owner: where(user: user), public: all}.with_indifferent_access[visibility] }
```
MrSerth, line 25: You are probably referring to this line in your comment, aren't you? I was really thinking about that and it also took me quite some time to understand it (there is some cleverness involved!). 

What about the "private" exercises? I remember that we discussed these are misleading (because of the missing differentiation between "private" and "own" tasks). Is this why you are not filtering for these here? And why can we use `all` for public where "private" / "own" exercises would be included as well? I fear, I miss something here (or is this missing and part of another MR?) Sorry for asking these (stupid) questions...

MrSerth (commented):
Wohoo, I like seeing Ransack instead of this custom-crafted search "engine" here 🙂

app/assets/javascripts/tasks_index.coffee
kkoehn: CoffeeScript already seems to handle that. See [1]

> Because the == operator frequently causes undesirable coercion, is intransitive, and has a different meaning than in other languages, CoffeeScript compiles == into ===, and != into !==. In addition, is compiles into ===, and isnt into !==.

[1]: https://coffeescript.org/#expressions

app/models/task.rb
kkoehn: Currently I'm only differentiating between "own" and "all" tasks. There is option yet to make a task private or something similar. Excluding own tasks during the public search makes perfect sense - I will fix that.

app/assets/javascripts/tasks_index.coffee
MrSerth: Oh, thanks for pointing that out. I wasn’t writing CoffeeScript since ages (but rather plain JS) and didn’t know that. Thanks!

MrSerth (approved):
Very nice; thanks for including all the review points!

PR #503, opened 2021-03-20
Refactor exercise export to tasks

kkoehn:
Refactor the export pipeline to work with `tasks` instead of `exercises`
Main part is the conversion from codeharbor Task to proforma-gem Task

strangely the 
```
after_create_commit :extract_text_data
```
in `models/task_file.rb` does not work anymore.
It should extract files that only contain text and save the content in the appropriate `content` field of the file. It seems that at `after_create_commit` the file does not yet exist in the filesystem.
- As a solution we can either keep the functionality and delegate the work to a job. The delay seems to fix the problem (I'm not sure how consistent though and I'm only using the default queuing system).
- Another solution would be to let the user himself decide, if he wants to extract the text-data for the file (in the edit-view)

This should probably be handled in its own PR though.


app/services/proforma_service/convert_task_to_proforma_task.rb
```diff
@@ -0,0 +1,105 @@
...
+          test_type: test.test_type,
+          files: test.files.map { |test_file| task_file test_file },
+          configuration: test_configuration(test, file)
+        )
+      end
+    end
+
+    def test_configuration(_test, file)
+      {
+        'entry-point' => file&.full_file_name
```
MrSerth, line 73: There is no testing framework / version any more as this is not required by ProFormA, I assume? And we don't need it for CodeOcean, right? Probably, that's the case; I just want to be sure :)

app/views/tasks/show.html.slim
```diff
@@ -490,13 +490,13 @@ legend.toggle-next
...
             li
               = link_to t('exercises.show.add_collection'), new_collection_path(current_user)
-  /.button-box
-    .button-description
-      = "Download your Exercise as a zip file or export them directly to an autograder."
-    - if can? :push_external, Exercise
+  - if can? :export, Task
+    .button-box
+      .button-description
+        = "Download your Task as a zip file or export them directly to an autograder."
```
MrSerth, line 496: We might want to move that to a locale file.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -228,4 +228,33 @@
...
+    end
+
+    it 'sets the correct Content-Type header' do
+      get_request
+      expect(response.header['Content-Type']).to eql 'application/zip'
+    end
+
+    it 'sets the correct Content-Disposition header' do
+      get_request
+      expect(response.header['Content-Disposition']).to include "attachment; filename=\"task_#{task.id}.zip\""
```
MrSerth, line 257: great, I wouldn't have thought of this test case.

spec/factories/task_files.rb
```diff
@@ -12,6 +12,13 @@
       fileable { build(:test) }
     end
 
+    trait :exportable do
```
MrSerth, line 15: Why is this trait called "exportable"? It somewhat implies to me that others can't be exported which I don't see at the moment.

spec/factories/task_files.rb
```diff
@@ -12,6 +12,13 @@
...
+      internal_description { 'internal_description' }
+      usage_by_lms { 'display' }
+      used_by_grader { true }
+      visible { 'yes' }
+    end
+
     trait(:with_text_attachment) do
```
MrSerth, line 22: In favor of consistency with the trait above. This also applies with the trait below.
```suggestion
    trait :with_text_attachment do
```

spec/services/proforma_service/export_task_spec.rb
```diff
@@ -165,75 +152,57 @@
       end
     end
 
-    context 'when exercise has multiple files with role reference implementation' do
+    xcontext 'when task has multiple files with role reference implementation' do
       let(:files) { build_list(:codeharbor_solution_file, 2) }
 
       it 'adds test node to tests node' do
         expect(xml.xpath('/task/model-solutions/model-solution')).to have(2).items
```
MrSerth, line 159: Just asking: Why is this not the case anymore? You've put the context on hold, so this requires additional work?

spec/services/proforma_service/export_tasks_spec.rb
```diff
@@ -32,26 +32,26 @@
...
-    let(:imported_exercises) { zip_files.transform_values! { |zip_file| ProformaService::Import.call(zip: zip_file, user: user) } }
+    # let(:imported_tasks) { zip_files.transform_values! { |zip_file| ProformaService::Import.call(zip: zip_file, user: user) } }
 
     it 'creates a zip-file with two files' do
       expect(zip_files.count).to be 2
     end
 
-    it 'creates a zip-file of two importable zip-files' do
-      expect(imported_exercises.values).to all be_an(Exercise)
+    xit 'creates a zip-file of two importable zip-files' do
```
MrSerth, line 41: We are currently not supporting the export of two exercises at the same time (the former cart), I assume?

MrSerth (commented):
This PR turned out to be easier reviewed than I first thought with the number of changes / touched files listed here. I mainly have a few questions; mostly about the changed model.

I am fine with extracting the bug fixing of the `extract_text_data` commit hook to a separate PR. However (without testing it locally), I don't quite understand why the file should not be available within the file system: As the `after_commit` callback is the last one (according to the [Rails docu](https://guides.rubyonrails.org/active_record_callbacks.html#creating-an-object) I would have expected that everything is persisted and ready. Is this related to the introduction of ActiveStorage?

I like both ideas you pointed out (especially the queuing, as this totally makes sense). For the second option, you noted that we can ask the user to decide. What if we always assume yes  (as we did so far) and hence change the `after_commit` hook to something else? Can we "prevent" the file from being attached to ActiveStorage in the first place and work with the temp files?

app/services/proforma_service/convert_task_to_proforma_task.rb
kkoehn: Yes it just seems to be a remnant of the old Codeharbor. But it may make sense to add this again in the future.
I'm currently considering to remove the entry-point as well, because it is a custom field too.
Do we need that for Codeocean? 
I actually think we do, because a test in ProFormA can have multiple files and in that case we wouldn't know which one to execute - what do you think?

spec/factories/task_files.rb
kkoehn: If the attributes in this trait aren't set, the exporter sets default values, because the fields are required, except internal-description where nil would turn to an empty string and therefore isn't equal as well.
It's basically just a helper to prepare the files in a way that i can export them, then reimport them and everything should be the same.
Maybe we can call it `:reimportable`? It feels a bit clunky though

spec/services/proforma_service/export_tasks_spec.rb
kkoehn: The service actually does already support exporting multiple tasks. 
The problem here is, that the specs uses the importer, which is not yet migrated to use `tasks` instead of `exercises` (#505)

spec/factories/task_files.rb
MrSerth: Okay, this makes sense. By providing fixed default values, you can expect those within your tests.
I don't have a strong opinion on the name, choose whatever you prefer. I see pros and cons for both names and would maybe leave it as it is for now.

app/services/proforma_service/convert_task_to_proforma_task.rb
MrSerth: You're right. Within CodeOcean, we require (additional) information to determine which test files can be executed. We further consider all other test files as "regular" files, i.e. non-executable files required for the test run. This could be a configuration file for a linter or some mock class / helper for a test. 

What if we store that information within [the test meta-data](https://github.com/ProFormA/proformaxml/blob/master/Whitepaper.md#the-test-meta-data-element)? According to the whitepaper, this might work?

spec/services/proforma_service/export_tasks_spec.rb
MrSerth: Ah, I see. This is really complex and hard for me to consider everything. If you'd like, we can also have a chat on (this /) the other MR ;).

kkoehn:
> This PR turned out to be easier reviewed than I first thought with the number of changes / touched files listed here. I mainly have a few questions; mostly about the changed model.
> 
> I am fine with extracting the bug fixing of the `extract_text_data` commit hook to a separate PR. However (without testing it locally), I don't quite understand why the file should not be available within the file system: As the `after_commit` callback is the last one (according to the [Rails docu](https://guides.rubyonrails.org/active_record_callbacks.html#creating-an-object) I would have expected that everything is persisted and ready. Is this related to the introduction of ActiveStorage?

Yes this is really strange, since I already made the changes, when I introduced ActiveStorage and am pretty sure everything worked fine. But now it really seems, that the file is persisted after the `after_commit` hook has concluded.
> I like both ideas you pointed out (especially the queuing, as this totally makes sense). For the second option, you noted that we can ask the user to decide. What if we always assume yes (as we did so far) and hence change the `after_commit` hook to something else? Can we "prevent" the file from being attached to ActiveStorage in the first place and work with the temp files?

The problem with the queuing might be that we that we'd have to set up a job handler. As fas as I understood the built in solution of rails should not be used in a production environment. 

As for the "not using ActiveStorage" approach: I'm not sure, that would require more research

MrSerth:
> > I am fine with extracting the bug fixing of the `extract_text_data` commit hook to a separate PR. However (without testing it locally), I don't quite understand why the file should not be available within the file system: As the `after_commit` callback is the last one (according to the [Rails docu](https://guides.rubyonrails.org/active_record_callbacks.html#creating-an-object) I would have expected that everything is persisted and ready. Is this related to the introduction of ActiveStorage?
> 
> Yes this is really strange, since I already made the changes, when I introduced ActiveStorage and am pretty sure everything worked fine. But now it really seems, that the file is persisted after the `after_commit` hook has concluded.

I also remember that and would, therefore, decouple this PR (as a dependency) from the bug fixing or other PR being blocked by this one.

> > I like both ideas you pointed out (especially the queuing, as this totally makes sense). For the second option, you noted that we can ask the user to decide. What if we always assume yes (as we did so far) and hence change the `after_commit` hook to something else? Can we "prevent" the file from being attached to ActiveStorage in the first place and work with the temp files?
> 
> The problem with the queuing might be that we that we'd have to set up a job handler. As fas as I understood the built in solution of rails should not be used in a production environment.
> 
> As for the "not using ActiveStorage" approach: I'm not sure, that would require more research

Puh, right. I've read the same. We can, of course, go with another job backend (such as sidekiq, which I've used previously in another project) but this would also introduce another dependency. What do you think: How blocking is the feature / bug?

app/services/proforma_service/convert_task_to_proforma_task.rb
kkoehn: I just realized that the entry-point parameter wasn't arbitrarily added, but actually part of ProFormA.
It's part of this sub-xsd for unittests: https://github.com/ProFormA/proformaxml/blob/master/proforma-unittest.xsd
I originally saw it in the external files tried to import: https://github.com/ProFormA/examples/tree/master/proforma_v2.0.x/ostfalia/task
Nonetheless, since it's not part of the main proforma.xsd I would ignore that sub-xsd for now and continue with the meta-data implementation, we can add support for those subxsd's at a later point

MrSerth: Hm, I see. Of course, I also didn't remember the sub-xsd (and their implications) any longer. I agree to have a look at those later

PR #505, opened 2021-04-06
Migrate import to task

kkoehn:
Refactor Importservice to work with `Task` instead of `Exercise`

I also refactored (mainly copied code that was deleted in #436) everything necessary to import zip-files through the UI again. (some services and views)


MrSerth (approved):
Yeah, the import is working again! :shipit:

MrSerth (approved):
Thanks for the update! I tested the code locally and especially the two small issues (upload no file / search with enter) are fine now.

There is still the small issue with the `array_equal?` matcher as I posted in the proforma gem MR.

PR #550, opened 2021-11-30
Update setup

tobiasrohloff:
- Update and polish readme, especially the Unix and macOS part
- Specify Ruby version to match production, and Rails for Rubocop


MrSerth (approved):
Thanks for updating and polishing the ReadMe!

PR #551, opened 2021-11-30
Replace migrations with init migration from current schema

tobiasrohloff:
Running the migrations on a clean system fails due to unsupported used features that were removed at some point. Therefore, a new init migration is provided which is generated from the current schema. This allows to remove all existing migrations while still being able to run `rake db:migrate`.

I tested this by removing the current schema file and then running the init migration file. This re-generated an identical schema file as in master.

⚠️ Please ensure that all production / staging / testing systems are already migrated to the latest version before merging this.


MrSerth (approved):
Thanks for cleaning up!

PR #553, opened 2021-12-07
LOM Sample Endpoint

tobiasrohloff:
This CL adds a [LOM](https://en.wikipedia.org/wiki/Learning_object_metadata) ([IEEE 1484.12.3-2020](https://standards.ieee.org/standard/1484_12_3-2020.html) compatible) endpoint for a sample response of task metadata. Therefore, we introduce a bridges API namespace similar to the ones used in the Xikolo project.

The endpoint can be requested under:  
`/bridges/lom/tasks/sample`

Also, we provide the ability to validate the response with an additional parameter:
`/bridges/lom/tasks/sample?validate=true`

This uses the provided XSD schema files for LOM and responds the error messages. This was quite handy for debugging and can be used for tests in case we will build a real feature for this in the future.

### DIFF
--- a/app/controllers/bridges/lom/tasks_controller.rb
+++ b/app/controllers/bridges/lom/tasks_controller.rb
@@ -1,0 +1,208 @@
+# frozen_string_literal: true
+
+module Bridges
+  module Lom
+    # rubocop:disable Metrics/ClassLength
+    class TasksController < ActionController::API
+      OML_SCHEMA_PATH = Rails.root.join('vendor/assets/schemas/lom_1484.12.3-2020/lom.xsd')
+
+      def show
+        builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') { |xml| sample_oml(xml) }
+
+        if params[:validate].present?
+          schema = Nokogiri::XML::Schema(File.read(OML_SCHEMA_PATH))
+          errors = schema.validate(builder.doc)
+
+          return render plain: errors.map(&:message).join("\n") if errors.any?
+        end
+
+        render xml: builder
+      end
+
+      private
+
+      def sample_oml(xml)
+        xml.lom(xmlns: 'http://ltsc.ieee.org/xsd/LOM') do
+          oml_general(xml)
+          oml_lifecycle(xml)
+          oml_meta_metadata(xml)
+          oml_technical(xml)
+          oml_educational(xml)
+          oml_rights(xml)
+          # Other top-level elements: relation, annotation, and classification.
+        end
+      end
+
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/BlockLength
+      # rubocop:disable Metrics/MethodLength
+      def oml_general(xml)
+        xml.general do
+          xml.identifier do
+            xml.catalog 'UUID'
+            xml.entry task.uuid
+          end
+          xml.title do
+            xml.string task.title, language: task_lang
+          end
+          xml.language task_lang
+          xml.description do
+            xml.string task.description, language: task_lang
+          end
+          if task.programming_language&.language.present?
+            xml.keyword do
+              xml.string "programming language: #{task.programming_language.language}", language: 'en'
+            end
+            if task.programming_language&.version.present?
+              xml.keyword do
+                xml.string "programming language version: #{task.programming_language.version}", language: 'en'
+              end
+            end
+          end
+          xml.structure do
+            xml.value 'atomic'
+          end
+          xml.aggregationLevel do
+            xml.value 1
+          end
+        end
+      end
+      # rubocop:enable all
+
+      def oml_lifecycle(xml)
+        xml.lifeCycle do
+          xml.version do
+            xml.string task_version, language: 'en'
+          end
+          xml.status do
+            xml.value 'final'
+          end
+          xml.contribute do
+            xml.role do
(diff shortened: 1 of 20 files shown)


app/controllers/bridges/lom/tasks_controller.rb
```diff
@@ -0,0 +1,206 @@
+# frozen_string_literal: true
+
+module Bridges
+  module Lom
+    # rubocop:disable Metrics/ClassLength
+    class TasksController < ActionController::API
+      OML_SCHEMA_PATH = Rails.root.join('vendor/assets/schemas/lom_1484.12.3-2020/lom.xsd')
```
tobiasrohloff, line 7: We provide these files in a similar way as the ProFormA gem. I'm not sure if this is the best place and if this will work in production. I would just give it a try.

As already mentioned, I would rather put the schema validation in a tests suite in the future, if this is to become a real feature.

app/controllers/bridges/lom/tasks_controller.rb
```diff
@@ -0,0 +1,206 @@
...
+
+      def sample_oml(xml)
+        xml.lom(xmlns: 'http://ltsc.ieee.org/xsd/LOM') do
+          oml_general(xml)
+          oml_lifecycle(xml)
+          oml_meta_metadata(xml)
+          oml_technical(xml)
+          oml_educational(xml)
+          oml_rights(xml)
+          # Other top-level elements: relation, annotation, and classification.
```
tobiasrohloff, line 32: @MrSerth I guess we should discuss the used/unused elements in person. Also, I left a couple of TODO notes in this file. Most of them are future work.

vendor/assets/schemas/lom_1484.12.3-2020/lom.xsd
```diff
@@ -0,0 +1,102 @@
...
+        **                extend/strict.xsd.  The XSD now is reprsentative of      **
+        **                a schema that can be used to validate strictly           **
+        **                conforming LOM XML Instances.                            **
+        *****************************************************************************
+      </xs:documentation>
+   </xs:annotation>
+
+   <!-- Learning Object Metadata -->
+
+   <xs:include schemaLocation="vendor/assets/schemas/lom_1484.12.3-2020/common/anyElement.xsd"/>
```
tobiasrohloff, line 59: I had to adjust these paths in this file. I hope I can revert that when we move this to the test suite and then leave the file untouched.

tobiasrohloff:
@MrSerth That's the current sample response: [sample.xml.zip](https://github.com/openHPI/codeharbor/files/7669595/sample.xml.zip)

app/controllers/bridges/lom/tasks_controller.rb
MrSerth: I agree. Let's try it this way and iterate over the results once it's merged and working.

app/controllers/bridges/lom/tasks_controller.rb
```diff
@@ -0,0 +1,208 @@
...
+
+      def show
+        builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') { |xml| sample_oml(xml) }
+
+        if params[:validate].present?
+          schema = Nokogiri::XML::Schema(File.read(OML_SCHEMA_PATH))
+          errors = schema.validate(builder.doc)
+
+          return render plain: errors.map(&:message).join("\n") if errors.any?
+        end
```
MrSerth, line 17: I really like the possibility to validate the resulting XML against the XSD. Super handy :+1:

app/controllers/bridges/lom/tasks_controller.rb
MrSerth: Fine for me. As discussed:

- *Relation*: We can only add relational information once the task model supports versioning or linking between different tasks. Hence, this is blocked for now.
- *Annotation*: This could be done right now and would be syntactically correct. However, the ratings we intended don't match the semantical XSD.
- *Classification*: The XSD requires the use of a predefined vocabulary. The ProFormA MD 1.0 format uses its own vocabulary which is not valid for strict LOM. In the future, we might use `customElements` instead.

Result: We don't add these top-level elements for now.

app/controllers/bridges/lom/tasks_controller.rb
```diff
@@ -0,0 +1,208 @@
...
+        xml.metaMetadata do
+          xml.metadataSchema 'ProFormA MD 1.0'
+        end
+      end
+
+      def oml_technical(xml)
+        xml.technical do
+          xml.format 'text/xml'
+          xml.location task_url(id: 'sample')
+          xml.location download_task_url(id: 'sample')
```
MrSerth, line 105: I am fine with this even though it somehow feels incorrect to add the `task_url` without specifying the format. For now, I would keep it this way (because both locations should be included) and postpone this to a later stage.

vendor/assets/schemas/lom_1484.12.3-2020/lom.xsd
MrSerth: Fine for me. We're creating a prototype anyway.

MrSerth (approved):
Looks good, thank you! I also like the idea of scoping the API which allows us to iterate quickly on the various components. It worked as expected locally, so let's give it a try and deploy these changes.

PR #556, opened 2021-12-14
Add BIRD XML sample endpoint

tobiasrohloff:
This CL adds a BIRD XML endpoint for a sample response of task metadata, similar to #553. This time without the validation, the XSD should not be added to the public repo. But I did that locally.

The endpoint can be requested under:
`/bridges/bird/tasks`

As already discussed, the XML schema is very limited for our use case. Attached you can find the sample response.

[bird_example.xml.zip](https://github.com/openHPI/codeharbor/files/7709998/bird_example.xml.zip)

### DIFF
--- a/app/controllers/bridges/bird/tasks_controller.rb
+++ b/app/controllers/bridges/bird/tasks_controller.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+module Bridges
+  module Bird
+    class TasksController < ActionController::API
+      def index
+        builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') { |xml| build_bird(xml) }
+        render xml: builder
+      end
+
+      private
+
+      def build_bird(xml)
+        xml.import(
+          xmlns: 'https://www.mathplan.de/moses/xsd/default',
+          'xmlns:xsi': 'http://www.w3.org/2001/XMLSchema-instance'
+        ) do
+          bird_academy(xml)
+          bird_courses(xml)
+        end
+      end
+
+      def bird_academy(xml)
+        xml.list(type: 'bird_academy') do
+          xml.bird_academy do
+            xml.id_ codeharbor_id
+            xml.name 'CodeHarbor: Your repository for auto-gradeable programming exercises'
+            xml.kurzname 'CodeHarbor'
+          end
+        end
+      end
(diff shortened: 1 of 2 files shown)


app/controllers/bridges/bird/tasks_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+              xml.academyId codeharbor_id
+              xml.lectureType 'Programming Exercise'
+            end
+          end
+        end
+      end
+
+      def codeharbor_id
+        root_url
+      end
```
MrSerth, line 48: Smart, I like that!

app/controllers/bridges/bird/tasks_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      def tasks
+        # TODO: Replace with DB query.
```
MrSerth, line 51: Potentially yes. I added a few demo exercises (for my screenshots) before importing all exercises from CodeOcean (still on the list when the proforma gem is updated to the newest version and all models related to the tasks are fixed).

Here's the current list of tasks:

Bildschirmfoto 2021-12-14 um 10 10 02

MrSerth (commented):
Thanks! Very clean implementation for a not-so-clever schema :+1:

app/controllers/bridges/bird/tasks_controller.rb
tobiasrohloff: Serlo uses the URLs as IDs, I took that approach from them. 😉 There is no dedicated field for URLs. 🙃

MrSerth: Great metadata schema if there are no URLs intended 🙈

app/controllers/bridges/bird/tasks_controller.rb
tobiasrohloff: Great, thanks. I still would like to keep the dummy data response, as BIRD has no pagination mechanism so far and this is an index route.

MrSerth: Fine for me. Let's merge and deploy it and ask the BIRD team to verify it.

MrSerth (approved):

PR #582, opened 2022-01-10
Rebuild authentication with Devise gem

tobiasrohloff:
Rebuild authentication with [Devise](https://github.com/heartcombo/devise).


app/controllers/users/registrations_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    # GET /resource/edit
+    # def edit
+    #   super
+    # end
+
+    # PUT /resource
+    def update
+      super do |resource|
+        resource.avatar.purge if params[:user][:avatar].nil? && params[:user][:avatar_present] == 'false'
```
tobiasrohloff, line 26: This piece was taken from the old `UsersController` to delete avatar images.

app/mailers/application_mailer.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
-  default from: 'openhpi-support@hpi.uni-potsdam.de'
+  default from: 'CodeHarbor <openhpi-support@hpi.uni-potsdam.de>'
```
tobiasrohloff, line 4: Bonus: This looks nicer in email clients and makes the origin more clear. Is this overwritten in production?

app/views/users/passwords/edit.html.slim
```diff
@@ -0,0 +1,37 @@
+.header
+  .std-heading
+    i.fa.fa-key style='color: #444444'
+    = ' ' + t('users.change_pwd')
```
tobiasrohloff, line 4: The new forms contain artifacts from the old ones for a unified look and feel. There is potential for refactorings in the views but I skipped that for now.

config/initializers/devise.rb
```diff
@@ -0,0 +1,315 @@
...
+# are not: uncommented lines are intended to protect your configuration from
+# breaking changes in upgrades (i.e., in the event that future versions of
+# Devise change the default values for those options).
+#
+# Use this hook to configure devise mailer, warden hooks and so forth.
+# Many of these configuration options can be set straight in your model.
+
+# rubocop:disable Metrics/BlockLength
+# rubocop:disable Layout/LineLength
+Devise.setup do |config|
```
tobiasrohloff, line 14: Please have a close look at this config. I have tried to choose reasonable options everywhere.

config/initializers/mail.rb
```diff
@@ -3,7 +3,7 @@
 yaml_file = File.read(Rails.root.join('config/action_mailer.yml'))
 options = YAML.safe_load(yaml_file, aliases: true, permitted_classes: [Symbol])[Rails.env]
 
-Rails.configuration.action_mailer.delivery_method = options.delete('delivery_method')
+Rails.configuration.action_mailer.delivery_method = options.delete('delivery_method') if options.key? 'delivery_method'
```
tobiasrohloff, line 6: This broke action mailer in the development env.

config/locales/devise.en.yml
```diff
@@ -0,0 +1,82 @@
...
+      already_authenticated: "You are already signed in."
+      inactive: "Your account is not activated yet."
+      invalid: "Invalid %{authentication_keys} or password."
+      locked: "Your account is locked."
+      last_attempt: "You have one more attempt before your account is locked."
+      not_found_in_database: "Invalid %{authentication_keys} or password."
+      timeout: "Your session expired. Please sign in again to continue."
+      unauthenticated: "You need to sign in or sign up before continuing."
+      unconfirmed: "You have to confirm your email address before continuing."
+    mailer:
```
tobiasrohloff, line 19: These locales are shipped by Devise. I only modified the email locales a bit.

app/mailers/application_mailer.rb
MrSerth: Nice addition, thanks. The value is overwritten in production but still didn't include the name. I changed it there as well. 🙏

Gemfile
```diff
@@ -96,6 +98,8 @@ group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
   gem 'web-console', '>= 3.7.0'
 
+  gem 'letter_opener'
```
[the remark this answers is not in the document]
MrSerth: I like that and tested it earlier today with your WIP branch. It just works :)

spec/requests/users_spec.rb
```diff
@@ -3,60 +3,62 @@
...
         get user_path(user)
         expect(response).to have_http_status(:ok)
       end
     end
 
-    describe 'PATCH /user/:id' do
-      it 'has http 302' do
-        patch user_path(user, user: user_params)
+    describe 'PATCH /users' do
+      xit 'returns http 302' do
```
tobiasrohloff, line 46: These tests have a weird error. I skipped them for now. They are not really useful anyway...

app/controllers/users/registrations_controller.rb
MrSerth: Fine for me, thanks for pointing that out

app/views/users/passwords/edit.html.slim
MrSerth: Yes, these views could get refactored, I acknowledge the need :)

MrSerth: When are these views (`app/views/users/passwords/*`) being used? Probably when resetting the password, but why do we need `edit` and `new`?

config/initializers/devise.rb
```diff
@@ -0,0 +1,315 @@
...
+  #
+  # Limiting the stretches to just one in testing will increase the performance of
+  # your test suite dramatically. However, it is STRONGLY RECOMMENDED to not use
+  # a value less than 10 in other environments. Note that, for bcrypt (the default
+  # algorithm), the cost increases exponentially with the number of stretches (e.g.
+  # a value of 20 is already extremely slow: approx. 60 seconds for 1 calculation).
+  config.stretches = Rails.env.test? ? 1 : 12
+
+  # Set up a pepper to generate the hashed password.
+  # config.pepper = 'db40a35284d7e2223050d6dc245fa80d10656b25cf2f3ced4f1d02293bc876b12bb996232a370e1c9285d228b53e3a23e304ea5df928f39fce8dd4eb31f16477'
```
MrSerth, line 132: It might be a good idea to configure the pepper. If we decide to do so, we can use the Settings gem to configure it

config/initializers/devise.rb
```diff
@@ -0,0 +1,315 @@
...
+  # If true, extends the user's remember period when remembered via cookie.
+  config.extend_remember_period = true
+
+  # Options to be passed to the created cookie. For instance, you can set
+  # secure: true in order to force SSL only cookies.
+  config.rememberable_options = {secure: true}
+
+  # ==> Configuration for :validatable
+  # Range for password length.
+  config.password_length = 8..128
```
MrSerth, line 184: Can we have an open range here (i.e. not to limit the maximum length)?

config/initializers/mail.rb
MrSerth: The whole file might be worth refactoring (not now, but in general). Fine for me

spec/requests/users_spec.rb
MrSerth: Okay, these tests are really weird. They can be fixed with a few lines:

```ruby
    let(:user_params) { attributes_for(:user, current_password: user.password) }


    describe 'PATCH /users' do
      it 'returns http 302' do
        patch user_registration_path(user, user: user_params, format: :html)
        expect(response).to have_http_status(:found)
      end
    end

    describe 'PUT /users' do
      it 'returns http 302' do
        put user_registration_path(user, user: user_params, format: :html)
        expect(response).to have_http_status(:found)
      end
    end

    describe 'DELETE /users' do
      it 'returns http 302' do
        delete user_registration_path(user, format: :html)
        expect(response).to have_http_status(:found)
      end
    end
```

spec/requests/users_spec.rb
```diff
@@ -3,60 +3,62 @@
...
+      it 'returns http 200' do
+        post user_session_path, params: {email: user.email, password: user.password}
+        expect(response).to have_http_status(:ok)
+      end
+    end
+  end
+
   context 'when logged in' do
-    let(:user) { create(:user) }
     let(:user_params) { attributes_for(:user) }
```
MrSerth, line 18:
```suggestion
    let(:user_params) { attributes_for(:user, current_password: user.password) }
```

MrSerth (commented):
Many thanks for your work here! I appreciate your support and probably cannot imagine the many 'OMG' moments you had. Especially the specs look ... fun

app/views/users/passwords/edit.html.slim
tobiasrohloff: One is the form to request a password reset, the other to actually reset it after clicking the link in the email.

app/controllers/application_controller.rb
```diff
@@ -8,19 +8,12 @@ class ApplicationController < ActionController::Base
...
-  end
-
   private
 
   def set_sentry_context
-    return if current_user.blank?
+    return unless user_signed_in?
 
-    Sentry.set_user(id: current_user.id, username: current_user.username)
+    Sentry.set_user(id: current_user.id, username: current_user.email)
```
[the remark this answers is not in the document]
tobiasrohloff: You are right. I will check if the `username` is optional here. Maybe we can omit it. Or do you have another idea?

app/views/users/passwords/edit.html.slim
MrSerth: Ah, thanks. `new` asks for the mail address (to receive a reset token via mail) and `edit` allows the user to change the password with a reset token

app/mailers/application_mailer.rb
tobiasrohloff: Nice, thx. :)

spec/requests/users_spec.rb
tobiasrohloff: Fixed.

app/controllers/application_controller.rb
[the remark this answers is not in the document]
tobiasrohloff: Yes, [it is optional](https://docs.sentry.io/platforms/ruby/enriching-events/identify-user/). I'll remove it.

spec/requests/users_spec.rb
tobiasrohloff: It also works without here...

MrSerth: Are you sure about that? I think, it only works because you check for status 200. That will, however, include a validation error as the current_password is required to update the user (check the `response.body` during the test run). If an error is included (e.g. the password is missing), no redirect will be returned but the same page with a validation error (status 200). If the user was updated, he / she gets redirected.

tobiasrohloff: Good point, bad test. 😜 I'll check how I can improve this.

config/initializers/devise.rb
tobiasrohloff: Unfortunately, that doesn't work out of the box. :( But I increased the maximum length at least.

Later on, a custom validator could be provided.

MrSerth: Alright, 1000 should be enough 😅

config/initializers/devise.rb
tobiasrohloff: I added a third commit that hopefully does what you meant. :)

MrSerth: Yes, indeed. Thanks! I configured a new pepper in our deploy repo which will overwrite that default in production.

tobiasrohloff: Great! :)

MrSerth (approved):
Thanks again for tackling the Devise integration. I think, there is no thread open any longer so that we might proceed with merging ;)

PR #583, opened 2022-01-10
Integrate SSO & SLO with SAML

### LINKED ISSUE #563: Enable SSO & SLO via SAML
As a user, I want to use a Single-Sign-On via SAML with the BIRD provider in order to authenticate to CodeHarbor. Furthermore, I want to make use of the Single-Log-Out feature to terminate my session.

<hr>

In order to provide SAML support, I would tend to integrate the OmniAuth gem:

- [OmniAuth](https://github.com/omniauth/omniauth)
- [OmniAuth-SAML](https://github.com/omniauth/omniauth-saml)

These two gems can be added directly to the existing Rails application and could be configured manually. So far, CodeHarbor does not use any authentication framework which would require us to do the integration ourselves.

As the existing session management can be improved (and has some bugs, e.g., the broken password reset #524), I would further suggest to take this opportunity and revamp the session management as a whole. I would assume that the existing session management was done with a tutorial similar to [this one](https://www.section.io/engineering-education/how-to-setup-user-authentication-from-scratch-with-rails-6/) somewhere in 2015. Initially, I would have suggested to use [Sorcery](https://github.com/Sorcery/sorcery) (which we have in CodeOcean), but that framework does not yet provide an OmniAuth integration. Hence, my take would be to go with [Devise](https://github.com/heartcombo/devise) that is super common in Rails and is used in mammooc.

Consequently, I see the following roadmap:

1. Replace the current session handling, registration, password reset, ... with [Devise](https://github.com/heartcombo/devise)
2. Add [OmniAuth](https://github.com/omniauth/omniauth) and [OmniAuth-SAML](https://github.com/omniauth/omniauth-saml)

This should provide us with the required SSO and SLO functionality without much code duplication and hopefully improve the maintainability. For the migration with Devise, I would not care much about existing users in our "production" environment. If possible, we keep the data and migrate those (manually) to the updated user table - I can take care of that.

MrSerth:
This PR:
- Adds the required integration to allow the usage of omniauth with Devise
- Adds two SAML omniauth strategies (SAML Test ID and BIRD)
- Enables Single-Sign-On and Single-Log-Out

Closes #563


config/initializers/devise.rb
```diff
@@ -275,6 +275,8 @@
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
   # config.omniauth :github, 'APP_ID', 'APP_SECRET', scope: 'user,public_repo'
+  config.omniauth :bird, strategy_class: OmniAuth::Strategies::Bird if Settings.omniauth.bird.enable
+  config.omniauth :samltestid, strategy_class: OmniAuth::Strategies::SamlTestId if Settings.omniauth.samltestid.enable
```
MrSerth, line 279: Disabling providers does not yet work as expected which results in the rspec failure and the links being displayed if the strategy should be disabled.

config/certs/samltestid_key.pem
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
MrSerth, line 1: This key is unique to development and not shared with any other SAML provider

config/initializers/devise.rb
MrSerth: This issue has been resolved.

lib/omni_auth/strategies/abstract_saml.rb
```diff
@@ -0,0 +1,101 @@
...
+        full_host + script_name
+      end
+
+      def slo_path
+        # This path is defined and handled by the `omniauth-saml` gem
+        "#{full_host}#{script_name}#{request_path}/slo"
+      end
+
+      def self.restore_deprecated_options
+        # The `omniauth-saml` gem still uses some of these deprecated options and does not work correctly otherwise
```
MrSerth, line 82: I created a [PR at omniauth-saml](https://github.com/omniauth/omniauth-saml/pull/206) to remove these lines

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -2,12 +2,29 @@
...
+        session['omniauth_provider'] = provider
+      else
+        # Removing extra as it can overflow some session stores
+        session["devise.#{provider}_data"] = request.env['omniauth.auth'].except(:extra)
+        redirect_to new_user_registration_url
+      end
+    end
+
+    alias bird sso_callback
+    alias samltestid sso_callback
```
MrSerth, line 23: I decided to take an alias and make the method above generic. The documentation "recommends" having a single method for each provider.

lib/omni_auth/strategies/abstract_saml.rb
```diff
@@ -0,0 +1,101 @@
...
+        logout_responses_signed: true, # Enable signature on Logout Response
+        want_assertions_signed: true, # Require the IdP to sign its SAML Assertions
+        want_assertions_encrypted: true # Invalidate SAML messages without an EncryptedAssertion
+      }
+
+      # Our auto-login mechanism passes a desired redirect path (signed with a
+      # secret to prevent tampering) to the request phase.
+      # If we forward this via SAML's standard "RelayState" parameter, we will
+      # get it back in the callback phase.
+      # TODO: This parameter is not yet used in the app
```
MrSerth, line 26: As we don't have any links to content from the start page (and we don't require redirecting the user after login for our prototype), I omitted the concrete implementation. The idea and option is taken from Xikolo.

lib/omni_auth/strategies/abstract_saml.rb
```diff
@@ -0,0 +1,101 @@
...
+      option :idp_sso_target_url_runtime_params, {
+        redirect_path: 'RelayState'
+      }
+
+      # This Lambda is responsible for terminating the session and will
+      # only be used for identity-provider-initiated logout requests
+      option :idp_slo_session_destroy, (proc { |env, _session|
```
MrSerth, line 33: The session destroy is tricky and I am not completely satisfied yet. This mechanism works (as tested) but has limitations due to cookie usage. An ideal solution would use another session store in `config/initializers/session_store.rb`. Any input here is appreciated.

lib/omni_auth/strategies/abstract_saml.rb
```diff
@@ -0,0 +1,109 @@
...
+        # We want to ensure that the new cookie (with an empty session) is set by the client.
+        # Doing so will definitely overwrite an existing session in the browser
+        # Our mechanism will only work through HTTPS and with the following two options
+        env['rack.session'].options[:secure] = true
+        env['rack.session'].options[:same_site] = :none
+      })
+
+      info do
+        {
+          email: @attributes['urn:oid:0.9.2342.19200300.100.1.3'],
```
tobiasrohloff, line 60: Can we provide a reference where these URNs are defined?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -2,12 +2,29 @@
...
 
-    # You should also create an action method in this controller like this:
-    # def twitter
-    # end
+    def sso_callback # rubocop:disable Metrics/AbcSize
+      @user = User.from_omniauth(request.env['omniauth.auth'])
+
+      if @user.persisted?
+        sign_in_and_redirect @user, event: :authentication # this will throw if @user is not activated
+        set_flash_message(:notice, :success, kind: OmniAuth::Utils.camelize(provider)) if is_navigational_format?
```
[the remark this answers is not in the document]
tobiasrohloff: Now I have seen that you've provided a config for this in the individual SAML strategies via `OmniAuth.config.add_camelization`.

lib/omni_auth/strategies/abstract_saml.rb
tobiasrohloff: Let's discuss this in person.

tobiasrohloff: Let's monitor this in production and create a ticket for it.

app/controllers/users/omniauth_callbacks_controller.rb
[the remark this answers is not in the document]
tobiasrohloff: Maybe add a comment about it. :)

lib/omni_auth/strategies/abstract_saml.rb
MrSerth: I created #589 for that

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -2,12 +2,29 @@
...
-    # You should configure your model like this:
-    # devise :omniauthable, omniauth_providers: [:twitter]
+    # SAML doesn't support the CSRF protection
+    protect_from_forgery except: %i[samltestid bird]
 
-    # You should also create an action method in this controller like this:
-    # def twitter
-    # end
+    def sso_callback # rubocop:disable Metrics/AbcSize
+      @user = User.from_omniauth(request.env['omniauth.auth'])
```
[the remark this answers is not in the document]
MrSerth: Devise has many methods that require the usage of instance variables with the scope name: 

```ruby
  def resource
    instance_variable_get(:"@#{resource_name}")
  end
```

As all example usages I was able to find within the documentation of the Devise gem use the instance variable, I decided to keep it here as well.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -2,12 +2,29 @@
...
+
+      if @user.persisted?
+        sign_in_and_redirect @user, event: :authentication # this will throw if @user is not activated
```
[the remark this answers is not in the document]
MrSerth: Well, it is referring to the confirmation. I tested it without assuming confirmed mail addresses through the SSO and noticed that it works as expected (a user gets a mail to confirm the address and won't be signed in). There is no exception being thrown but just a nice flash message informing the user about that extra step

app/models/user.rb
```diff
@@ -55,6 +57,26 @@ def destroy
...
+      # If you are using confirmable and the provider(s) you use validate emails,
+      # uncomment the line below to skip the confirmation emails.
+      new_user.skip_confirmation!
+    end
+
+    # Update some profile information on every login
+    user.email = auth.info.email
+    user.first_name = auth.info.first_name
+    user.last_name = auth.info.last_name
+    user.save
```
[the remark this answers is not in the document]
MrSerth: As expected, using the bang method here does not work directly. If we do so, the call from the `omniauth_callbacks_controller.rb` might receive an exception (which is not handled yet). We could handle it there or keep the current `persisted?` mechanism.

tobiasrohloff (approved):

db/migrate/20220107173125_create_user_identity.rb
```diff
@@ -0,0 +1,12 @@
+class CreateUserIdentity < ActiveRecord::Migration[6.1]
+  def change
+    create_table :user_identities, id: :uuid, force: :cascade do |t|
```
tobiasrohloff, line 3: Bonus: You could add 
```ruby
config.generators do |g|
  g.orm :active_record, primary_key_type: :uuid
end
```
to `config/application.rb` and make UUIDs the default primary key type. :)

MrSerth: Thanks for pointing that out! I created a new PR for that change: #590

PR #585, opened 2022-01-10
add external import actions

kkoehn:
Provide the capability to import exercises from external tools (such as `CodeOcean`) as a task in `Codeharbor`
Since the feature already existed, the code is based on code removed in #436


kkoehn:
This has been tested with the corresponding PR in Codeocean (https://github.com/openHPI/codeocean/pull/939)

app/controllers/tasks_controller.rb
```diff
@@ -138,4 +164,21 @@ def task_params
   def import_confirm_params
     params.permit(:import_id, :subfile_id, :import_type)
   end
+
+  def user_for_api_request
+    authorization_header = request.headers['Authorization']
+    api_key = authorization_header&.split(' ')&.second
+    user_by_api_key(api_key)
+  end
```
MrSerth, line 172: We recently integrated [Devise](https://github.com/heartcombo/devise) with #582. I haven't checked, but it might provide some handy methods to handle the authorisation via API tokens as well.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -344,4 +344,95 @@
       end
     end
   end
+
+  # rubocop:disable RSpec/MultipleExpectations
+  describe '#import_uuid_check' do
+    let!(:task) { create(:task, valid_attributes) }
+    let(:account_link) { create(:account_link, user: user) }
+    let(:uuid) { task.reload.uuid }
+    let(:post_request) { post :import_uuid_check, params: {uuid: uuid} }
```
MrSerth, line 353: This is the test subject, right? We might also consider [naming it accordingly](https://relishapp.com/rspec/rspec-core/docs/subject/explicit-subject)

MrSerth (commented):
Looks good, thanks for proceeding here! I just made two minor comments, you might find them interesting to have a look at these resources. 😊

app/controllers/tasks_controller.rb
kkoehn: That's a good idea, but let's take a look at that at a later date and move this to an issue for now

MrSerth (approved):

MrSerth (approved):
LGTM, thank you!

PR #744, opened 2022-09-24
update models for railsadmin, fix naming

### LINKED ISSUE #542: Remove occurrences of ExerciseFile
There are still some occurrences of `ExerciseFile` in the project rendering some issues. Check the following locations:

- [x] Seed data
- [x] RailsAdmin (see Sentry CODEHARBOR-7).

kkoehn:
fixes #542 

and removes another occurrence of exercise

### DIFF
--- a/app/controllers/account_links_controller.rb
+++ b/app/controllers/account_links_controller.rb
@@ -15,7 +15,7 @@ class AccountLinksController < ApplicationController
 
   def new
     @account_link = AccountLink.new(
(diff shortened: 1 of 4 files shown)


MrSerth (approved):
I verified the changes locally (both for Rails Admin and the seeds) and think you got everything related to `ExerciseFile`. There might be some more occurrences of the term "exercise" (some in CSS classes, others in instance variables or paths) which we might also need to fix. But as there are so many broken things, let's iterate quickly and incorporate these changes now. :1+:

PR #745, opened 2022-09-25
reenable and fix comments

kkoehn:
Reenables comments on index and show of tasks

### DIFF
--- a/app/assets/javascripts/tasks_index.coffee
+++ b/app/assets/javascripts/tasks_index.coffee
@@ -1,17 +1,18 @@
-ready =->
+ready = ->
   initializeSelect2()
   initCollapsable($('.description'), '95px')
   initializeDynamicHideShow()
   initializeFilter()
+  initializeIndexComments()
 
 $(document).on('turbolinks:load', ready)
 
-initializeDynamicHideShow =->
+initializeDynamicHideShow = ->
   $('body').on 'click', '.more-btn-wrapper', (event) ->
     event.preventDefault()
     toggleHideShowMore $(this)
 
-initializeSelect2 =->
+initializeSelect2 = ->
   $('.defaultSelect2').select2
     minimumResultsForSearch: 10
     width: '100%'
@@ -33,7 +34,7 @@ toggleHideShowMore = (element) ->
     $parent.css 'max-height', 'unset'
     # Save old height somewhere for later use
     $text.prop 'default-max-height', $text.css 'max-height'
-    $text.css 'max-height', $text.prop('scrollHeight')+'px'
+    $text.css 'max-height', $text.prop('scrollHeight') + 'px'
   else
     $text.css 'max-height', $text.prop 'default-max-height'
   $less_tag.toggleClass 'hidden'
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/tasks_index.coffee
```diff
@@ -88,21 +89,43 @@ intializeAdvancedFilter =->
...
     else
-      $search.css("border-bottom-left-radius","0px")
+      $search.css("border-bottom-left-radius", "0px")
       $advanced.css("border-bottom-right-radius", "0px")
       $advancedFilterActive.val(true)
       $drop.removeClass('fa-caret-down').addClass('fa-caret-up')
 
   if $('#order_param')
-    #    $('#' + order.value).addClass('active')
+#    $('#' + order.value).addClass('active')
```
MrSerth, line 98: Can / should we remove this line? It doesn't seem to work, currently

app/assets/javascripts/tasks_index.coffee
```diff
@@ -88,21 +89,43 @@ intializeAdvancedFilter =->
...
     $('#order_created').addClass('active')
 
-    #  $('#order_rating').click ->
-    #    $('#order_rating').addClass('active')
-    #    $('#order_created').removeClass('active')
-    #    document.getElementById('order_param').value = 'order_rating'
+  #  $('#order_rating').click ->
+  #    $('#order_rating').addClass('active')
+  #    $('#order_created').removeClass('active')
+  #    document.getElementById('order_param').value = 'order_rating'
```
MrSerth, line 104: Can / should we remove these line? They don't seem to work, currently (I re-added them, of course)

MrSerth (commented):
Great, some more places are working again -- I tested it locally, works fine :)

One additional question: Would you mind having a look at the create and update action of the CommentsController, please? A few days ago, I added a comment about the unused flash messages. Can you confirm that they are not working? Or should we remove them?

app/assets/javascripts/tasks_index.coffee
kkoehn: I plan to add the rating next, so I would leave it there for now

app/assets/javascripts/tasks_index.coffee
kkoehn: see below

app/assets/javascripts/tasks_index.coffee
MrSerth: Alright, fine for me -- thanks for clarifying!

MrSerth:
> > One additional question: Would you mind having a look at the create and update action of the CommentsController, please? A few days ago, I added a comment about the unused flash messages. Can you confirm that they are not working? Or should we remove them?
> 
> In my testing the flash messages actually worked. The problem is, that the flash-messages get displayed at the top of the page and you may not see them, because you are scrolled down. A fix would be to refactor the flash system....

Oh wow, thanks for making me aware! I really wondered how that would work (and already had a look) but could find out how (and as the comments didn't work; I couldn't test it). Now, I repeated my investigation and identified that the flash system uses custom headers to transport the messages 😩

https://github.com/openHPI/codeharbor/blob/9a95415ba68184a044f1389f861dd83038839626/app/controllers/application_controller.rb#L19-L26

https://github.com/openHPI/codeharbor/blob/daa46d0aa116f13d5d0c7407030d991d255e4ee6/app/assets/javascripts/flash_message.js.coffee#L7-L10

This resolves my question -- I'll drop my #ToDo note on the master branch directly. Sorry again!

MrSerth (approved):

PR #755, opened 2022-10-12
Fix deprecation warnings

kkoehn:
There were a few deprecation warnings:
- `DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: tasks/_export_actions.html.slim`
- ``Please modify the call from
`fixture_file_upload("files/proforma_import/testfile.zip")` to `fixture_file_upload("proforma_import/testfile.zip")`.
``

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -229,7 +229,8 @@ class TasksController < ApplicationController
   end
 
   def render_export_actions(task:, exported:, error: nil, task_found: nil, update_right: nil)
-    render_to_string(partial: 'export_actions.html.slim',
+    render_to_string(partial: 'export_actions',
+                     formats: [:'html.slim'],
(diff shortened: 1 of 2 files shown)


app/controllers/tasks_controller.rb
```diff
@@ -229,7 +229,8 @@ def tempfile_from_string(string)
   end
 
   def render_export_actions(task:, exported:, error: nil, task_found: nil, update_right: nil)
-    render_to_string(partial: 'export_actions.html.slim',
+    render_to_string(partial: 'export_actions',
+                     formats: [:html],
```
MrSerth, line 233: The array is not mandatory here, is it? I am unsure what I prefer...

Either `formats: :html` or the version you chose `formats: [:html]`

MrSerth (approved):
Thanks for taking care!

PR #756, opened 2022-10-12
Add Metadata to task show

kkoehn:
Adds meta_data (for task and test) in the show view for task (if present)


MrSerth (commented):
Nice, that's really helpful

Two minor questions:
1. Should we make those additional metadata collapsable and hide them by default? Otherwise, it might be too much for teachers. For the "(2) Additional Information", I am thinking of making that collapsed by default, which then could include the meta data as of now without a second, nested collapsable element.
2. Files cannot have metadata, right (only a task and tests)?

MrSerth:
Edit for 2: Files have a ID, that we aren't showing right now, do we? 

This makes it a bit more difficult to identify which file is annotated with with role in the metadata:

Bildschirm­foto 2022-10-26 um 11 07 02

MrSerth (approved):
Nice, thanks for adding the XML ID and collapsing the additional information by default. I successfully tested these changes locally and think the code looks clean, so let's proceed with merging 🏁

PR #758, opened 2022-10-18
refactor ux for file attachment deletion

### LINKED ISSUE #541: Deleting a file attachment does not remove it
It is currently not possible to remove a file attachment from a task file. When removing it with the X button and saving it, the file won't get deleted and reappears.

Bildschirmfoto 2021-11-16 um 10 33 55

kkoehn:
Refactor and rethink the file-upload (and delete) flow
added a new (not persistent) attribute to task_file to mark if the user wants to use the attached file

fixes #541 

- [x] disable (validation) save the task without file (if upload file is selected)
- [x] remove additional column
- [x] delete file in active storage if its unlinked


db/migrate/20221017194346_add_use_attached_file_to_task_file.rb
```diff
@@ -0,0 +1,11 @@
+class AddUseAttachedFileToTaskFile < ActiveRecord::Migration[6.1]
+  def change
+    add_column :task_files, :use_attached_file, :boolean, null: true
+
+    TaskFile.all.each do |file|
+      file.update_attribute(:use_attached_file, file.attachment.attached?)
+    end
+
+    change_column :task_files, :use_attached_file, :boolean, null: false
```
MrSerth, line 9: Columns with `null: false` usually should have a default, I would say. This is probably `false`, as there is no file attached by default?

MrSerth (commented):
I know this PR is still a draft, but I just tested it and can confirm it resolves the issue.

Besides that, I am wondering why we need a new attribute in the database: Shouldn't we always use an attached file if present in ActiveStorage and otherwise use the native file content? Just a first thought, I am sure there are pretty good reasons to do so

MrSerth (approved):
Awsome, file handling now works as expected again! :muscle:

I like that we no longer require the database migration and think this is a more suitable way to solve this issue 🙂

PR #764, opened 2022-10-24
Extract text data from file

### LINKED ISSUE #506: Provide UI to extract_text_data upon request
As discussed in #503, we disabled the `extract_text_data` as it failed for an unknown reason. Within the conversation, we agreed on adding a button so that the user can request to extract the text data if required for further editing. By performing the extraction upon request, we remove the need to add a queue or some other callbacks that would have been required otherwise.

The idea is that users only request the action if required to further edit the content of the text files within CodeHarbor.

kkoehn:
Enable the user to extract the text of uploaded text files to the built in editor

Closes #506


MrSerth:
Right, I remember we talked about this long time ago.. Nice seeing you tackling that one as well! 👍

MrSerth:
Closes #506

PR #790, opened 2022-11-17
add js routes for frontend route helpers

kkoehn:


MrSerth (approved):
Nice, glad to see this move. It's also much smaller than I expected it, the integration seems to work really well. I just added two other occurrences and will deploy it afterwards as well.

PR #795, opened 2022-11-30
add i18n-js and replace hardcoded strings in js-files

### LINKED ISSUE #734: Add i18n-js library for frontend localisation
In CodeOcean, we use the [i18n-js](https://github.com/fnando/i18n-js) library to make localised strings available in the frontend. I was just updating the gem and library to the newest version, see my changes [here](https://github.com/openHPI/codeocean/pull/1315).

We should add i18n-js to CodeHabror as well.

_Originally posted by @MrSerth in https://github.com/openHPI/codeharbor/pull/716#discussion_r959363581_

kkoehn:
I had to use an old version of i18n-js because we still use sprockets

fixes #734

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@
 
 source 'https://rubygems.org'
 
+gem 'i18n-js', '< 4.0.0' # Newer versions require the npm package `i18n-js`
 gem 'js-routes'
 # Create Zip files
 gem 'rubyzip'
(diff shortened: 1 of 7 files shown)


MrSerth (approved):
Thanks for adding I18n-js. I think, this will be really useful for more JavaScript functions. Using the older version is fine, I assume, and we can upgrade to the latest once we switched to another JavaScript pipeline.

Also, I quickly checked the JavaScript files for additional (missing) translations, but couldn't find any as well. Good job!

PR #797, opened 2022-12-01
move rescue_from from individual controllers to application_controller

### LINKED ISSUE #480: move `rescue_from CanCan::AccessDenied` to application controller

kkoehn:
also let the user go back to the index action, when he is logged in - otherwise he gets redirected to root

fixes #480


MrSerth (approved):
Finally! 💪 

It's so great to see these small maintenance improvements, and the corresponding ticket is almost two years old. We're getting there!

PR #805, opened 2022-12-06
update to latest proforma version

kkoehn:
update to proforma gem version 0.8.0 which added support for proforma 2.1

After merge and release of the gem this gemfile.lock has to be updated


MrSerth (approved):
Thanks! I added some additional comments and merged your changes 👍

Issue #807, opened 2022-12-08
XML ID of Task Files not used for export


MrSerth:
When exporting a task with regular task files, we do not use the XML ID for the task files:

https://github.com/openHPI/codeharbor/blob/75a0b1e677d20b07e028de444f14042f38a7a0b1/app/services/proforma_service/convert_task_to_proforma_task.rb#L71-L82

Instead, we just use the database ID of the file:

https://github.com/openHPI/codeharbor/blob/75a0b1e677d20b07e028de444f14042f38a7a0b1/app/services/proforma_service/convert_task_to_proforma_task.rb#L73

This causes issues when providing additional metadata, such as a desired file role, as they are not linked together correctly.

Issue #413, opened 2020-10-12
Small issue compilation


kkoehn:
Functionality
- [ ] Collections can't be deleted manually (the might be deleted, once they have no user attached)
- [ ] You have to verify your mail after registration, but not after a change (of the email address)
- [ ] simple_form is not used consistently
- [ ] A user can message himself (and reply)
- [ ] Accepting a non existing collection errors (You get a message with a link)
- [ ] Own exercises in collections are "locked"
- [ ] Adding an exercise to a group has a strange flow (you have to go to edit and enter it in a select2 there)
  - [ ] This doesn't work

Refactoring
- [ ] Split large controllers, move more logic to services

Styling
- [ ] Form errors are not correctly used in all forms

Technology
- [ ] Paperclip is deprecated, switch to Active Storage
- [ ] Sprockets update?
- [ ] Yarn?
- [ ] Bootstrap 4? (we use twitter-bootstrap-rails, which seems sparely updated)
- [ ] Ransack for Exercise search

kkoehn:
- [ ] remove cart
- [ ] exercise -> add to collection -> add option on create collection on the fly
- [ ] add relation between collection and group
- [ ] request access with message -> author can choose the right level (read only / write access)

kkoehn:
Must have before release:
- [ ] Exercise to task refactor
- [ ] active storage
- [ ] automatic collection creation and assignment on import from CO

MrSerth:
Discovered during Code Review:

- [ ] Select2 gets duplicated when moving back and forth (caused by turbolinks)
- [ ] "Show Related" on `/exercises` renders arrows without content if no exercises are related

kkoehn:
- [ ] uuid and parent_uuid are not changable via edit > update

PR #808, opened 2022-12-08
Update Seeds and fix example

MrSerth:
When testing the export from CodeHarbor to CodeOcean, we noticed some issues:

- The filename `Main.java` didn't match the class name `HelloWorld
- CodeOcewn got confused with two names having the same name
- A Makefile was missing preventing direct code execution in CodeOcean
- The file role for CodeOcean was missing

I fixed the seeds, but we are still facing two issues preventing a direct import and use in CodeOcean:
- https://github.com/openHPI/codeocean/issues/1466
- https://github.com/openHPI/codeharbor/issues/807

Both should be fixed as well. As long as this isn't done, the seeds won't work correctly in CodeOcean.

### DIFF
--- a/db/seeds/development.rb
+++ b/db/seeds/development.rb
@@ -67,7 +67,7 @@ Label.create!(name: 'Conditions', color: 'DF0101')
 Label.create!(name: 'Data Structures', color: '3333CC')
 
 pl_java = ProgrammingLanguage.create!(language: 'Java', version: '17')
-ProgrammingLanguage.create!(language: 'Python', version: '3')
+pl_python = ProgrammingLanguage.create!(language: 'Python', version: '3.8')
 
 TestingFramework.create!(name: 'JUnit', version: '5')
 TestingFramework.create!(name: 'Pytest', version: '6')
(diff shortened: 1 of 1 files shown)


MrSerth:
> Looks good. I was wondering if we may add another more basic task to the seeds, that doesn't come with the overhead needed in Codeocean

Thanks for having a look! I just added another small basci task with 51199c1d, but that still requires some metadata to be included in CodeOcean right away. I would prefer to have those metadata in place, so that new developers can start right away and have an example that would work in CodeOcean.

> The "same path for multiple files"-problem is already noted in #413

Sounds good, thanks for that as well! The main issue in CodeOcean was related to both files being visible and thus behaving unexpectedly. I fixed the specific occurrance with https://github.com/openHPI/codeocean/commit/838019343e6b1920d29bdcf03a004fcf9114f835 as well.

MrSerth:
I rebased the current branch and updated the seed files to always include an XML ID (as required by #814).

Waiting for review, especially from @zuhrasofyan and @Melhaya (as we discovered the issues together during our testing).

MrSerth:
Both issues preventing to merge this PR have been resolved. I am proceeding with merging now, besides the two reviews missing.

PR #814, opened 2022-12-12
Fix unused xml_id for files

### LINKED ISSUE #807: XML ID of Task Files not used for export

kkoehn:
xml_id was added to task_files
xml_id is now required for files (the same as tests and model_solutions already) and will be automatically filled out in the form with a default value

fixes #807 
helps #808

### DIFF
--- a/app/models/task_file.rb
+++ b/app/models/task_file.rb
@@ -7,7 +7,7 @@ class TaskFile < ApplicationRecord
   validates :name, presence: true
   validates :attachment, presence: true, if: -> { use_attached_file == 'true' }, on: :force_validations
   validates :xml_id, presence: true
-  validates :xml_id, uniqueness: {scope: %i[fileable_id fileable_type]}
+  validate :unique_xml_id, if: -> { !fileable.nil? }
 
   attr_accessor :use_attached_file, :file_marked_for_deletion
 
@@ -46,5 +46,11 @@ class TaskFile < ApplicationRecord
 
   def remove_attachment
     attachment.purge if use_attached_file != 'true' && attachment.present?
+  end
+
+  def unique_xml_id
+    task = fileable.is_a?(Task) ? fileable : fileable.task
+    xml_ids = (task.all_files - [self]).map(&:xml_id)
+    errors.add(:xml, I18n.t('task_files.xml_id_not_unique')) if xml_ids.include? xml_id
   end
 end
--- a/app/models/task_file.rb
+++ b/app/models/task_file.rb
@@ -6,6 +6,8 @@ class TaskFile < ApplicationRecord
   has_one_attached :attachment
   validates :name, presence: true
   validates :attachment, presence: true, if: -> { use_attached_file == 'true' }, on: :force_validations
+  validates :xml_id, presence: true
+  validates :xml_id, uniqueness: {scope: [:fileable_id, :fileable_type]}
 
   attr_accessor :use_attached_file, :file_marked_for_deletion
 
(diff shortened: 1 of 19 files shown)


app/models/task_file.rb
```diff
@@ -6,6 +6,8 @@ class TaskFile < ApplicationRecord
   has_one_attached :attachment
   validates :name, presence: true
   validates :attachment, presence: true, if: -> { use_attached_file == 'true' }, on: :force_validations
+  validates :xml_id, presence: true
+  validates :xml_id, uniqueness: {scope: %i[fileable_id fileable_type]}
```
MrSerth, line 10: At least in my tests, this validation doesn't work as expected. I was able to add two new files with the same XML ID and save those. Only when editing, this validation kicks in and prevents saving the task. Do we skip the validation for new task_files for some reason?

Bildschirm­foto 2022-12-17 um 10 29 32
Bildschirm­foto 2022-12-17 um 10 30 25

app/models/task_file.rb
```diff
@@ -6,6 +6,8 @@ class TaskFile < ApplicationRecord
   has_one_attached :attachment
   validates :name, presence: true
   validates :attachment, presence: true, if: -> { use_attached_file == 'true' }, on: :force_validations
+  validates :xml_id, presence: true
```
MrSerth, line 9: If we force the XML ID to be present, shouldn't we always display it? And what about (a migration for) existing data? We could clear the test data, but probably should consider this as well.

MrSerth: Indeed, existing exercises break:

Bildschirm­foto 2022-12-17 um 10 38 36

```
ActiveRecord::RecordInvalid - Validation failed: Files xml can't be blank, Tests files xml can't be blank, Model solutions files xml can't be blank
```

app/models/task_file.rb
MrSerth: Additionally, the scope is probably not sufficient yet. The following entry is valid according to the validation:

Bildschirm­foto 2022-12-17 um 10 45 20

but generates an invalid XML:

Bildschirm­foto 2022-12-17 um 10 44 45

```
Proforma::PostGenerateValidationError - [#<Nokogiri::XML::SyntaxError: 36:0: ERROR: Element '{urn:proforma:v2.0.1}file': Duplicate key-sequence ['0'] in key identity-constraint '{urn:proforma:v2.0.1}task-fileids'.>]
```

MrSerth (changes requested):
Thanks for working on the XML ID integration for files! Despite finding some issues with the current validation of user data, I was successfully able to verify your fix regarding the original issue #807. 💪 

Please have another look at the issues I found with this PR; we might also need to change the seed data to either accommodate or auto-generate an XML ID if not created yet.

app/models/task_file.rb
kkoehn: I added a custom validation that should cover this

MrSerth (approved):
Thanks! I tested various scenarios on my system and think this is ready to go now 🏁 💪

MrSerth:
I just rebased this branch (to test with the latest Ruby 3.2) and will merge now.

PR #843, opened 2023-01-25
Update show.html.slim

Melhaya:
Fix view in the collection (with Zuhra)

Replaced excerice_path with tasks_path

### DIFF
--- a/app/views/collections/show.html.slim
+++ b/app/views/collections/show.html.slim
@@ -29,14 +29,14 @@
           tr
             td
               - if can? :read, exercise
-                = link_to exercise.title, exercise_path(exercise), class: 'btn btn-default'
+                = link_to exercise.title, task_path(exercise), class: 'btn btn-default'
               - else
-                = link_to exercise.title, exercise_path(exercise), class: 'btn btn-default disabled'
+                = link_to exercise.title, task_path(exercise), class: 'btn btn-default disabled'
             td
               - if cannot? :read, exercise
                 span.fa.fa-lock.lock
                 = ' '
-                = link_to t('collections.show.contribute'), contribute_exercise_path(exercise), method: :post, class: 'btn btn-default'
+                /= link_to t('collections.show.contribute'), contribute_exercise_path(exercise), method: :post, class: 'btn btn-default'
               = link_to remove_exercise_collection_path(@collection, exercise: exercise), method: :patch, class: 'pull-right' do
                 i.fa.fa-remove style=("color: #800000; font-size: 20px;")
 br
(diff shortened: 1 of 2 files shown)


zuhrasofyan (approved):

MrSerth (changes requested):
Thanks for working on these changes and working towards restoring the functionality of collections. :+1: 

There are a few things I would ask you to consider:

- At least in my local testing, your change is not working as intended. The link generated is wrong, referring to `/tasks.1` rather than `/tasks/1`. There is a small typo in the Rails route helper, see more details on the naming [here](https://guides.rubyonrails.org/routing.html#path-and-url-helpers)
  Bildschirm­foto 2023-01-25 um 11 44 25

- The commit message should be improved. Usually, we try to provide a short summary of the change in the first line (ideally, in less than 50 characters, even though I don't stick to that too often). Then, followed by an empty new line, some further description. GitHub does the new line automatically (that's why there are two input fields), but on other clients, you might need to do so manually. [Here](https://cbea.ms/git-commit/) are some more details on writing a commit message.
- If you want to add multiple authors to a commit message, please use the Git default with `Co-authored-by` as [explained here](https://github.blog/2018-01-29-commit-together-with-co-authors/).
- I would be super happy if you could add a suitable test case here. The existing tests failed to recognize the wrong helper, because (for example) the request tests used to test against an empty collection: https://github.com/openHPI/codeharbor/blob/66f114d2c6d2179de99b786b46d3026bca8d9074/spec/requests/collections_spec.rb#L8
  If we could change that to test with a collection containing at least one exercise, we would have noticed the previous error.
- Furthermore, we might want to add a second test case to the [controller test](https://github.com/openHPI/codeharbor/blob/acf10d4a347082a3a25c1816fd75812b6364dff8/spec/controllers/collections_controller_spec.rb#L27-L34) expecting that the `response.body` includes the `task/{:id}` path --- that would have caught the mistake still present in this PR.
- During testing, you might come along with a missing `contribute_exercise_path` that was previously a member action on exercises: 
https://github.com/openHPI/codeharbor/blob/2e08353bbbcf3f33369d602c0f1e6eae8c849d4a/config/routes.rb#L105
  This was intended to allow users to request access to an exercise they would like to contribute to:
  https://github.com/openHPI/codeharbor/blob/2e08353bbbcf3f33369d602c0f1e6eae8c849d4a/app/controllers/exercises_controller.rb#L266-L272
 For now, I would probably leave repairing the `contribute` action to another ticket.

Looking forward to your first contribution 🎉

Melhaya:
this task is left. More research on test cases needed to be able to do it.

> Furthermore, we might want to add a second test case to the [controller test](https://github.com/openHPI/codeharbor/blob/acf10d4a347082a3a25c1816fd75812b6364dff8/spec/controllers/collections_controller_spec.rb#L27-L34) expecting that the response.body includes the task/{:id} path --- that would have caught the mistake still present in this PR.

Melhaya:
This progress has been combined with #844

PR #885, opened 2023-04-21
Fix #881 by rescuing the error and redirecting to the root page.

### LINKED ISSUE #881: ActionController::UrlGenerationError: No route matches {:action=>"index", :controller=>"users", :id=>"56"} (ActionController::UrlGenera...
Sentry Issue: [CODEHARBOR-14](https://codeocean.sentry.io/issues/4095404362/?referrer=github_integration)

```
CanCan::AccessDenied: You are not authorized to access this page. (CanCan::AccessDenied)
  cancancan (3.5.0) lib/cancan/ability.rb:180:in `authorize!'
  cancancan (3.5.0) lib/cancan/controller_additions.rb:339:in `authorize!'
  cancancan (3.5.0) lib/cancan/controller_resource.rb:40:in `authorize_resource'
  cancancan (3.5.0) lib/cancan/controller_resource.rb:34:in `load_and_authorize_resource'
  cancancan (3.5.0) lib/cancan/controller_resource.rb:17:in `block in add_before_action'
...
(95 additional frame(s) were not displayed)

ActionController::UrlGenerationError: No route matches {:action=>"index", :controller=>"users", :id=>"56"} (ActionController::UrlGenerationError)
  app/controllers/application_controller.rb:13:in `block in <class:ApplicationController>'
    redirect_to({action: :index}, alert: t('controllers.authorization'))
...
(88 additional frame(s) were not displayed)
```

Mathis-Z:
Another way to fix this would be to check if there is a route for the #index action of the current controller before trying to redirect there but I could not find a method to do this so I think rescuing the error is the simplest solution. But maybe you have another idea?
I also ran all tests on this.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,7 +10,11 @@ class ApplicationController < ActionController::Base
 
   rescue_from CanCan::AccessDenied do |_exception|
     if current_user
-      redirect_to({action: :index}, alert: t('controllers.authorization'))
+      begin
+        redirect_to({action: :index}, alert: t('controllers.authorization'))
+      rescue ActionController::UrlGenerationError
+        redirect_to root_path, alert: t('controllers.authorization')
+      end
     else
       redirect_to root_path, alert: t('controllers.authorization')
     end
@@ -18,6 +22,8 @@ class ApplicationController < ActionController::Base
 
   rescue_from ActiveRecord::RecordNotFound do |_exception|
     redirect_to({action: :index}, alert: t('controllers.not_found'))
+  rescue ActionController::UrlGenerationError
+    redirect_to root_path, alert: t('controllers.not_found')
   end
 
   private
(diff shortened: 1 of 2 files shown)


MrSerth (approved):
Awesome, it's so great you had a look here and found this elegant solution—many thanks for that!

And yes, the approach you chose is perfectly fine. With your changes, we ensure that rescuing from the error always works as expected, regardless of the actions that are available for a specific controller. Therefore, I am pretty much in favor of this solution.

Also, I am really glad (and already was so a couple of days ago, when I first saw your PR), that you managed to a) identify the issue, b) fix it, and c) write some suitable tests for this 👏 

---

As you might have seen, I was also pushing some (minor) changes to your branch and hope that's okay; I just wanted to merge your branch afterward and had to do a rebase (to resolve merge conflicts). The background for this is that Mohamed also took this issue and worked on (quite a different) solution for the same problem. Hence, I rebased your changes and changed the test case to test with the `AccountLinksController`, where we've had the same underlying issue. As you can see, your tests are still preserved and nothing at all has changed for the `ApplicationController`.

PR #887, opened 2023-04-24
Fix for #881 "user accessing another user's profile"

Melhaya:
Added tests to track this new fix


MrSerth (approved):
Thanks for committing. I made a few adjustments:
- Remove outdated comments in the test, you took a note about your previous question there
- Simplify the test cases, by reducing the number of variables (and reusing the existing ones)
- Introduce shared examples to further reduce the test size. This is, of course, optional.

I would be glad if you could choose a more meaningful name for your branch the next time; this will make it easier to identify it later.

PR #891, opened 2023-04-28
Close #813 Reimplement licenses

### LINKED ISSUE #813: (Re-)Integrate Licenses for Exercises
Licenses should be available for exercises. Previously, CodeHarbor already had support for licenses, so that we should re-enable them:

- [ ] Allow users to select one of administrator-predefined licenses to an exercise
- [ ] Show the license when viewing an exercise

Mathis-Z:
I assumed it is okay for tasks to not have a license but also added a foreign key constraint for licenses to prevent them from being deleted if tasks still refer to them.
I added tests for the license model as well.
Looking forward to your feedback :)

### DIFF
--- a/app/views/tasks/_form.html.slim
+++ b/app/views/tasks/_form.html.slim
@@ -83,11 +83,11 @@
         /.field-element
           = f.label :labels, t('exercises.labels'), class: 'form-label'
           = f.select :labels, options_for_select(Label.all.collect{ |l| [l.name, l.name] }, @task.labels.pluck(:name)), {}, {class: "my-tag form-control", multiple: "multiple", style: "width: 300px;"}
-        /.field-element
+        .field-element
           = f.label :license, t('exercises.license'), class: 'form-label'
           .license
             .select
-              = f.select :license_id, options_for_select(License.all.collect{ |l| [l.name, l.id, {'data-link'=> l.link}] }, @license_default), {}, {class: 'form-control', disabled: @license_hidden}
+              = f.select :license_id, options_for_select(License.all.collect{ |l| [l.name, l.id, {'data-link'=> l.link}] }, @task.license_id), {include_blank: t('exercises.form.none')}, {class: 'form-control'}
             .link
     fieldset.form-group
       legend.toggle-next
--- a/app/views/tasks/_form.html.slim
+++ b/app/views/tasks/_form.html.slim
@@ -87,7 +87,7 @@
           = f.label :license, t('exercises.license'), class: 'form-label'
           .license
             .select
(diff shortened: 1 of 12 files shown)


app/views/tasks/_form.html.slim
```diff
@@ -83,11 +83,11 @@
         /.field-element
           = f.label :labels, t('exercises.labels'), class: 'form-label'
           = f.select :labels, options_for_select(Label.all.collect{ |l| [l.name, l.name] }, @task.labels.pluck(:name)), {}, {class: "my-tag form-control", multiple: "multiple", style: "width: 300px;"}
-        /.field-element
+        .field-element
           = f.label :license, t('exercises.license'), class: 'form-label'
           .license
             .select
-              = f.select :license_id, options_for_select(License.all.collect{ |l| [l.name, l.id, {'data-link'=> l.link}] }, @license_default), {}, {class: 'form-control', disabled: @license_hidden}
+              = f.select :license_id, options_for_select(License.all.collect{ |l| [l.name, l.id, {'data-link'=> l.link}] }, @task.license_id), {include_blank: t('exercises.form.none')}, {class: 'form-control'}
```
MrSerth, line 90: Would you mind updating the translation key to something below the `task` prefix? I know, it's still not consistent, but I would propose to move everything away from `exercise` to `task`. This will make it more consistent and potentially allow so-called [lazy lookups](https://guides.rubyonrails.org/i18n.html#lazy-lookup). Personally, I am not a fan of lazy lookups, as they could make it more difficult to move translations around, but that's another story ;).

If you move and change the locale, one could also consider introducing an explicit / new translation: What if we define no license chosen as "All rights reserved"? I am not a legal expert, but would argue that you don't gain permission in case someone doesn't provide an explicit license stating otherwise?

app/views/tasks/show.html.slim
```diff
@@ -164,8 +164,7 @@ legend.toggle-next.closed
         = ': '
         = link_to @task.license.link, @task.license.link
       - else
-        = t('exercises.show.no_license_chosen') + ' '
-        = link_to('Edit', edit_exercise_path(@task)) + '.'
+        = t('exercises.show.none')
```
MrSerth, line 167: Similar to above, we could update the local here.

MrSerth (commented):
Nice, well done! 📚

I tested your changes locally and think they look fine, there are just two minor remarks I've made.

app/views/tasks/_form.html.slim
Mathis-Z: I think not selecting a license should not result in "All rights reserved" being shown as a license. Just showing that no license was selected is most accurate in my opinion.

MrSerth: That's true. Let's keep it this way and allow all users to be aware of a missing, explicit license.

MrSerth (approved):
Awesome, that looks pretty good! I'd consider this increment to be ready, so let's continue merging 🚢.

PR #898, opened 2023-05-06
Re-enable labels for tasks

Mathis-Z:
This re-enables labels for tasks but they are not yet included in the task search. I think the look of the labels and the selection field for them has room for improvement but it should be fine for now. 
I had some trouble matching the style of the labels between the selection form and the task show page because of the limited customizability select2 offers for selected elements. Matching the style accurately would probably require overriding the SelectionAdapter.


spec/controllers/tasks_controller_spec.rb
```diff
@@ -154,6 +155,11 @@
         expect(assigns(:task).license).to match(valid_params[:license_id])
       end
 
+      it 'sets labels correctly' do
+        post_request
+        expect(assigns(:task).labels).to eq(valid_params[:label_ids])
```
kkoehn, line 160: the expectation would have to be updated accordingly

I would also like a spec to check if something has been saved into the database (ie check if TaskLabel count increases)

spec/controllers/tasks_controller_spec.rb
```diff
@@ -252,6 +274,11 @@
         expect { put_update }.to change { task.reload.license.name }.to('new_license')
       end
 
+      it 'updates the tasks labels' do
+        expect(task.labels).to eq([existing_label])
```
kkoehn, line 278: You don't really need this line, since you are basically testing if the factory works.
If you really want to make sure that `existing_label` is correctly set, you could add a `from` to the other expectation (expect-to change-from-to)

kkoehn (changes requested):
Thanks for taking care of the labels, they really needed some love.
I really like your implementation, but have one smallish request (see comments) front-end wise and a few notes about the specs

Great work!

app/views/tasks/show.html.slim
```diff
@@ -137,13 +137,13 @@ legend.toggle-next.closed
...
     .row-value
       = @task.private ? t('exercises.visibility.private') : t('exercises.visibility.public')
-  /.row
+  .row
     .row-label
       = t('exercises.labels')
       | :
     .row-value
       - @task.labels.each do |label|
-        .exercise_label style=("background-color: #{'#' + label.color.to_s}; color: #{'#' + label.font_color.to_s};")
```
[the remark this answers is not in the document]
Mathis-Z: I missed that this method existed and thought font_color was previously a label property that had been removed. I added it again.

Mathis-Z:
Looking at the label model I also found that I did not add the has_many statements in the label model. I included that in the new commit.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -132,6 +132,7 @@
           programming_language_id: create(:programming_language, :ruby).id,
           license_id: create(:license),
           language: 'de',
+          label_ids: [create(:label)],
```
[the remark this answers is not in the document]
Mathis-Z: I was not aware of this method. Included it in the new commit.

spec/factories/task_labels.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
Mathis-Z: I used that for a test but it turned out I did not need it and forgot to delete it. Removed it in the new commit.

spec/controllers/tasks_controller_spec.rb
Mathis-Z: I added the test for the TaskLabel count.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -252,6 +278,10 @@
         expect { put_update }.to change { task.reload.license.name }.to('new_license')
       end
 
+      it 'updates the tasks labels' do
+        expect { put_update }.to change { task.reload.labels }.to([new_label]).from([existing_label])
```
kkoehn, line 282: I would put it the outer way around - sounds more natural:

```suggestion
        expect { put_update }.to change { task.reload.labels }.from([existing_label]).to([new_label])
```

app/assets/javascripts/labels.coffee
```diff
@@ -1,22 +1,24 @@
...
+      $(container).css("background-color", "#"+$(data.element).attr("label_color"));
+      $(container).css("color", "#"+$(data.element).attr("label_font_color"));
+      $(container).children("span").css("color", "#"+$(data.element).attr("label_font_color"));
+
+      $template = $('<span></span>').text(data.text);
+      $template.css({"font-size" : "80%", "font-weight" : "bold", "color" : "#"+$(data.element).attr("label_font_color")});
+      return $template;
+
+    templateResult: (data, container) ->
+      $template = $('<div></div>').text(data.text).addClass("exercise_label");
```
kkoehn, line 18: Please migrate the css-class as well (move from `exercises.scss` to `tasks.scss` and rename to `task_label`)
Don't forget to search and replace for all occurrences of the old class

app/views/tasks/_form.html.slim
```diff
@@ -80,9 +80,10 @@
                     = ', '
                   = group.name
 
-        /.field-element
+        .field-element
           = f.label :labels, t('exercises.labels'), class: 'form-label'
```
kkoehn, line 84: please migrate the translation(s?) as well

app/assets/javascripts/labels.coffee
```diff
@@ -1,22 +1,24 @@
...
+    maximumSelectionLength: 5
+    closeOnSelect: false
+
+    templateSelection: (data, container) ->
+      $(container).css("background-color", "#"+$(data.element).attr("label_color"));
+      $(container).css("color", "#"+$(data.element).attr("label_font_color"));
+      $(container).children("span").css("color", "#"+$(data.element).attr("label_font_color"));
+
+      $template = $('<span></span>').text(data.text);
+      $template.css({"font-size" : "80%", "font-weight" : "bold", "color" : "#"+$(data.element).attr("label_font_color")});
```
kkoehn, line 14: line 10 should be enough to set the color correctly

```suggestion
      $template.css({"font-size" : "80%", "font-weight" : "bold"});
```

app/assets/javascripts/labels.coffee
Mathis-Z: I moved the content of exercises.scss into tasks.scss and removed exercises.scss.

kkoehn: That's not quite what I meant, sorry for being inconclusive.
I wanted you to migrate only the classes you use in this PR (mainly exercise_label/task_label) to the new css-file.
The plan is to slowly migrate all the classes over to the new file, whenever we reimplement a feature and at some point we can just get rid of the old css-file.

kkoehn (approved):
Everything looks good to me now. I just moved the task_label styles to the tasks.scss.
You can merge this if you want or we can take a look at your other branch first.

MrSerth (commented):
Nice, thanks for working on the labels! 💪 I just made two minor comments, but otherwise think it's good to go.

app/views/tasks/_tasks.html.slim
```diff
@@ -64,7 +64,7 @@
               = t('tasks.defaults.no_labels')
           - else
             - exercise.labels.each do |label|
-              .exercise_label style=("background-color: #{'#' + label.color.to_s}; color: #{'#' + label.font_color.to_s};")
+              .task_label style=("background-color: #{'#' + label.color.to_s}; color: #{'#' + label.font_color.to_s};")
```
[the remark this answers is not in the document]
Mathis-Z: You are right, it is commented out. I renamed the CSS class from exercise_label to task_label and changed all occurrences including this one so it would not be broken if we decide to show the labels on the search page again.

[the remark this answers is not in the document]
MrSerth: I am fine with showing the labels on the search page again 👍. When doing so, there are also a few occurrences of `exericse` that needs to be fixed 😉.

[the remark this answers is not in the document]
Mathis-Z: Move to PR #912

MrSerth (approved):
Awesome, thanks for working on the labels! 💪

Issue #894, opened 2023-05-01
Re-enable privacy settings for tasks


MrSerth:
Previously, we allowed tasks to be public and private. As part of our ongoing refactoring, we would like to re-enable the privacy settings for tasks.

Probably, the following steps should be considered:

- [x] Include a setting in the edit and show views, allowing users to either select "private" or "public".
  - [x] This setting should be independent of any groups or collections the task is in. A group might grant additional users permissions to edit a task, and this should still work: A private task belonging to a group of three users is visible only to these three users
- [x] Add an icon or badge when viewing all tasks (index route) or a specific task (show)
  - [x] We could either use a badge (as done on GitHub) or an icon, such as one from [FontAwesome](https://fontawesome.com/search).
- [x] Re-include the privacy settings in the search
- [x] Allow anyone to view public exercises, but without edit or delete functionalities
  - [x] Admins should still be able to do everything.

MrSerth:
This ticket has been resolved with #905.

PR #905, opened 2023-05-16
Implements #894: Visibility settings for tasks

florian-str:
implements #894


app/models/ability.rb
```diff
@@ -60,16 +60,23 @@ def collection_abilities(user)
...
+      task.author?(user) || task.in_same_group_admin?(user)
+    end
   end
 
   def task_file_abilities(user)
     can %i[download_attachment extract_text_data], TaskFile do |task_file|
       fileable = task_file.fileable
       task = fileable.is_a?(Task) ? fileable : fileable.task
-      task.can_access(user)
+      task.author?(user)
```
kkoehn, line 79: according to #894 this should be possible for everyone who has access to the task itself

app/models/task.rb
```diff
@@ -45,20 +45,40 @@ class Task < ApplicationRecord
...
+    in_same_group_member?(user) || in_same_group_admin?(user)
+  end
+
+  def in_same_group_member?(user)
+    groups.any? {|group| group.confirmed_member?(user) }
+  end
+
+  def in_same_group_admin?(user)
+    groups.any? {|group| group.admin?(user) }
+  end
```
kkoehn, line 81: I'm not really happy with these methods as well, but have no idea how to fix this currently, will come back to this

kkoehn (commented):
Thanks for taking on the privacy_settings.
You took a good approach and have a nice implementation.

I wasn't able to review it fully, but already wanted to give you some feedback

app/models/task.rb
```diff
@@ -45,20 +45,40 @@ class Task < ApplicationRecord
...
-    %i[created_before_days min_stars]
+    %i[created_before_days min_stars access_level]
   end
 
   def self.ransackable_attributes(_auth_object = nil)
     %w[title description programming_language_id created_at]
   end
 
   def can_access(user)
+    author?(user)
```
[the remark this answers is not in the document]
MrSerth: While I haven't had a detailed look, I at least identify two discussions in this thread:

1) The suggestion about using cancan methods from the model, without re-defining them. I would like to have that, as it would remove the duplication. However, since it is still used in the ProFormA service, this would require us touching other parts as well.
2) The question about moving the method to the `user.rb`, rather than the `task.rb`. Indeed, this is the part we talked about. At first, my impression was to move it to the user as well. However, when thinking about how to implement that (and the semantics), I wasn't sure and rather thought of making it similar to the existing method and therefore keeping it with the task.

app/models/ability.rb
MrSerth: Indeed. However, I am unsure about `extract_text_data`, since this behaves similar to a `update` operation (and changes the task / task file).

app/models/task.rb
MrSerth: If you have an idea, I would really appreciate that. Otherwise, we just leave it for now.

app/models/ability.rb
kkoehn: You are right.
- `download_attachment` should be possible if you can view the task
- `extract_text_data` only if you can update the task

app/models/task.rb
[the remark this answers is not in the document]
kkoehn: Concerning 2: I think I would prefer adding a `can` method to `user` and use that to implement a `is_accessible_by` in `task`

app/models/task.rb
kkoehn: it seems counter-intuitive to solve it like this, but I can't think of a better way, I think we can leave it as well

db/schema.rb
```diff
@@ -250,6 +250,7 @@
     t.jsonb "meta_data", default: {}
     t.bigint "license_id"
     t.index ["license_id"], name: "index_tasks_on_license_id"
+    t.integer "access_level", limit: 2, default: 0, null: false, comment: "Used as enum in Rails"
```
kkoehn, line 253: the migration add the this line before the `licence_id`-index for me
try to redo the migration or move this line up by yourself

spec/models/task_spec.rb
```diff
@@ -8,9 +8,17 @@
 
     let(:user) { nil }
     let(:task_user) { create(:user) }
-    let(:task) { create(:task, user: task_user) }
-    let(:private?) { false }
+    let(:group_user) { create(:user) }
+    let(:access_level) { 'private' }
+    let!(:group) { create(:group) }
```
kkoehn, line 13:
```suggestion
    let(:group_memberships) { [build(:group_membership, :with_admin), build(:group_membership, user: group_user, role:)] }
    let(:group) { create(:group, group_memberships:) }

```

and remove the `before`-block in line 18++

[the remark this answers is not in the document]
kkoehn: see suggested change

app/views/tasks/_tasks.html.slim
```diff
@@ -27,6 +27,9 @@
       h3.panel-title
         span data-toggle="tooltip" data-placement="bottom" title="#{task.title}"
           = truncate(task.title, length: 100)
+        - if @visibility == 'owner'
+          span.programming-language.pull-left data-toggle="tooltip" data-placement="bottom" title="#{t('exercises.visibility.label')}"
```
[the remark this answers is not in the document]
florian-str: In this version, I have added the `.visibility` class to the definition of `programming-language` as the style is currently identical. However, it still allows editing the design in the future easily.

app/models/ability.rb
```diff
@@ -60,16 +60,28 @@ def collection_abilities(user)
...
-
     alias_action :export_external_start, :export_external_check, :export_external_confirm, to: :export
-    can %i[crud export download add_to_collection], Task, user: {id: user.id}
+    can %i[show export download add_to_collection], Task do |task|
+      task.access_level_public? || task.in_same_group?(user) || task.author?(user)
+    end
+    can %i[update], Task do |task|
+      task.in_same_group_member?(user)
+    end
+    can %i[crud], Task do |task| # , user: {id: user.id}
```
kkoehn, line 70: `# , user: {id: user.id}`
is the comment still needed?

app/views/tasks/_tasks.html.slim
```diff
@@ -27,6 +27,9 @@
       h3.panel-title
         span data-toggle="tooltip" data-placement="bottom" title="#{task.title}"
           = truncate(task.title, length: 100)
+        - if @visibility == 'owner'
+          span.visibility.pull-left data-toggle="tooltip" data-placement="bottom" title="#{t('exercises.visibility.label')}"
+            = t("exercises.visibility.#{task.access_level}")
```
kkoehn, line 32:
```suggestion
          span.visibility.pull-left data-toggle="tooltip" data-placement="bottom" title="#{t('tasks.visibility.label')}"
            = t("tasks.visibility.#{task.access_level}")
```

app/views/tasks/index.html.slim
```diff
@@ -62,6 +62,12 @@
             b = t('tasks.index.search.created')
           span.form-group
             = f.select 'created_before_days', options_for_select({t('tasks.index.search.all_time') => 0, t('tasks.index.search.today') => 1, t('tasks.index.search.last_week') => 7, t('tasks.index.search.last_month') => 30}, selected: @created_before_days), {}, class: 'form-control defaultSelect2 ransack-filter'
+      - if @visibility == 'owner'
+        .other.section
+          p
+            b = t('exercises.visibility.label')
+          span.form-group
+            = f.select 'access_level', options_for_select({t('exercises.search.all_access_levels') => '', t('exercises.visibility.private') => 'private', t('exercises.visibility.public') => 'public'}, selected: @access_level), {}, class: 'form-control defaultSelect2 ransack-filter'
```
kkoehn, line 70:
```suggestion
            b = t('tasks.visibility.label')
          span.form-group
            = f.select 'access_level', options_for_select({t('tasks.search.all_access_levels') => '', t('tasks.visibility.private') => 'private', t('tasks.visibility.public') => 'public'}, selected: @access_level), {}, class: 'form-control defaultSelect2 ransack-filter'
```

app/views/tasks/show.html.slim
```diff
@@ -131,12 +131,12 @@ legend.toggle-next.closed
...
     .row-value
       = @task.language
-  /.row
+  .row
     .row-label
       = t('exercises.visibility.label')
       | :
     .row-value
-      = @task.private ? t('exercises.visibility.private') : t('exercises.visibility.public')
+      = t("exercises.visibility.#{@task.access_level}")
```
kkoehn, line 139:
```suggestion
      = t("tasks.visibility.#{@task.access_level}")
```

kkoehn (commented):
looks very good
please use the tasks-translations everywhere

app/models/ability.rb
```diff
@@ -60,16 +60,28 @@ def collection_abilities(user)
 
   def task_abilities(user)
     can %i[index create import_start import_confirm], Task
-
     alias_action :export_external_start, :export_external_check, :export_external_confirm, to: :export
-    can %i[crud export download add_to_collection], Task, user: {id: user.id}
+    can %i[show export download add_to_collection], Task do |task|
```
MrSerth, line 64: With this ability, we allow users to add a public task to a collection. Now, I wondered what happens if someone adds a public task to a collection and the task's author later decides to switch the visibility to private. In this scenario, the external user probably should not have access to the task any longer. But I assume that it (and the title) is still shown in the collection? Do we need to add some special handling to this case of changing the visibility to private later on? One option could be to hide all private tasks for collections, so that they are still included if the visibility is turned back to "public". 🤔

MrSerth (commented):
Well done, many thanks for working on the visibility settings for tasks! I cannot wait to see this live, with such nice improvements on the way. I just made a few comments along the way 😊

app/views/tasks/_form.html.slim
```diff
@@ -123,7 +123,7 @@
     fieldset.form-group
       legend.toggle-next
         span.number = '4'
-        = t('exercises.tests.label') + ' '
+        = t('tasks.tests.label') + ' '
```
MrSerth, line 126:
```suggestion
        => t('tasks.tests.label')
```

app/views/tasks/_form.html.slim
```diff
@@ -149,7 +149,7 @@
           / .field-element
           = f.add_nested_fields_link :tests, ' ', class:'btn btn-main', id: 'addTestButton' do
             i.fa.fa-plus style=("color: #008000")
-            = ' ' + t('exercises.form.tests.add')
+            = ' ' + t('tasks.form.tests.add')
```
MrSerth, line 152:
```suggestion
            =< + t('tasks.form.tests.add')
```

app/assets/javascripts/tasks_form.coffee
```diff
@@ -21,3 +22,11 @@ initializeFileTypeSelection = ->
...
     changeEditorMode(editor, getModeByFileExtension($(this).val()))
   $('.file-name').keyup()
+
+initializeVisibilityWarning = ->
+  warning_message = $('#task_visibility_warning')
+  $('#task_access_level_private').on 'change', ->
+    if warning_message.data("external-collection-membership")
+      warning_message.removeClass('hidden')
+  $('#task_access_level_public').on 'change', ->
+      warning_message.addClass('hidden')
```
MrSerth, line 32: It would be great if you could add an empty line here. This is usually desired, since appending the file will then show correctly in a Git diff.

config/locales/en.yml
```diff
@@ -1055,6 +1055,7 @@ en:
         add: "Add Model solution"
       model_solution:
         remove: "Remove Model solution"
+      visibility_warning: "This task is currently in a collection. Other users will not be able to view or modify your task. However the title will remain visible to some users."
```
MrSerth, line 1058:
```suggestion
      visibility_warning: "This task is currently in a collection. Other users will not be able to view or modify your task. However, the title will remain visible to these users."
```

app/models/task.rb
```diff
@@ -35,7 +35,10 @@ class Task < ApplicationRecord
 
   scope :owner, ->(user) { where(user:) }
   scope :public_access, -> { where(access_level: 'public') }
-  scope :visibility, ->(visibility, user = nil) { {owner: owner(user), public: public_access}.with_indifferent_access[visibility] }
+  scope :private_access, ->(user) { where(access_level: 'private', user:) }
```
MrSerth, line 38: This private access shows only those tasks the given `user` is the author _and_ those who are `private`. Depending on the use case, this is okay / desired, but might change depending on the discussion in https://github.com/openHPI/codeharbor/pull/905#discussion_r1211467594.

app/views/tasks/_form.html.slim
MrSerth: Thanks for fixing this style in the whole file :+1:

PR #906, opened 2023-05-17
Reimplement labels with ajax option loading and label creation

Mathis-Z:
This allows users to create new labels and also changes the select option loading to use ajax in order to not always download all labels.

### DIFF
--- a/app/assets/javascripts/labels.coffee
+++ b/app/assets/javascripts/labels.coffee
@@ -25,8 +25,8 @@ ready = ->
 
     templateSelection: (data, container) ->
       text = data.text;
-      bg_color = "#" + ($(data.element).attr("label_color") || data.label_color || default_label_color);
-      color = "#" + ($(data.element).attr("label_font_color") || data.label_font_color || default_label_font_color);
+      bg_color = "#" + (data.label_color || $(data.element).attr("label_color") || default_label_color);
+      color = "#" + (data.label_font_color || $(data.element).attr("label_font_color") || default_label_font_color);
 
       if data.newTag == true
         text += I18n.t('labels.new.selection_suffix');
--- a/app/assets/javascripts/labels.coffee
+++ b/app/assets/javascripts/labels.coffee
@@ -1,9 +1,14 @@
 ready = ->
+  max_label_length = 15
+  default_label_color = "e4e4e4"
+  default_label_font_color = "000000"
+
   $('.my-tag').select2
     width: '100%'
     tags: true
     multiple: true
     maximumSelectionLength: 5
+    maximumInputLength: max_label_length
     closeOnSelect: false
     tokenSeparators: [',']
 
@@ -20,11 +25,11 @@ ready = ->
 
     templateSelection: (data, container) ->
       text = data.text;
-      bg_color = "#" + ($(data.element).attr("label_color") || data.label_color || "e4e4e4");
-      color = "#" + ($(data.element).attr("label_font_color") || data.label_font_color || "000000");
+      bg_color = "#" + ($(data.element).attr("label_color") || data.label_color || default_label_color);
+      color = "#" + ($(data.element).attr("label_font_color") || data.label_font_color || default_label_font_color);
 
       if data.newTag == true
-        text += " (new)";
+        text += I18n.t('labels.new.selection_suffix');
 
       $(container).css({"background-color" : bg_color, "color" : color});
       $(container).children("span").css("color", color);
@@ -38,11 +43,11 @@ ready = ->
         return data.text;
 
       text = data.text;
-      bg_color = "#" + (data.label_color || "e4e4e4");
-      color = "#" + (data.label_font_color || "000000");
+      bg_color = "#" + (data.label_color || default_label_color);
+      color = "#" + (data.label_font_color || default_label_font_color);
 
       if data.newTag == true
-        text += " (new)";
+        text += I18n.t('labels.new.selection_suffix');
 
       $template = $('<div></div>').text(text).addClass("task_label");
       $template.css({"background-color" : bg_color, "color" : color});
@@ -52,7 +57,7 @@ ready = ->
       term = $.trim(params.term)
       selection = $('#task_label_names').select2('data').map (element) -> element.id
 
-      if term == '' || term.length > 15 || selection.includes(term)
+      if term == '' || term.length > max_label_length || selection.includes(term)
         return null;
 
       return {
--- a/app/assets/javascripts/labels.coffee
+++ b/app/assets/javascripts/labels.coffee
@@ -1,36 +1,82 @@
 ready = ->
   $('.my-tag').select2
     width: '100%'
+    tags: true
     multiple: true
+    minimumInputLength: 1
     maximumSelectionLength: 5
     closeOnSelect: false
 
-    templateSelection: (data, container) ->
-      $(container).css("background-color", "#"+$(data.element).attr("label_color"));
-      $(container).css("color", "#"+$(data.element).attr("label_font_color"));
-      $(container).children("span").css("color", "#"+$(data.element).attr("label_font_color"));
+    ajax: {
+      url: '/labels/search',
+      dataType: 'json',
+      data: (params) ->
+        query = {
+          search: params.term,
+          page: params.page || 1
+        }
+        return query;
+    }
 
-      $template = $('<span></span>').text(data.text);
+    ajax: {
+      url: '/labels/search',
+      dataType: 'json',
+      data: (params) ->
+        query = {
+          search: params.term,
+          page: params.page || 1
+        }
+        return query;
+    }
+
+    templateSelection: (data, container) ->
+      text = data.text;
+      bg_color = "#" + ($(data.element).attr("label_color") || data.label_color || "e4e4e4");
+      color = "#" + ($(data.element).attr("label_font_color") || data.label_font_color || "000000");
+
+      if data.newTag == true
+        text += " (new)";
+
+      $(container).css({"background-color" : bg_color, "color" : color});
+      $(container).children("span").css("color", color);
+
+      $template = $('<span></span>').text(text);
       $template.css({"font-size" : "80%", "font-weight" : "bold"});
       return $template;
 
-    templateResult: (data, container) ->
-      $template = $('<div></div>').text(data.text).addClass("task_label");
-      $template.css({"background-color" : "#"+$(data.element).attr("label_color"), "color" : "#"+$(data.element).attr("label_font_color")});
+    templateResult: (data) ->
+      if (data.loading)
+        return data.text;
+
+      text = data.text;
+      bg_color = "#" + (data.label_color || "e4e4e4");
+      color = "#" + (data.label_font_color || "000000");
+
+      if data.newTag == true
+        text += " (new)";
+
+      $template = $('<div></div>').text(text).addClass("task_label");
+      $template.css({"background-color" : bg_color, "color" : color});
       return $template;
 
-    createSearchChoice: (term, data) ->
-      if $(data).filter((->
-        @text.localeCompare(term) == 0
-      )).length == 0
-        return {
-          id: term
-          text: term
-        }
-      return
+    createTag: (params) ->
+      if params.term == '' || params.term.length > 15
+        return null;
+
+      return {
+        id: params.term,
+        text: params.term,
+        newTag: true
+      }
 
     formatSelectionTooBig: (limit) ->
       I18n.t('labels.can_only_create_5')
+
+    $('#task_label_names').on 'select2:select', clear_input
+  return
+
+clear_input = ->
+  $('#task_label_names').siblings(".select2").find("input").val("");
   return
 
 $(document).on 'turbolinks:load', ready
--- a/app/assets/javascripts/labels.coffee
+++ b/app/assets/javascripts/labels.coffee
@@ -3,20 +3,9 @@ ready = ->
     width: '100%'
     tags: true
     multiple: true
-    minimumInputLength: 1
     maximumSelectionLength: 5
     closeOnSelect: false
-
-    ajax: {
-      url: '/labels/search',
-      dataType: 'json',
-      data: (params) ->
-        query = {
-          search: params.term,
-          page: params.page || 1
-        }
-        return query;
-    }
+    tokenSeparators: [',']
 
     ajax: {
       url: '/labels/search',
@@ -60,12 +49,13 @@ ready = ->
       return $template;
 
     createTag: (params) ->
-      if params.term == '' || params.term.length > 15
+      term = $.trim(params.term)
+      if term == '' || term.length > 15
         return null;
 
       return {
-        id: params.term,
-        text: params.term,
+        id: term,
+        text: term,
         newTag: true
       }
 
--- a/app/assets/javascripts/labels.coffee
+++ b/app/assets/javascripts/labels.coffee
@@ -50,7 +50,9 @@ ready = ->
 
     createTag: (params) ->
       term = $.trim(params.term)
-      if term == '' || term.length > 15
+      selection = $('#task_label_names').select2('data').map (element) -> element.id
+
+      if term == '' || term.length > 15 || selection.includes(term)
         return null;
 
       return {
--- a/app/controllers/labels_controller.rb
+++ b/app/controllers/labels_controller.rb
@@ -4,6 +4,18 @@ class LabelsController < ApplicationController
   load_and_authorize_resource
 
   def search
-    @labels = Label.order(:name)
+    page = params[:page].to_i
+    if page <= 0
+      page = 1
+    end
+
+    results = Label.ransack(name_i_cont: params[:search]).result
+    paginated = results.paginate(per_page: 3, page:)
+
+    response = {
+      results: paginated.map {|l| {id: l.name, text: l.name, label_color: l.color, label_font_color: l.font_color} },
+      pagination: {more: (page * 3 < results.length)},
+    }
+    render json: response
   end
 end
--- a/app/controllers/labels_controller.rb
+++ b/app/controllers/labels_controller.rb
@@ -3,19 +3,21 @@
 class LabelsController < ApplicationController
   load_and_authorize_resource
 
+  before_action :sanitize_page_param, only: :search
   def search
-    page = params[:page].to_i
-    if page <= 0
-      page = 1
-    end
-
     results = Label.ransack(name_i_cont: params[:search]).result
-    paginated = results.paginate(per_page: 3, page:)
+    paginated = results.paginate(per_page: 3, page: @page)
 
-    response = {
+    render json: {
       results: paginated.map {|l| {id: l.name, text: l.name, label_color: l.color, label_font_color: l.font_color} },
-      pagination: {more: (page * 3 < results.length)},
+      pagination: {more: (@page * 3 < results.length)},
     }
-    render json: response
+  end
+
+  def sanitize_page_param
+    @page = params[:page].to_i
+    if @page <= 0
+      @page = 1
+    end
   end
 end
--- a/app/controllers/labels_controller.rb
+++ b/app/controllers/labels_controller.rb
@@ -2,22 +2,13 @@
 
 class LabelsController < ApplicationController
   load_and_authorize_resource
-
-  before_action :sanitize_page_param, only: :search
   def search
     results = Label.ransack(name_i_cont: params[:search]).result
-    paginated = results.paginate(per_page: 3, page: @page)
+    paginated = results.paginate(per_page: 3, page: params[:page])
 
     render json: {
       results: paginated.map {|l| {id: l.name, text: l.name, label_color: l.color, label_font_color: l.font_color} },
-      pagination: {more: (@page * 3 < results.length)},
+      pagination: {more: paginated.next_page.present?},
     }
-  end
-
-  def sanitize_page_param
-    @page = params[:page].to_i
-    if @page <= 0
-      @page = 1
-    end
   end
 end
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -139,6 +139,7 @@ class Task < ApplicationRecord
 
   def label_names=(label_names)
     self.labels = label_names.uniq.compact_blank.map {|name| Label.find_or_initialize_by(name:) }
+    Label.where.not(id: TaskLabel.select(:label_id)).destroy_all
   end
 
   def label_names
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -35,6 +35,7 @@ class Task < ApplicationRecord
   accepts_nested_attributes_for :tests, allow_destroy: true
   accepts_nested_attributes_for :model_solutions, allow_destroy: true
   accepts_nested_attributes_for :group_tasks, allow_destroy: true
+  accepts_nested_attributes_for :labels
 
   scope :owner, ->(user) { where(user:) }
   scope :public_access, -> { where(access_level: :public) }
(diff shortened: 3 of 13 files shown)


spec/controllers/tasks_controller_spec.rb
```diff
@@ -252,6 +278,16 @@
         expect { put_update }.to change { task.reload.license.name }.to('new_license')
       end
 
+      it 'updates the tasks labels' do
+        expect { put_update }.to change { task.reload.labels }.from([existing_label]).to([new_label])
+      end
+
+      it 'creates new task label when requested' do
+        changed_attributes[:label_names] = ['some new label']
```
kkoehn, line 286: this would fit in a new context, but you can probably just set it in the `let` in line 243

We may want to add an example again where we check if a new `Label` is created.

kkoehn (changes requested):
Again a good implementation, I marked some minor things
I still think, we should merge the other PR before, if it doesn't lead to merge major merge conflicts.

app/assets/javascripts/labels.coffee
```diff
@@ -1,35 +1,82 @@
...
+    ajax: {
+      url: '/labels/search',
+      dataType: 'json',
+      data: (params) ->
+        query = {
+          search: params.term,
+          page: params.page || 1
+        }
+        return query;
+    }
```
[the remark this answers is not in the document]
Mathis-Z: I struggled a bit with git because I had to rebase this branch onto the newer reimplement_labels branch and there were some conflicts. That is probably where this comes from.

app/assets/javascripts/labels.coffee
```diff
@@ -1,35 +1,82 @@
...
+
+      if data.newTag == true
+        text += " (new)";
+
+      $template = $('<div></div>').text(text).addClass("task_label");
+      $template.css({"background-color" : bg_color, "color" : color});
+      return $template;
+
     createTag: (params) ->
-      term = $.trim(params.term)
```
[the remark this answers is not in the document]
Mathis-Z: I missed that. Readded it.

app/assets/javascripts/labels.coffee
```diff
@@ -1,35 +1,82 @@
...
+      return {
+        id: params.term,
+        text: params.term,
         newTag: true
       }
-    width: '100%'
-    tokenSeparators: [
-      ','
-      ' '
-    ]
```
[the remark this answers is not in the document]
Mathis-Z: The problem with having space as a tokenSeparator is that you cannot create labels like "Data Structures" anymore because it will separate on the space. I added the colon as a separator though.

app/controllers/labels_controller.rb
```diff
@@ -3,7 +3,21 @@
 class LabelsController < ApplicationController
   load_and_authorize_resource
 
+  before_action :sanitize_page_param, only: :search
   def search
-    @labels = Label.order(:name)
+    results = Label.ransack(name_i_cont: params[:search]).result
+    paginated = results.paginate(per_page: 3, page: @page)
+
+    render json: {
```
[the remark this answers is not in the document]
Mathis-Z: select2 preloads more than one page so in order to notice the pagination you need more labels. On my system it preloads 3 pages so you would need to have more than 9 labels in your system to notice the pagination.

spec/controllers/tasks_controller_spec.rb
Mathis-Z: I am not quite sure what you mean by that. Maybe you could make a code suggestion?

Mathis-Z:
While testing the tokenSeparators for the select2 I noticed it was possible to create the same label multiple times.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -252,6 +278,16 @@
...
+        expect { put_update }.to change { task.reload.labels.map(&:name) }.from([existing_label.name]).to(['some new label'])
+        expect(Label.where(name: 'some new label')).not_to be_empty
+      end
```
kkoehn, line 289: since you already checked that the update works in the example before I would just add the following
```suggestion
      it 'creates new task label when requested' do
        expect { put_update }.to change(Label, :count).by(1)
      end
```

spec/controllers/tasks_controller_spec.rb
kkoehn: see other comment

spec/controllers/tasks_controller_spec.rb
Mathis-Z: I moved the test into a new context. Maybe that is better.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -252,6 +278,27 @@
...
+      it 'updates the tasks labels' do
+        expect { put_update }.to change { task.reload.labels }.from([existing_label]).to([new_label])
+      end
+
+      context 'when requesting new label to be created' do
+        subject(:put_update) do
+          put :update, params: {id: task.to_param, task: attributes}
+        end
+
+        let(:attributes) { {label_names: [not_existing_label_name]} }
```
kkoehn, line 290: Almost, but you don't really need to redefine the put:

```suggestion
        let(:changed_attributes) { {label_names: [not_existing_label_name]} }
```

kkoehn: I would also add the following expectation, since we identified the issue of not being exactly sure what Label gets created:

```ruby
        it 'creates a label with the correct name' do
          put_update
          expect(Label.last.name).to eql not_existing_label_name
        end
```

Mathis-Z: I committed your suggestions. Thanks for the feedback!

app/controllers/labels_controller.rb
```diff
@@ -3,7 +3,21 @@
...
+      results: paginated.map {|l| {id: l.name, text: l.name, label_color: l.color, label_font_color: l.font_color} },
+      pagination: {more: (@page * 3 < results.length)},
```
MrSerth, line 13: `results.length` is not optimal here, it will render the pagination somewhat useless. In order to understand why, one has to look what happens here:

The `length` method will get the total number of entries of an array. At this point, `results` is not an array, but rather an `ActiveRecord_Relation`. This is a wrapper for the database call, and the actual results or only lazy loaded, when needed. Since the pagination only uses the database query and further customizes it (by limiting the result size and specifying an offset), this object (and thus all label objects) still aren't loaded. However, keep in mind that the original database query is unchanged (i.e., with the limit clause). Now, you are getting the length of the result array. Hence, the original query has to be fulfilled, in order to answer this question. Consequently, all matching labels will be fetched from the database, instantiated in an object and put into this array. Once everything is done, you get the length. As you see, we were still doing the regular processing for all matching labels, and the pagination didn't kick in. In this case, it would have been better to query the database just for the one number you're interested in: The result size. This would be done with a `COUNT(*)` query, and Active Record allows you to do so easily: `results.count`. This way, we omit transferring all labels and instantiating them, and leaving the hard work to the database.

Try the difference locally, you'll notice a change in the database query:

```
Label Load (2.8ms)  SELECT "labels".* FROM "labels"
Label Count (1.6ms)  SELECT COUNT(*) FROM "labels"
```

This change is better, but still not perfect. If the `paginate` already knows how many objects we have, we aren't using the result. Hence, we should leave it to paginate to decide whether another database query is needed for the overall number of results. Do so with `paginated.total_entries`. [Here](https://github.com/mislav/will_paginate/blob/8aa3a87022e4c17cb6ebf2ec0c9467bcdfd76b18/lib/will_paginate/active_record.rb#L68-L79) is the definition of the method. As you see, it will try calculating the overall size without further querying the database if possible, and otherwise switch to a database query using a raw `COUNT`. Therefore, we should use this method.

MrSerth, line 13: After my previous two suggestion, this line reads as:

```rb
      pagination: {more: (@page * paginated.limit_value < paginated.total_entries)},
```

I feel, this is better, but still not perfect. When checking the library further, I saw the following possibilities:

```rb
      pagination: {more: paginated.next_page.present?},
```

[According to the source](https://github.com/mislav/will_paginate/blob/8113fddf4b506867e9af3aa59fe2df314b79860c/lib/will_paginate/collection.rb#L24-L27), the `next_page` will either return the number of the next page or nil. Both is fine, and we can convert this result to the desired boolean value with `present?`.

Overall, I would go with this solution. I hope you agree that this solution is more elegant and also easier to read :).

app/controllers/labels_controller.rb
```diff
@@ -3,7 +3,21 @@
...
+    }
+  end
+
+  def sanitize_page_param
```
MrSerth, line 17: Is there a specific reason you introduced this method? Sure, it will catch the case where no page number or a negative one is given, but is this likely to occur? There is one case you aren't checking so far: Someone could also get a page number that is larger than the total number of available pages. For the latter, WillPaginate already comes with a [helper method](https://github.com/mislav/will_paginate/blob/8113fddf4b506867e9af3aa59fe2df314b79860c/lib/will_paginate/collection.rb#L29-L35).

Since we no longer need the `@page` variable twice, I would suggest to just use the return value of this method (and without it being a before filter) if we want to keep it. Then, we should probably also add a test case 😉.

app/models/task.rb
```diff
@@ -32,6 +35,7 @@ class Task < ApplicationRecord
   accepts_nested_attributes_for :tests, allow_destroy: true
   accepts_nested_attributes_for :model_solutions, allow_destroy: true
   accepts_nested_attributes_for :group_tasks, allow_destroy: true
+  accepts_nested_attributes_for :labels
```
MrSerth, line 38: Here, we are creating new labels indirectly. We should probably also consider what happens if a label is no longer referred by any task. Currently, we are still keeping it. Is this desired ("yes" is a valid response) or should we clean unused labels?

spec/controllers/labels_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  render_views
+
+  let(:user) { create(:user) }
+  let(:label) { create(:label) }
+
+  describe 'GET #search' do
+    context 'without being signed in' do
+      it 'redirects to other page' do
+        get :search, params: {search: label.name, page: 1}
+        expect(response.body).to match(%r{^<html><body>You are being.*redirected.*</body></html>$})
```
MrSerth, line 15: Rather than checking for the HTTP response no one should ever read, I would check for the actual redirect. What do you think of these expectations:

```rb
expect(response).to have_http_status(:redirect)
expect(response).to redirect_to(root_url)
```

spec/controllers/labels_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+    end
+
+    context 'with valid params' do
+      subject(:request) { get :search, params: {search: label.name, page: 1} }
```
MrSerth, line 20: Usually, we used the name `get_request` for similar test cases.

MrSerth, line 20: If you move this subject to the above `describe` context, you should be able to reuse it in the `without being signed in` context. It should work there as well, I assume 🤔

spec/controllers/labels_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      before { sign_in user }
+
+      it 'returns valid json' do
+        request
+        expect(response.body).to match(/^{"results":\[.*\],"pagination":{"more":(true|false)}}$/)
```
MrSerth, line 26: While the test suggests that you are checking for a valid JSON, this is not exactly true. You are checking that the response matches a given pattern, which will match a valid JSON, but there are also constellations, where an invalid JSON would still match. "Just" matching for a valid JSON would be performed by trying to parse the JSON and checking for any parse errors. However, I doubt this test would be useful, since generating and parsing the JSON is mostly performed by libraries we haven't written ourselves (and thus the test would cover the library functionality...). Therefore, ask yourself: What do you _actually_ want to test here? I think you had a good intention, but just need to rework the test description `returns valid json` and the expectations.

For the expectations, the following might come handy: `response.parsed_body` will give you a hash representation of the JSON, allowing you to specify more fine granular tests. For example, you could check for some predefined keys to be present in the response or some data types (`results` should be an array, ...)

spec/controllers/labels_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      it 'returns response containing label' do
+        request
+        expect(response.body).to include(label.name)
```
MrSerth, line 31: Depending on how you proceed with the tests, this expectation could also be reworked to be more specific / the selector be more strict (rather tan just checking the whole `body`).

spec/controllers/tasks_controller_spec.rb
```diff
@@ -154,6 +155,15 @@
         expect(assigns(:task).license).to match(valid_params[:license_id])
       end
 
+      it 'sets labels correctly' do
+        post_request
+        expect(assigns(:task).labels.map(&:name)).to eq(valid_params[:label_names])
```
MrSerth, line 160: The `eq` matcher will care about the order of elements in this array. If you ensure a stable sorting: Keep it. If you don't, you might also switch to another array matcher as outlined [here](https://medium.com/beambenefits/array-matchers-in-rspec-42b370e3983e).

spec/controllers/tasks_controller_spec.rb
```diff
@@ -252,6 +278,28 @@
         expect { put_update }.to change { task.reload.license.name }.to('new_license')
       end
 
+      it 'updates the tasks labels' do
+        expect { put_update }.to change { task.reload.labels }.from([existing_label]).to([new_label])
+      end
+
+      context 'when requesting new label to be created' do
```
MrSerth, line 285:
```suggestion
      context 'when requesting a new label to be created' do
```

spec/controllers/tasks_controller_spec.rb
```diff
@@ -252,6 +278,28 @@
...
+        let(:changed_attributes) { {label_names: [not_existing_label_name]} }
+        let(:not_existing_label_name) { 'some new label' }
+
+        it 'creates new task label' do
+          expect { put_update }.to change(Label, :count).by(1)
+        end
+
+        it 'creates a label with the correct name' do
+          put_update
+          expect(Label.last.name).to eql not_existing_label_name
```
MrSerth, line 295: I am still not a fan of the `Label.last` usage in tests. What if we search for the label by its name and expect that only one result is returned? Something like:

```
          expect(Label.where(name: not_existing_label_name).size).to be 1
```

spec/controllers/tasks_controller_spec.rb
```diff
@@ -252,6 +278,28 @@
...
+      it 'updates the tasks labels' do
+        expect { put_update }.to change { task.reload.labels }.from([existing_label]).to([new_label])
+      end
+
+      context 'when requesting new label to be created' do
+        let(:changed_attributes) { {label_names: [not_existing_label_name]} }
+        let(:not_existing_label_name) { 'some new label' }
+
+        it 'creates new task label' do
+          expect { put_update }.to change(Label, :count).by(1)
```
MrSerth, line 290: We should also add a test that reuses an existing label and ensure that the overall count isn't changed.

MrSerth (commented):
Well done! Here are a few more comments, but noting major. During my review, I focussed on the new parts not present in your previous PR #898. 

Probably, it would be helpful to rebase this branch on the other one and target your previous branch, so that the diff is easier to review. Since this rebasing is not easy, I would suggest doing it together; until then, you might just focus on my comments.

app/assets/javascripts/labels.coffee
```diff
@@ -1,35 +1,74 @@
...
+      $template = $('<span></span>').text(text);
+      $template.css({"font-size" : "80%", "font-weight" : "bold"});
+      return $template;
+
+    templateResult: (data) ->
+      if (data.loading)
+        return data.text;
+
+      text = data.text;
+      bg_color = "#" + (data.label_color || "e4e4e4");
```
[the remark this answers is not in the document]
Mathis-Z: good idea, I will do that.

app/assets/javascripts/labels.coffee
```diff
@@ -1,35 +1,74 @@
...
     createTag: (params) ->
       term = $.trim(params.term)
-      if term == ''
-        return null
-      {
-        id: term
-        text: term
+      selection = $('#task_label_names').select2('data').map (element) -> element.id
+
+      if term == '' || term.length > 15 || selection.includes(term)
```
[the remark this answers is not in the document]
Mathis-Z: I will extract it into a variable and also set the select2 maximumInputLength to this value to give the user a better error message.

app/controllers/labels_controller.rb
Mathis-Z: Yes, that is true.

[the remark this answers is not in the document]
Mathis-Z: Just saw your other comment further down so I will change it to paginated.next_page.present? now. :)

app/controllers/labels_controller.rb
Mathis-Z: We could probably just remove the function and have rails give out an error if no page parameter is given. I did not feel comfortable have a missing parameter lead to an error but that is probably okay. 
Maybe it is even better not having the function because an error is better than silently correcting something without the caller knowing.

app/models/task.rb
Mathis-Z: Since users can always create new labels when needed it is probably best to delete the unused ones.

spec/controllers/labels_controller_spec.rb
Mathis-Z: I was not aware of the redirect_to() expectation but that is much cleaner of course.

spec/controllers/labels_controller_spec.rb
Mathis-Z: Yes, that works.

spec/controllers/labels_controller_spec.rb
Mathis-Z: I changed the expectations to check for keys and variable types.

spec/controllers/labels_controller_spec.rb
Mathis-Z: I changed that.

spec/controllers/tasks_controller_spec.rb
Mathis-Z: i switched to contain_exactly because the order is not ensured to stay the same.

spec/controllers/tasks_controller_spec.rb
Mathis-Z: Thats probably better.

spec/controllers/tasks_controller_spec.rb
Mathis-Z: I added a test for that.

app/controllers/labels_controller.rb
Mathis-Z: I just went ahead and removed the function but I am also happy to re-add it if you want.

MrSerth: I think, it's fine the way it is. If no page parameter is given, WillPaginate will automatically assume page 1, similar to the method you previously added.

app/models/task.rb
MrSerth: Yes, this probably would be desirable. The only downside I can think of is a change in the color -- since every new label gets a new random color.

app/assets/javascripts/labels.coffee
```diff
@@ -1,35 +1,74 @@
...
+        query = {
+          search: params.term,
+          page: params.page || 1
+        }
+        return query;
+    }
+
+    templateSelection: (data, container) ->
+      text = data.text;
+      bg_color = "#" + ($(data.element).attr("label_color") || data.label_color || "e4e4e4");
```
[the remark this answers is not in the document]
MrSerth: Ah, I see; got that -- thanks. What a pity that select2 doesn't parse the data attributes of the `<option>` nodes. Then, I am fine with going for this option. Just a final thought on this topic: If we switch the order, we might improve performance: Initializing jQuery for a given element is more "expensive" than accessing an attribute of an object. Thus, if we first check for `data.label_color`, and only in the second step for `$(data.element).attr("label_color")`, we can slightly reduce the overhead of JavaScript operations (and it's not harmful anyway, I think).

app/models/task.rb
kkoehn: We could try to generate the color from the string. Either something like
`Digest::MD5.hexdigest(label)[0..5]`


alternativly we use a hash to generate a reproduceable number and pick a color from the COLORS-array in the label-model (that way we would have more control of the available colors)
`Digest::MD5.hexdigest(label).to_i(16) % 50`

spec/controllers/labels_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        expect(response.parsed_body['pagination']).to have_key('more')
+
+        expect(response.parsed_body['results']).to be_an_instance_of(Array)
+        expect(response.parsed_body['pagination']['more']).to be_in([true, false])
+      end
+
+      it 'returns response containing label' do
+        get_request
+        expect(response.parsed_body['results']).to include(a_hash_including('text' => label.name))
+      end
```
kkoehn, line 39: I would prefer a spec that gives an easy to read example of how the resulting json would look:

```suggestion
      it 'returns valid json' do
        get_request

        expect(JSON.parse(response.body, symbolize_names: true)).to eql(
          {
            pagination: {more: false},
            results: [{id: label.name, label_color: label.color, label_font_color: label.font_color, text: label.name}],
          }
        )
     end
```

kkoehn (commented):
Great progress - I found one more point in the specs

app/models/task.rb
Mathis-Z: I went with the second solution because it sticks to the pre-defined colors.

spec/controllers/labels_controller_spec.rb
Mathis-Z: Thanks for the suggestion. I switched to your solution.

MrSerth:
Ah, thanks for noticing this behavior!

Indeed, using an `after_initialize` filter would always overwrite the color persisted. Therefore, I would rather go for the following:

```rb
  validates :color, presence: true, format: { with: /\A[a-fA-F0-9]{6}\z/ }

  before_validation :choose_label_color, if: -> { color.blank? }
```

This adds a validator to ensure we always have a color and a conditional hook to add a color if not present. Still, it doesn't fix the underlying issue (but "just" prevents we are returning a 500 server error). In order to show the newly added (but not yet saved) label correctly, you should change

```rb
@task.labels.pluck(:name))
```

to 

```rb
@task.labels.map(&:name))
```

in `app/views/tasks/_form.html.slim`. The difference is that `pluck` only gets the name of persisted objects, which is not the case for the newly created label. Let's give this a try and also add a test ;).

spec/controllers/tasks_controller_spec.rb
```diff
@@ -233,6 +233,17 @@
         post_request
         expect(response).to render_template('new')
       end
+
+      context 'when new label needs to be created' do
+        let(:invalid_attributes) { {title: '', label_names: [not_existing_label_name]} }
+        let(:not_existing_label_name) { 'some new label' }
+
+        it "re-renders the 'new' template successfully" do
```
MrSerth, line 241: In this test, we should also check that the `not_existing_label_name` is present and *checked* (Hint: Write a test that fails without the `@task.labels.map(&:name))` change, but only succeeds otherwise. If you test this manually, you'll quickly notice the difference in the HTML -- that is the `selected` attribute for this `option`).

Mathis-Z: I think that would be a good integration test. Something like this:
```
expect(rendered).to have_selector('option[value="some new label"][selected="selected"]')
```

Other than this I found rails-dom-testing to be in the Gemfile but could not find any usages in CodeHarbor and also no examples of how to use it online.
The last option I could think of is using regex again although that is probably not a good idea.

MrSerth: Indeed, I would also go for an integration test. Let's use this as an example and follow up on it next week ;).

The `rails-dom-testing` is only included as an indirect dependency in the `Gemfile.lock`, and therefore not (directly) used so far.

MrSerth (approved):
Awesome, thanks for taking care of the test as well and adding these nice features! 🎷

PR #922, opened 2023-06-13
Add integration test setup and authentication spec

MrSerth:
As we discussed the integration specs earlier, I just went ahead and added the test setup. Feel free to take a quick look. However, most of the parts are really test setup code that I just copied over -- so don't spend too much time on that 😇 . The most interesting part I am eager to hear your feedback about is the actual authentication spec.


Mathis-Z:
Looks good to me. You might also test that logged in users can access the task page and similarly that logged out users cannot access any page but the root page. Also non-root users should not be able to access any admin functionality.
I am not sure if it makes sense to test that though because we would basically test if CanCanCan works.

MrSerth:
Thanks for taking a look! 

> You might also test that logged in users can access the task page and similarly that logged out users cannot access any page but the root page. 

Some of those tests are already existing, and are more suited to be written as a traditional unit test. For example, we already check that regular users currently logged in can access the task page:

https://github.com/openHPI/codeharbor/blob/bf88036a0aae07df29619a442c9bde417eb04e4f/spec/controllers/tasks_controller_spec.rb#L23-L26

Hence, I just added another test statement that works for users not signed in (any longer):

https://github.com/openHPI/codeharbor/blob/bf88036a0aae07df29619a442c9bde417eb04e4f/spec/controllers/tasks_controller_spec.rb#L78-L86

> Also non-root users should not be able to access any admin functionality.

That's true. For the majority of the code, this functionality is already validated with the ability specs in the respective model. However, the RailsAdmin dashboard (and subsequent pages) is not directly tested through this behavior, so that I created a new integration test with 7c4fcd79.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -74,6 +74,15 @@
         end
       end
     end
+
+    context 'when no user is signed in' do
+      before { sign_out user }
+
+      it 'redirects to root page' do
+        get_request
+        expect(response).to redirect_to(root_path)
```
florian-str, line 83: I would also consider checking the flash message.

As a user, it is important to know, the reason behind a "failure". For that reason, the existence of the message is essential.

florian-str (commented):
Looks good as far as I'm concerned.

I would maybe add the check for flash messages, as commented above.

spec/controllers/tasks_controller_spec.rb
MrSerth: Sure, this makes sense here as well. I just added a corresponding test case.

MrSerth:
> Looks good as far as I'm concerned.
> 
> I would maybe add the check for flash messages, as commented above.

Thanks for your feedback as well!

Mathis-Z (approved):

florian-str (approved):

Issue #902, opened 2023-05-12
Re-enable duplication of tasks


MrSerth:
- [x] Add a button to the task show, allowing users to duplicate / clone the tasks.
- [x] Each file and all associated models should be copied, if suitable.  
For example, files should be duplicated as well, whereas the license should not. For labels, the relationship needs to be copied, but the labels itself not.
- [x] Update permission for the new task one: 
  - [x] The task should be private
  - [x] If the task was part of a collection or group, only re-add it to the same group / collection if the user has permission to do so (otherwise, don't add it). EDIT: This behaviour has been modified to match the comments below.

Of course, the existing task should remain untouched.

florian-str:
> If the task was part of a collection or group, only re-add it to the same group / collection if the user has permission to do so (otherwise, don't add it).

I think that a duplicated task should not automatically be re-added to the group.

Reasons for this include:

- The task will probably (temporarily) get back into a Work-In-Progress-State, making it less useful in the group-list.
- In a group with many users, each making minor changes to better suit their specific class, will result in a cluttered group with many similar tasks.
- A teacher will in many cases only change minor details, to make the task better suited for his/ her specific class. Therefor, the additional value of the modified task is probably low.

As it is more likely, that a "private" modification is done, it seems more useful to go for an opt-in rather than opt-out solution.

MrSerth:
> As it is more likely, that a "private" modification is done, it seems more useful to go for an opt-in rather than opt-out solution.

Alright, fine for me -- you're right, we might otherwise clutter the group with too many private copies. Let's just ensure we communicate clearly to the users what happened (and inform them that the task they duplicated is no longer part of any group / collection).

florian-str:
As a corresponding PR (#924) was merged, this issue can be closed.

PR #924, opened 2023-06-15
Implements #902: Duplication of tasks

florian-str:
Implements #902: Duplication of tasks

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -21,6 +21,7 @@ class TasksController < ApplicationController
     new_entry.groups = []
     new_entry.collections = []
     new_entry.access_level_private!
+    new_entry.title = "#{t('tasks.copy_of_task')} ##{@task.id}: #{new_entry.title}"
     if new_entry.save(context: :force_validations)
       redirect_to new_entry, notice: t('tasks.notification.created')
     else
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -2,7 +2,7 @@
 
 require 'zip'
 
-class TasksController < ApplicationController
+class TasksController < ApplicationController # rubocop:disable Metrics/ClassLength
   load_and_authorize_resource except: %i[import_external import_uuid_check]
 
   before_action :handle_search_params, only: :index
@@ -16,12 +16,7 @@ class TasksController < ApplicationController
   end
 
   def duplicate
-    new_entry = @task.duplicate
-    new_entry.user = current_user
-    new_entry.groups = []
-    new_entry.collections = []
-    new_entry.access_level_private!
-    new_entry.title = "#{t('tasks.copy_of_task')} ##{@task.id}: #{new_entry.title}"
+    new_entry = @task.clean_duplicate(current_user)
     if new_entry.save(context: :force_validations)
       redirect_to new_entry, notice: t('tasks.notification.created')
     else
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -20,7 +20,7 @@ class TasksController < ApplicationController
     new_entry.user = current_user
     new_entry.groups = []
     new_entry.collections = []
-    # new_entry.access_level = 'private' Activate after rebase
+    new_entry.access_level_private!
     if new_entry.save(context: :force_validations)
       redirect_to new_entry, notice: t('tasks.notification.created')
     else
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -15,6 +15,16 @@ class TasksController < ApplicationController
     @tasks = @search.result(distinct: true).paginate(per_page: 5, page:)
   end
 
+  def duplicate
+    new_entry = @task.duplicate
+    new_entry.user = current_user
+    if new_entry.save(context: :force_validations)
+      redirect_to new_entry, notice: t('tasks.notification.created')
+    else
+      redirect_to @task, alert: t('tasks.notification.duplicate_failed')
+    end
+  end
+
   def show
     @files = @task.files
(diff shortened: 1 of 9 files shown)


app/controllers/tasks_controller.rb
```diff
@@ -15,6 +15,15 @@ def index
     @tasks = @search.result(distinct: true).paginate(per_page: 5, page:)
   end
 
+  def duplicate
+    new_entry = @task.clean_duplicate(current_user)
+    if new_entry.save(context: :force_validations)
+      redirect_to new_entry, notice: t('tasks.notification.created')
+    else
+      redirect_to @task, alert: t('tasks.notification.duplicate_failed')
```
florian-str, line 23: This case should not be able to happen, however I don't know how to test this.

The issue is, that you would need to create an invalid task to duplicate it, which seems to be impossible in RSpec.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -404,6 +404,48 @@
...
+    let!(:task) { create(:task, valid_attributes) }
+    let(:valid_attributes) do
+      {
+        user:,
+        title: 'title',
+        license: create(:license, name: 'license'),
+        groups: [group],
+        collections: [collection],
+      }
+    end
```
MrSerth, line 422: I haven't tested the suggestion, but you might be able to leverage the `build_list` feature here:

```suggestion
    let(:groups) { create_list(:group, 1) }
    let(:collections) { create_list(:collection, 1) }

    let!(:task) { create(:task, valid_attributes) }
    let(:valid_attributes) do
      {
        user:,
        title: 'title',
        license: create(:license, name: 'license'),
        groups:,
        collections:,
      }
    end
```

spec/controllers/tasks_controller_spec.rb
```diff
@@ -404,6 +404,48 @@
...
+
+    it 'resets the groups' do
+      post_duplicate
+      expect(Task.last.groups).to eq([])
+    end
+
+    it 'resets the collections' do
+      post_duplicate
+      expect(Task.last.collections).to eq([])
+    end
```
MrSerth, line 446: As discussed, I am not a fan of the `Task.last` statement (even though used already), since this will change with another primary key.

spec/models/task_spec.rb
```diff
@@ -237,6 +248,28 @@
...
+      it 'is private' do
+        expect(clean_duplicate.access_level).to eq(:private.to_s)
+      end
+
+      it 'is not part of a group' do
+        expect(clean_duplicate.groups).to eq([])
+      end
+
+      it 'has a modified title' do
+        expect(clean_duplicate.title).to eq("#{I18n.t('tasks.copy_of_task')}: #{task.title}")
```
MrSerth, line 270: This actually resembles the implementation very much. In this case, I would actually suggest having a static string defined for the expected way. Otherwise, we might miss some errors in the future.

MrSerth (commented):
Nice and well done, thank you! I just have a few minor comments 🏕️.

app/controllers/tasks_controller.rb
MrSerth: As discussed, you might be able to test it with a mock, such as

```rb
let(:invalid_task) { Task.new }
// ...
allow(task).to receive(:clean_duplicate).with(user).and_return(invalid_task)
```

florian-str: For some reason, this does not work as expected.

I have the option to use `any_instance_of(Task)`, but rubocop does not like that solution.

Do you have an alternative idea?

spec/controllers/tasks_controller_spec.rb
florian-str: Yes, it does indeed work. I'm going to upload it alongside the other suggestions.

spec/controllers/tasks_controller_spec.rb
florian-str: It now implemented and is going to be uploaded alongside the other suggestions.

spec/models/task_spec.rb
florian-str: You're right. I have made the changes.

app/controllers/tasks_controller.rb
MrSerth: Indeed, it doesn't work as expected. The main reason is that we are specifically stubbing one method on one object. However, the controller will create a new object from the database (mirroring the same underlying task), but hence our stubbing no longer works. We can, however, instruct the controller to find exactly the same object:

```rb
context 'when saving fails' do
  let(:invalid_task) { Task.new }
  
  before do
    # We need to stub the find method, because otherwise another object is returned from the database
    allow(Task).to receive(:find).with(task.id.to_s).and_return(task)
    # By design, we return an invalid task, which will fail to save
    allow(task).to receive(:clean_duplicate).with(user).and_return(Task.new)
  end

  it 'shows an error message' do
    post_duplicate
    expect(flash[:alert]).to eq(I18n.t('tasks.notification.duplicate_failed'))
  end
end
```

florian-str: It makes sense. Thank you for your help.

MrSerth (approved):
Well done! I just tested your changes locally again and found a minor issue that I fixed. The problem was related to the button you added and the font it got:

Bildschirmfoto 2023-06-27 um 21 40 05

As you see, the font is different compared to all other buttons. I fixed this:

Bildschirmfoto 2023-06-27 um 21 40 12

The issue was that the text indention was not correct. You made the text a child of the icon (where FontAwesome is applied) rather than having it on the same indention level:

Unintended side effect:
```slim
i.fa.fa-clone
  =< t('tasks.show.duplicate')
```

Desired:
```slim
i.fa.fa-clone
=< t('tasks.show.duplicate')
```

PR #938, opened 2023-07-01
Re-enabling testing frameworks for tests

### LINKED ISSUE #895: Fix References to Testing Framework
We still have references to the `testing_framework` in our codebase.

Let's:
- [ ] Investigate where the testing framework was / is used
- [ ] Check whether we should fix the usage (and re-enable it) or drop it

<hr>

Sentry Issue: [CODEHARBOR-16](https://codeocean.sentry.io/issues/4126397356/?referrer=github_integration)

```
PG::UndefinedColumn: ERROR:  column tests.testing_framework_id does not exist (PG::UndefinedColumn)
LINE 1: SELECT "tests".* FROM "tests" WHERE "tests"."testing_framewo...
                                            ^

  activerecord (7.0.4.3) lib/active_record/connection_adapters/postgresql_adapter.rb:837:in `prepare'
  activerecord (7.0.4.3) lib/active_record/connection_adapters/postgresql_adapter.rb:837:in `block in prepare_statement'
  activesupport (7.0.4.3) lib/active_support/concurrency/load_interlock_aware_monitor.rb:25:in `handle_interrupt'
  activesupport (7.0.4.3) lib/active_support/concurrency/load_interlock_aware_monitor.rb:25:in `block in synchronize'
  activesupport (7.0.4.3) lib/active_support/concurrency/load_interlock_aware_monitor.rb:21:in `handle_interrupt'
...
(186 additional frame(s) were not displayed)

ActionView::Template::Error: PG::UndefinedColumn: ERROR:  column tests.testing_framework_id does not exist (ActionView::Template::Error)
LINE 1: SELECT "tests".* FROM "tests" WHERE "tests"."testing_framewo...
                                            ^

ActionView::Template::Error: PG::UndefinedColumn: ERROR:  column tests.testing_framework_id does not exist (ActionView::Template::Error)
LINE 1: SELECT "tests".* FROM "tests" WHERE "tests"."testing_framewo...
                                            ^
```

Mathis-Z:
I thought about adding the testing framework info to the exported proforma xml but apparently proforma tests do not have any field where something like a testing framework would fit in.
There is test-configuration > test-meta-data but as far as I understand it this is meant for information on how to run the test?
Should we omit this or just add it next to the "title", "description", "internal description", ... fields?

Fixes #895

### DIFF
--- a/spec/controllers/tasks_controller_spec.rb
+++ b/spec/controllers/tasks_controller_spec.rb
@@ -397,14 +397,14 @@ RSpec.describe TasksController do
 
       context 'when task has a test' do
         subject(:put_update) do
-          put :update, params: {id: task.to_param, task: changed_attributes.merge({'tests_attributes' => tests_attributes})}
+          put :update, params: {id: task.to_param, task: changed_attributes.merge(tests_attributes:)}
         end
 
         let(:test) { build(:test) }
         let(:new_testing_framework) { create(:testing_framework) }
         let!(:task) { create(:task, valid_attributes.merge(tests: [test])) }
 
-        let(:tests_attributes) { {'0' => test.attributes.merge('title' => 'new_test_title', 'testing_framework_id' => new_testing_framework.id)} }
+        let(:tests_attributes) { {'0': test.attributes.symbolize_keys.merge(title: 'new_test_title', testing_framework_id: new_testing_framework.id)} }
 
         it 'updates the requested task' do
           expect { put_update }.to change { task.reload.title }.to('new_title')
--- a/spec/controllers/tasks_controller_spec.rb
+++ b/spec/controllers/tasks_controller_spec.rb
@@ -164,6 +164,16 @@ RSpec.describe TasksController do
         get_request
         expect(assigns(:tests)).to include(test)
       end
+
+      context 'when test has a framework' do
+        let(:testing_framework) { create(:testing_framework) }
+        let(:test) { create(:test, task:, testing_framework:) }
+
+        it "includes the frameworks's name and version in response" do
+          get_request
+          expect(response.body).to include(testing_framework.name_with_version)
+        end
+      end
     end
   end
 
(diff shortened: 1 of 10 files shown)


MrSerth (commented):
Nice, that was a home run 🏃 

Still, may I ask you to take care of the following three bits as well?:

- A test for the new code would be awesome
- By doing so, you'll probably already take care of this one: Adding a factory for the testing framework (and linking it to at least one existing test).
- Ensuring that the seed data provided to developers contains one example with a testing framework.

> There is test-configuration > test-meta-data but as far as I understand it this is meant for information on how to run the test?

Indeed, or it could specify restrictions (such as memory or time constraints) that are applicable to the test execution. Those information are vendor-specific, as far as I know, but would also allow us to host anything we want. At the current point, I would actually leave it in the current state (without adding it to the ProFormA XML export). If you'd like, we could open a follow-up ticket for this to be added later :+1:

spec/controllers/tasks_controller_spec.rb
```diff
@@ -391,17 +401,22 @@
         end
 
         let(:test) { build(:test) }
+        let(:new_testing_framework) { create(:testing_framework) }
         let!(:task) { create(:task, valid_attributes.merge(tests: [test])) }
 
-        let(:tests_attributes) { {'0' => test.attributes.merge('title' => 'new_test_title')} }
+        let(:tests_attributes) { {'0' => test.attributes.merge('title' => 'new_test_title', 'testing_framework_id' => new_testing_framework.id)} }
```
MrSerth, line 407: Is there any reason you are using the old rocket-styled syntax for the has? What about using `testing_framework_id: new_testing_framework.id` (and similar for the others)?

MrSerth (commented):
Wohoo, we're about to get this over the finish line as well (that was quickly!) 🪂

spec/controllers/tasks_controller_spec.rb
Mathis-Z: Originally I just wanted the syntax to be consistent with the already existing code but after your comment I tried to use the ':' syntax instead and found that the tests actually fail when you use 
```'title': 'new_test_title'``` instead of ```'title' => 'new_test_title'```.
You can see why when you output test.attributes.merge('title' => 'new_test_title') in the console:
```
{"id"=>1, "created_at"=>..., "title"=>"title", "description"=>..., :title=>"new_test_title"}
```
So apparently merge() converts strings to symbols when you use the ':' syntax but not when you use the rocket syntax. The same happens with double-quoted strings as keys.
I decided to work around that with symbolize_keys and I also changed the rest of the context to use the new hash syntax.

MrSerth: Ah, that sounds plausible. But thanks for changing, I think it looks cleaner (even though the additional `symbolize_keys` call is required).

MrSerth (approved):

PR #959, opened 2023-08-02
implement support for proforma 0.9

kkoehn:
add support for (custom) test-types


MrSerth (commented):
Looks good so far, thanks! I was just wondering whether we should show the test configuration somehow, or whether we omit it by design (since it cannot be edited). One approach could be to display it in the view similar to the metadata (which we have as read-only JSON as well), potentially collapsed.

MrSerth (approved):
Awesome, that's great! And now we also have a read-only view of the test configuration, as I was able to confirm 💪:

Bildschirmfoto 2023-08-17 um 00 31 41

PR #1005, opened 2023-09-23
Making task show route accessible to guest users

Mathis-Z:
This allows users to view and download tasks without signing in. The idea is that OAI-PMH harvesters can present their users with working links for the records they fetched from CodeHarbor. 

The show.html.slim does get harder to read with this but I could not find a better approach.
I am looking forward to your feedback :)

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -4,6 +4,9 @@ class Ability
   include CanCan::Ability
 
   def initialize(user)
+    # abilities for logged out users/guests
+    guest_abilities
+
     return unless user
 
     alias_action :create, :show, :update, :destroy, to: :crud
@@ -38,6 +41,12 @@ class Ability
 
     # Label
     label_abilities
+  end
+
+  def guest_abilities
+    can %i[show download], Task, &:access_level_public?
+
+    can %i[show read], Comment
   end
 
   def admin_abilities(user)
--- a/app/assets/stylesheets/exercises.scss
+++ b/app/assets/stylesheets/exercises.scss
@@ -338,7 +338,7 @@ legend {
     width: 100%;
     text-align: center;
     display: block;
-    .btn-light,
+    .btn,
     .btn-group {
       width: 100%;
       margin-bottom: 5px;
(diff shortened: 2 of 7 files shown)


app/models/ability.rb
```diff
@@ -40,6 +43,12 @@ def initialize(user)
     label_abilities
   end
 
+  def guest_abilities
+    can %i[show download], Task, &:access_level_public?
+
+    can %i[show read], Comment
```
MrSerth, line 49: Since comments are loaded asynchronously through JavaScript, it seems quite simple to change the URL and access some other comments. Can / should we include a check for the access level of the task as well? For example, we could decide that comments can only be read if the task is public.

MrSerth (commented):
Well done, I like it pretty much and think that the tooltip / disabled buttons are also a great hint 👍

app/models/ability.rb
Mathis-Z: Yes, that is true. Actually the same problem exists with users that are logged in but are not allowed to access a certain task. For guest_abilities something like this works:
```
can %i[show read], Comment, task: {access_level: :public}
```

However this does not work for limiting access for users that are logged in because the access conditions in this case are more complex (is in same group?, is admin?, ...) and cannot be expressed as CanCanCan hash conditions. I found another solution though which authorizes access to comments depending on whether or not the user can access the comments task. 
https://github.com/CanCanCommunity/cancancan/blob/develop/docs/nested_resources.md

MrSerth: > Actually the same problem exists with users that are logged in but are not allowed to access a certain task.

Indeed, I just noticed that as well.

> I found another solution though which authorizes access to comments depending on whether or not the user can access the comments task.

This looks perfect and is similar to the authorization concept we use in CodeOcean for nested resources. There, we don't have CanCanCan but another library, but nevertheless we are also checking the parent resource and then the actual object. 

Many thanks for these changes and the awesome test coverage you added! 👏

MrSerth (approved):
Awesome, thank you! I think that these changes are absolutely fine, even though you argued that the `show.html.slim` gets a little more difficult to read. That is, indeed, true, but nothing we can change right now. In the long term, one could separate those actions into components and then have each component just return the correct representation (for guests and authenticated users). Maybe one day 😉

PR #1015, opened 2023-10-02
Styling users pages

zuhrasofyan:
Fix styling for forms in the various users views page


MrSerth (approved):
Perfect -- well done. Thank you!

I was testing all pages locally and compared those with the current versions; that's a really nice fix for all of them.

PR #1028, opened 2023-10-11
Adding resumptionTokens to the OAI-PMH endpoint

### LINKED ISSUE #1004: Implement OAI-PMH resumption tokens

Mathis-Z:
This closes #1004.

I also fixed two minor unrelated things that I noticed. I am looking forward to your feedback :)

### DIFF
--- a/app/controllers/bridges/oai/oai_controller.rb
+++ b/app/controllers/bridges/oai/oai_controller.rb
@@ -5,12 +5,15 @@ module Bridges
     # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/ClassLength
     class OaiController < ActionController::API
       SUPPORTED_METADATA_PREFIXES = %w[oai_dc lom].freeze
+      MAX_RECORDS_PER_RESPONSE = 100
       ERROR_RESPONSE_XSD = 'vendor/assets/schemas/oai-pmh/error_response_combined.xsd'
       SUCCESSFUL_RESPONSE_XSD = 'vendor/assets/schemas/oai-pmh/successful_response_combined.xsd'
 
       def handle_request
         builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
-          verb = oai_params[:verb]
+          parse_parameters!(params)
+
+          verb = @oai_params[:verb]
 
           raise OaiError.new('Missing verb parameter', 'badVerb') if verb.nil?
 
@@ -40,8 +43,8 @@ module Bridges
       end
 
       def handle_get_record(xml)
-        task = find_record!(oai_params)
-        metadata_prefix = parse_metadata_prefix!(oai_params[:metadataPrefix])
+        task = find_record!(@oai_params)
+        metadata_prefix = parse_metadata_prefix!(@oai_params[:metadataPrefix])
 
         xml.GetRecord do
           export_record(xml, task, metadata_prefix)
@@ -62,19 +65,21 @@ module Bridges
       end
 
       def handle_list_identifiers(xml)
-        parse_metadata_prefix!(oai_params[:metadataPrefix])
+        parse_metadata_prefix!(@oai_params[:metadataPrefix])
 
         xml.ListIdentifiers do
-          find_records!(oai_params).each do |task|
-            export_record_header(xml, task)
+          records = find_records!(@oai_params, @resumption_params)
 
-            # TODO: resumptionTokens for large answers
+          records.first(MAX_RECORDS_PER_RESPONSE).each do |task|
+            export_record_header(xml, task)
           end
+
+          export_resumption_token(xml, records)
         end
       end
 
       def handle_list_metadata_formats(xml)
-        find_record!(oai_params) if params[:identifier].present?
+        find_record!(@oai_params) if @oai_params[:identifier].present?
 
         xml.ListMetadataFormats do
           xml.metadataFormat do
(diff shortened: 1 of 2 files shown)


MrSerth (commented):
Awesome, this is really great! 👏 

I tested the implementation with two validators and with our beloved edu-sharing instance and all three were able to consume the data as expected and without any error. That's awesome!

Below are two minor remarks. Otherwise, I have two additional ones:
- For a brief moment, I was considering whether we should add the `cursor` and `completeListSize` to the response. The `completeListSize` would, of course, require another `.size` call, but getting the number in line 212 (before applying the `.limit`) would do the trick. Then, we still have the cursor, which we could just start with 0 in the first request, and add the number of records to the base64-encoded resumptionToken. On the next request, we just use this number as a cursor, and add the sum of this plus the number of newly returned records again to the base64-encoding... I don't know whether it's worth the effort, but probably we should either go for both or none...
- Whenever a resumptionToken is used, we get a warning of unpermitted parameters: `Unpermitted parameters: :last_id, :ts_from, :ts_until. Context: {  }`. Can you permit those, so that the warning disappears, please?

app/controllers/bridges/oai/oai_controller.rb
```diff
@@ -172,6 +179,20 @@ def export_record(xml, task, format)
...
+        if records.size > MAX_RECORDS_PER_RESPONSE
+          last_returned_record = records[MAX_RECORDS_PER_RESPONSE - 1]
+
+          ts_from = last_returned_record.updated_at.strftime('%Y-%m-%dT%H:%M:%S.%NZ')
+          ts_until = @resumption_params.present? ? @resumption_params[:ts_until] : Time.zone.now.strftime('%Y-%m-%dT%H:%M:%S.%NZ')
+          token = @oai_params.merge({last_id: last_returned_record.id, ts_from:, ts_until:})
+
+          xml.resumptionToken Base64.encode64(token.to_json)
+        elsif @resumption_params.present?
+          xml.resumptionToken
```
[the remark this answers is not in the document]
Mathis-Z: That is true. I added your proposed comment.

app/controllers/bridges/oai/oai_controller.rb
```diff
@@ -172,6 +179,28 @@ def export_record(xml, task, format)
...
+          new_token = @oai_params.merge(
+            {
+              last_id: returned_records.last.id,
+              ts_from: returned_records.last.updated_at.strftime('%Y-%m-%dT%H:%M:%S.%NZ'),
+              ts_until: @resumption_params.present? ? @resumption_params[:ts_until] : Time.zone.now.strftime('%Y-%m-%dT%H:%M:%S.%NZ'),
+              cursor: cursor + returned_records.size,
+            }
+          )
+
+          xml.resumptionToken Base64.encode64(new_token.to_json), cursor:, completeListSize: complete_list_size
```
MrSerth, line 197: Currently, this encoding adds line breaks. We can avoid so when using another method:

```suggestion
          xml.resumptionToken Base64.urlsafe_encode64(new_token.to_json), cursor:, completeListSize: complete_list_size
```

app/controllers/bridges/oai/oai_controller.rb
```diff
@@ -200,41 +235,68 @@ def filter_tasks_by_set(tasks, set_spec)
...
-        tuntil = nil
-
-        if params[:from]
+      def parse_parameters!(params)
+        if params.key? :resumptionToken
           begin
-            tfrom = DateTime.iso8601(params[:from])
-          rescue Date::Error
-            raise OaiError.new("The specified time '#{params[:from]}' is not in iso8601 format", 'badArgument')
+            params = ActionController::Parameters.new(JSON.parse(Base64.decode64(params[:resumptionToken])))
```
MrSerth, line 249: Similar to above, we should probably parse without linefeeds and URL safe:

```suggestion
            params = ActionController::Parameters.new(JSON.parse(Base64.urlsafe_decode64(params[:resumptionToken])))
```

app/controllers/bridges/oai/oai_controller.rb
```diff
@@ -172,6 +179,20 @@ def export_record(xml, task, format)
         end
       end
 
+      def export_resumption_token(xml, records)
+        if records.size > MAX_RECORDS_PER_RESPONSE
+          last_returned_record = records[MAX_RECORDS_PER_RESPONSE - 1]
```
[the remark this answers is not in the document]
MrSerth: Ah, I see. Yes, that's indeed better -- thanks!

app/controllers/bridges/oai/oai_controller.rb
MrSerth, line 249: Whenever a resumptionToken is used, we get a warning of unpermitted parameters: `Unpermitted parameters: :last_id, :ts_from, :ts_until. Context: {  }`.

Can you permit those, so that the warning disappears, please?

MrSerth (commented):
Thanks for adding the `cursor` and `completeListSize`.

I just have two minor comments about the base64 encoding of the `resumptionToken`, which will also avoid line breaks and special URL characters, in addition to the unpermitted parameters. 

These should be the last comments, I think 🤞

app/controllers/bridges/oai/oai_controller.rb
Mathis-Z: That is a good catch. I changed that but since this had to be changed in the tests as well I created an own commit. Thanks!

app/controllers/bridges/oai/oai_controller.rb
Mathis-Z: Yes, I fixed that now.

MrSerth (approved):
Awesome, thank you!

Issue #1021, opened 2023-10-09
Clean and generalize locales


MrSerth:
Our locales (i.e., those defined in `config/locales/en.yml`) grew over the years. There are some translations not used any longer, and other strings are multiple defined (such as for "closing" or similar).

Therefore, we should clean the locales and also reuse the concept we have with CodeOcean. This allows us to reduce the size of the locale file, sort it better and reduce redundancy. Further, it allows us to add dynamic titles to CodeHarbor.

PR #1069, opened 2023-11-05
Enable dynamic page titles

MrSerth:
Thanks to the work performed in #1021, we can now enable dynamic page titles for CodeHarbor. This is done through a mechanism already used in CodeOcean (the code is mainly copied over). While the mechanism is not that easy to read (I agree!), it is relatively complete and works in many edge cases.

Unfortunately, we have to repeat some Devise-related translations. Otherwise, the Devise views would break in case of validation errors.

### DIFF
--- a/app/views/application/_breadcrumbs_and_title.html.slim
+++ b/app/views/application/_breadcrumbs_and_title.html.slim
@@ -26,7 +26,8 @@
   - active_action = t("breadcrumbs.#{controller_name}.#{params[:action]}")
 
 
-- title = "#{active_action} - #{application_name}"
+- title = "#{application_name}"
+- title = "#{active_action} - #{title}" unless %w[index show].include? params[:action]
 - content_for :breadcrumbs do
   .container.mb-4
     ul.breadcrumb.bg-body-secondary.px-3.py-2
--- a/app/views/application/_breadcrumbs_and_title.html.slim
+++ b/app/views/application/_breadcrumbs_and_title.html.slim
@@ -1,1 +1,50 @@
-- content_for :title, application_name
+- model = controller_path.classify.constantize rescue nil
+- if model
+  - object = model.find_by(id: params[:id])
+  - if (parent_model = model.try(:parent_resource))
+    - parent_route_key = parent_model.model_name.singular_route_key
+    - if params["#{parent_route_key}_id"].present?
+      - parent_object = object.try(parent_route_key) || parent_model.find_by(id: params["#{parent_route_key}_id"])
+      - parent_element = link_to_if(current_user && can?(:show, parent_object), parent_object, {controller: parent_model.model_name.route_key, action: :show, id: parent_object.id})
+      - parent_root_element = link_to_if(current_user && can?(:index, parent_model), parent_model.model_name.human(count: 2), {controller: parent_model.model_name.route_key, action: :index})
+      - root_element = link_to_if(current_user && can?(:index, model), model.model_name.human(count: 2), send(:"#{parent_route_key}_#{model.model_name.route_key}_path", parent_object))
+      - if object
+        - current_element = link_to_if(current_user && can?(:show, object), object, send(:"#{parent_route_key}_#{model.model_name.singular}_path", parent_object, object))
+    - else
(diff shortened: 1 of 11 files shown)


app/views/application/_breadcrumbs_and_title.html.slim
```diff
@@ -1 +1,50 @@
...
+  - else
+    - root_element = link_to_if(current_user && can?(:index, model), model.model_name.human(count: 2), {controller: model.model_name.route_key, action: :index})
+    - if object
+      - current_element = link_to_if(current_user && can?(:show, object), object, {controller: model.model_name.route_key, action: :show, id: object.id})
+  - if I18n.exists?("shared.#{params[:action]}")
+    - active_action = t("shared.#{params[:action]}", model: model&.model_name&.human)
+  - else
+    - active_action = t("#{controller_name}.index.#{params[:action]}", model: model&.model_name&.human)
+- else
+  - active_action = t("breadcrumbs.#{controller_name}.#{params[:action]}")
```
Mathis-Z, line 26: I am just wondering if we should add a test or a default translation in case the required translation is missing? I18n-tasks cannot mark these translations as missing because it only performs static analysis.
Missing keys might be caught by other tests anyway though.

app/views/application/_breadcrumbs_and_title.html.slim
```diff
@@ -1 +1,50 @@
...
+
+
+- title = "#{active_action} - #{application_name}"
```
Mathis-Z, line 29: Personally I think "Tasks - CodeOcean" would read better than "Tasks - Index - CodeOcean". Similarly we might omit the show action because in that case we have the resource name (for example "Example Task - CodeOcean" instead of "Example Task - Show - CodeOcean"). That is just preference though.

Mathis-Z (approved):

app/views/application/_breadcrumbs_and_title.html.slim
MrSerth: We could add a test or default translation, both is possible. However, we would probably need to add that for many of the different translations.

In the current setup (and with the current test coverage), there is no translation missing. I am sure about that, because we raise an exception during any test if a translation is missing:

https://github.com/openHPI/codeharbor/blob/ba46512c2792c626f2654c9c3ebc06ae7bf25dfb/config/environments/test.rb#L61-L62

If not covered by the tests and not noticed during development (I tried clicking through all pages myself as well), we would still render a page in production, but of course not with a nice text.

Do you think we should proceed with any further action here?

app/views/application/_breadcrumbs_and_title.html.slim
MrSerth: That's a valid suggestion, which I implemented with 02c1bc4d.

app/views/application/_breadcrumbs_and_title.html.slim
Mathis-Z: I would have said we do not need to do something further since the tests should cover this but I just stumbled over a missing translation when signing up as a new user: "breadcrumbs.registrations.create"
:)

Mathis-Z: I would have said the test cover this and we do not need to do anything but I just stumbled over a missing translation when signing up as a new user: "breadcrumbs.registrations.create".
:)

MrSerth: Thanks! I tested many aspects of the registration (that's why some keys are already duplicated), but obviously I didn't enter invalid data during the registration (that's how you got the missing key, right?).

Mathis-Z: I also found the missing translation "breadcrumbs.registrations.update" while working on the error message layout. This happens when you edit your password but do not enter the correct old password.

MrSerth: Good finding, I added that as well! 🙏

Mathis-Z: Other cases should get caught by the tests so I will resolve the conversation now.

PR #1089, opened 2023-11-15
Use common defaults for shared/form_errors

MrSerth:
I was wondering whether we could simplify the usage of the `shared/form_errors`, since some information could be extracted automatically. While I like the support for custom arguments, I also think we can remove some explicit reference.

This is just a suggestion and proof-of-concept, but I think it was worth to try it (and think we could also discuss merging it).


Mathis-Z (approved):
I like the changes and I think we should merge them :+1:

Issue #930, opened 2023-06-21
Reimplement contributions for tasks


florian-str:
It should be possible, that a user can create a contribution to a public task by another user.

The workflow is supposed to be the following:

1. user B creates a contribution (T2)
2. B saves the changes -> creates an "PR"
3. An email gets sent to the members of A, which are users with write access to the original task (T1)
4. On the page of T1, there should be a notification indicating a pending contribution
5. A-users are able to approve or reject a task

In case of approval:
- Comments on T2 get dropped
- T1 gets replaced by T2
- T2 gets dropped
- all other pending contributions T3...Tn get rejected/ invalidated
- Emails

In case of rejection or edit of T1:
- T2 will be a "real duplicate"
- Emails

If T2 gets withdrawn:
- B can convert T2 to a real duplicate or select to delete the entry
- Emails

Other constraints:
- Each user may only create one active contribution per task
- Contributions can't be added to groups or collections
- Contributions can't modify the `parent_uuid`, `license` and `access_level`
- Contributions don't have ratings
- Contributions, which have not been answered for a certain (?) duration should be viewed as rejected

This issue could be implemented with the following steps:

- [x] Implementation of the basic workflow
- [x] Adding constraints to the contributions
- [x] Implementing the approval
- [x] Implementing the rejection/ withdraw/ change of original tasks
- [ ] Quality of life improvements (e.g. showing changed files)

MrSerth:
We've implemented task contributions with #1103, which cover all mandatory steps outlined above. The quality of life improvements will be subject of a follow-up ticket.

PR #1103, opened 2023-11-21
TaskContributions

florian-str:
This PR adds initial support for TaskContributions.
Currently, this includes:
- Option to create, accept, deny TaskContributions
- Email notification for requests, approved and denied contributions
- The required abilities and checks

There are still some tests missing which I need to add before these changes can be merged. However, a review of the implementation is still possible.

- [x] `unique_pending_contribution`
- [x] `can %i[update destroy], TaskContribution do |contrib|*`
- [x] `ParentValidation if parent_entry.blank?`
- [x] `TransferValues`


MrSerth:
Here are some first comments:

- When not signed in but viewing a public task, the "contribute" action button should be disabled as well (similar to "duplicate").
- It would be awesome, if a contribution could be marked as such when viewing it (in addition to the other constraints mentioned in #930) 😊
- When testing your current changes, everything went smoothly -- except that my changes in files, model solutions, and tests are not transferred.
- When accepting a contribution, we should not transfer the `created_at` attribute (since this should reflect when the task was originally created).
- When viewing a task that already has an open contribution started by me earlier, it would great to be redirect to that contribution when clicking "Contribute" (rather than viewing an alert that I am not authorised").
- When editing a previous contribution (still pending), the update fails with an exception:
  ```rb
  undefined method `task' for #<TaskContribution ...>
  ```
  This exception is caused [here](https://github.com/openHPI/codeharbor/blob/03befd73ebf08aad30095c798b053ba3868d3213/app/controllers/contributions_controller.rb#L59).

app/views/task_contribution_mailer/approval_info.html.erb
MrSerth, line 1: I have three rather general remarks on all mailers. While I acknowledge that those remarks weren't newly introduced with your changes (but rather are already present in the repository), we should fix those now along the way:

1. I would prefer to have `.html.slim` templates rather than `.html.erb`. This makes mailers more similar to all other views and works without any further configuration (besides updating the file name and content).
2. A mail is not a complete HTML page. Hence, it just contains the _content_ of the `body` tag, but nothing besides that. 
3. We should move all strings to the correct localizations, so that they can get translated as well.

Feel free to take a look at the mailers located at `app/views/users/mailer`.

app/views/tasks/_form.html.slim
```diff
@@ -1,6 +1,13 @@
...
   .col-md-12.my-4
-    = simple_form_for @task, html: { class: 'project exercise-validation', multipart: true } do |f|
+    - if @task.contribution?
+      - if action_name == 'show' || action_name == 'new'
+        - url = task_contributions_path(@task.parent_id)
+      - else
+        - url = task_contribution_path(@task.parent, @task.task_contribution)
+    - else
+      - url = nil
+    = simple_form_for @task,  url:, html: { class: 'project exercise-validation', multipart: true } do |f|
```
MrSerth, line 10:
```suggestion
    = simple_form_for @task, url:, html: { class: 'project exercise-validation', multipart: true } do |f|
```

db/migrate/20230726100204_add_parent_ids.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddParentIds < ActiveRecord::Migration[7.0]
+  def change
+    add_column :model_solutions, :parent_id, :integer, null: true, limit: 1
+    add_column :tests, :parent_id, :integer, null: true, limit: 1
+    add_column :task_files, :parent_id, :integer, null: true, limit: 1
```
MrSerth, line 7: Shouldn't this be rather a foreign key constrain (with `null: true`) rather than just a single integer column?

spec/factories/task_contributions.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :task_contribution do
+  end
```
MrSerth, line 5: Can we add an association to a task here, so that the factory can be used just by calling?

```
create(:task_contribution)
```

MrSerth (commented):
Here are a few high-level comments. I know that you're still working on the changes and that the tests are still missing. Nevertheless, I thought it would be useful to have some feedback early on.

app/controllers/contributions_controller.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+class ContributionsController < ApplicationController
+  load_and_authorize_resource class: TaskContribution, except: %i[create new]
+  def approve_changes
+    contrib = TaskContribution.find(params[:contribution_id])
+    @task = Task.find(params[:task_id])
+    if @task.apply_contribution(contrib)
```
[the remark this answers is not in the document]
MrSerth: Something like this could work (adopted for the given scenario):

```
raise Pundit::NotAuthorizedError if @uef.present? && @uef.exercise != @exercise
```

app/views/task_contribution_mailer/approval_info.html.erb
MrSerth, line 1: Please transform all mailers to `html.slim` files. Further, a mail only contains everything usually included in `body`, but not the `body` tag itself (or `head` or `html`, `doctype`, ...).

MrSerth, line 1: Further, it would be great if you could remove the `send_contribution_request` from `AccessRequestMailer` together with the respective views, since these are no longer needed.

app/controllers/task_contributions_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+      TaskContributionMailer.approval_info(@task_contribution).deliver_later
+      redirect_to @task, notice: t('.success')
+    else
+      redirect_to [@task, @task_contribution], alert: t('.error')
+    end
+  end
+
+  def discard_changes
+    if @task_contribution.close
+      TaskContributionMailer.rejection_info(@task_contribution).deliver_later
```
MrSerth, line 20: I would suggest not to send this email if the contribution was closed by the contributor.

app/views/tasks/show.html.slim
MrSerth, line 1: Once a contribution got rejected, we have some unexpected behavior with the "Duplicate" button (it's my own contribution):

Bildschirmfoto 2024-04-20 um 21 05 03

app/controllers/task_contributions_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+      self_closed = @task_contribution.user == current_user
+      redirect_to [@task, (self_closed ? @task_contribution : nil)], notice: t('.success')
```
MrSerth, line 22: The flash message for rejecting changes as the original task author could be improved: Currently, the message reads that the decoupling was successful. However, this framing is the one the contributor would expect, not the original task author (for them, it is more about successfully rejecting a change).

app/policies/task_contribution_policy.rb
```diff
@@ -0,0 +1,43 @@
...
+    define_method(action) do
+      record_owner?
+    end
+  end
+
+  private
+
+  def record_owner?
+    @user.present? && @user == suggestion.user
+  end
```
MrSerth, line 42: I would suggest allowing these actions for a platform admin, too.

app/views/tasks/show.html.slim
MrSerth, line 1: When I have an active contribution, the button is correctly labeled as "Edit Contribution". However, the link target is still `.../task_contributions/new` and not directly linking to the `edit` route.

MrSerth, line 1: When visiting a task without being logged in, we currently only show the `duplicate` button. In order to advertise the new contribution feature, we could include a disabled button for contributions as well.

MrSerth, line 1: Without the email, I don't have any option to visit open / pending contributions. Should we include those in our "message center", too? Or, (alternatively / in addition), link to some overview page?

app/controllers/contributions_controller.rb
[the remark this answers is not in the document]
MrSerth: I was still able to manipulate the URL

`http://localhost:7500/tasks/23/task_contributions/2` is invalid, since the contribution is for task `24`. However, this is not checked correctly.

app/views/task_contribution_mailer/approval_info.html.erb
MrSerth: Opening this thread again, since I still have the same feedback (at least regarding my first two remarks).

MrSerth (commented):
Awesome feature, many thanks for that. It's really great to see this complex workflow being enabled and supported by CodeHarbor, pushing our platform closer to the user needs 👏 .

Below is some first feedback, mostly from an end-user perspective (I haven't looked much at the actual code). Please have a look and try to tackle my comments, so that we can proceed merging. Furthermore, please rebase on the latest master to resolve all conflicts.

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
florian-str: In this case, the error message is not correct, however as previously discussed, contributions should not be able to be added to a collection or exported.

app/views/task_contribution_mailer/approval_info.html.erb
florian-str: While I have now converted the `.erb` templates to `.slim`, we should reconsider the general format as there are currently three different formats.

1. Own top-level folder for the mailer, e.g. (`app/views/access_request_mailer`) with `.erb` files
2. Own top-level folder for the mailer, e.g. (`app/views/task_contribution_mailer`) with `.slim` files
3. Sublevel folder for the mailer, e.g. (`app/views/users/mailer`)

If I recall correctly, we decided on choosing the top-level folder based on the similarities to the other platforms. However, if you suggest `app/views/users/mailer` as a source of inspiration, it would no longer match the discussed format. Did your suggestion only concern the template format, or did it also involve the general storage format?

Regarding localization: As there is no saved language preference as opposed to the main openHPI application, we can't deliver the message in the correct language anyway. This raises the question if moving the contents even makes sense at the current stage of the application.

MrSerth: Thanks for pointing out these differences. I'll try to answer your questions to your satisfaction, but don't know whether I succeed.

1. The top-level folder with `.erb` is not what we want, rather use the second approach.
2. I would say that this is the desired format (at least regarding the location, not regarding the code style within the files).
3. This is the desired code style, but a more "complex" example regarding the file location.

Let me explain the file location: A template for mailers is always put in `app/views`, for sure. Then, similar to any controller, a namespace or module name can be added, if present (in the third case that's `users`). Finally, the actual class name of the mailer is used through its underscore syntax (`access_request_mailer`, `task_contribution_mailer`, `mailer`).

The `access_request_mailer`, `task_contribution_mailer` don't have any custom module name or namespace, so that there is one subfolder "missing". The other one, `users/mailer` is therefore expected to be defined as `Users::Mailer`, which would correspond to a location in `app/mailers/users/mailer.rb`. That being said, you'll notice that we don't have such as file, but rater `app/mailers/users/devise_mailer.rb` (and a custom overwrite for the template path). That's not ideal, too. From my point of view, this custom overwrite should be removed and instead the views should be placed at `app/views/users/devise_mailer`. Hence, I just did so in #1365.

Now, we still have the original difference: Some mailers being placed in a subfolder and others in a top-level folder. But this is based on the namespace used for the modules and the fact that the `DeviseMailer` is part of our authentication, where a lot is managed externally. So, for short: Devise-related code is placed in a `users` subfolder (everywhere: controllers, views, mailers) and everything else doesn't use these subfolders.

MrSerth: Regarding localisation: I would still suggest to move it. Thereby, we would ensure to have a single place where all locales are stored. Furthermore, we recently introduced a user locale that is stored in the database (`current_user.preferred_locale`). This could be used for mail recipients, too (or we just include both locales). If you want to get a dedicated locale, just use `I18n.t('key', locale: :de)`.

app/controllers/contributions_controller.rb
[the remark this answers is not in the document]
florian-str: I'm sorry. I thought this scenario was already covered by a different test, but it seems like I was wrong.
The correct fix is now implemented.

app/views/tasks/show.html.slim
florian-str: I have done a redirect in the backend rather than changing the path. This is now changed.

florian-str: This might not provide the best user experience.
We would be unable to differentiate between the owner or a different user. For the owner, the button would be replaced with the edit button once logged in.

While this is correct, it might not be immediately obvious "where" the promised feature went.

florian-str: I have previously suggested a similar feature, but you considered it a quality of life improvement, which should not be implemented in the initial set of functionality.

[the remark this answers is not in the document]
florian-str: The method was now added.

db/migrate/20230726100204_add_parent_ids.rb
florian-str: It is now added.

config/locales/en/models.yml
```diff
@@ -184,6 +189,9 @@ en:
       user_identity:
         one: User identity
         other: User identities
+  task_contribution:
```
MrSerth, line 192: I would have assumed that the plural form is advisable here (and in the German locale, of course). Why is it not?

config/locales/en/models.yml
```diff
@@ -184,6 +189,9 @@ en:
...
+    model:
+      title_string: Contribution for %{task_title}
```
MrSerth, line 194: Usually, I am avoiding to include an object type in a method or variable name (doesn't work always, but often). [Here](https://ruby-for-beginners.rubymonstas.org/bonus/good_names.html) are some further examples. Do you think we can update this name to be more self-explanatory?

app/views/tasks/show.html.slim
```diff
@@ -561,17 +560,39 @@
...
+          = link_to duplicate_task_path(@task), method: :post, class: 'btn btn-important' do
+
             i.fa-regular.fa-clone
             =< t('common.button.duplicate')
-
+        - else
+          div[data-bs-toggle='tooltip' title=t('.guest.disabled.tooltip') data-bs-delay=150]
+            = link_to duplicate_task_path(@task), method: :post, class: 'btn btn-important disabled' do
+              i.fa-regular.fa-clone
+              =< t('common.button.duplicate')
```
MrSerth, line 595: Can we add a disabled button for the task contribution as well? I think, it is free marketing for the feature 🙂

app/views/tasks/show.html.slim
MrSerth, line 1: I am still wondering if we can somehow mark a contribution as such when viewing it (like some change in the icon, title, badge, ....). Maybe also check with @zuhrasofyan for a good idea.

I already noticed that you changed the tab title accordingly, which is nice and a useful first step!

MrSerth: Maybe, this could also show some specific details about the contribution itself (like metadata: Who created it, when, ...)

app/controllers/contributions_controller.rb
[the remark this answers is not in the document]
MrSerth: Keeping this thread open. The following are left:

1. Migrating the remaining mailers to Slim files (could be another PR).
2. Dropping the outer HTML parts of the reaming mailers (could be another PR).
3. Dropping outdated mailers and views (i.e., `send_contribution_request`) which are now superseded by your changes.
4. Moving hard-coded strings to locales
5. Either using the `current_user.preferred_locale` or adding both languages to a mail.
6. In text emails, do not use any HTML tags (i.e., no `<br>`, rather `== "\n\n"`). See the other text mailers.

db/migrate/20230726100204_add_parent_ids.rb
MrSerth: Where did you add these? I would have expected to see a change in `db/migrate/20230726100204_add_parent_ids.rb`, but cannot see it there.

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
MrSerth: Right, then we might want to update the message here (while keeping the correct behavior as previously agreed to).

MrSerth: Ah, I see. For sure, this is a dilemma we might always face for anonymous users. Generally, it would be expected that an anonymous user is not the owner, thus promoting features for other users.

I would say that when the owner of a task sees the message before signing in, and then the button is replaced by editing the task, they still could observe that contributing is available for other tasks.

MrSerth: Okay, I see. Then, let's maybe start a follow-up ticket, listening these improvements? It could ensure that we don't miss them, but can finish this PR timely.

MrSerth (commented):
Thank you again. While I see that you've marked this PR as draft again, I finally managed to catch up with your changes of the past two weeks. Just a few minor comments.

<hr>

Unresolved:

- https://github.com/openHPI/codeharbor/pull/1103#discussion_r1598174749 (previous)
- https://github.com/openHPI/codeharbor/pull/1103#discussion_r1654304809 (later)

app/views/task_contribution_mailer/approval_info.html.erb
MrSerth: Keeping this thread open. The following are left:

1. Migrating the remaining mailers to Slim files (could be another PR).
2. Dropping the outer HTML parts of the reaming mailers (could be another PR).
3. Dropping outdated mailers and views (i.e., `send_contribution_request`) which are now superseded by your changes.
4. Moving hard-coded strings to locales
5. Either using the `current_user.preferred_locale` or adding both languages to a mail.
6. In text emails, do not use any HTML tags (i.e., no `<br>`, rather `== "\n\n"`). See the other text mailers.

db/migrate/20230726100204_add_parent_ids.rb
florian-str: I have added a separate migration converting the column to a foreign key.
It can be found here `db/migrate/20240513071837_parent_id_to_foreign_key.rb`.

MrSerth: Thank you 🙏 

Using a dedicated migration is fine. Still, we could have used the very same migration as well, since the changes are not yet merged upstream.

Regarding the actual migration, I continue to have a small change request: The data type..
- With your first migration, you specify the data type as `integer` manually and with your second migration you specify the foreign key constraint. When looking at the schema, the original table IDs use the `:serial` format, or specifically are `:bitint`. Now, those two data types are mixed and it could happen that a numerically large ID cannot be stored as a `parent`. 
- The issue might even occur faster than with large numbers, since the option `limit: 1` was used in your original migration. `:limit` specifies the "number of bytes" used for this column (the minimum, can be potentially larger I assume). With 1 byte being 8 bits, we have a maximum of 2^8 = 256. I haven't tested, but are we sure that ` parent_id`s >= 256 can be stored?

app/views/task_contributions/_task_contributions.html.slim
```diff
@@ -0,0 +1,22 @@
...
+  .card.my-4
+    .card-header
+      h6
+        span = "#{t('common.created_at')} #{l(contrib.suggestion.created_at, format: :long)}"
+        span.float-end
+          = t("task_contributions.status.#{contrib.status}")
+    .card-body
+      .row.px-2
+        .col-md-12 style='overflow-wrap: break-word;'
+          = t('common.created_by') + ' ' + contrib.suggestion.user.name
```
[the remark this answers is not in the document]
MrSerth: I just see that you modified this behavior in 9b10a060a02c038da509189c6e4378378dca0df7...

app/policies/task_policy.rb
```diff
@@ -47,7 +47,7 @@ def edit?
   end
 
   def destroy?
-    record_owner? || admin?
+    (record_owner? || admin?) && task.task_contribution.blank?
```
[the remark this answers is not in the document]
MrSerth: No, `task.task_contribution.blank?` is only used to identify whether the current "task" is a suggestion / contribution.

Hence, I suggest to extract that to

```rb
def suggestion?
  task.task_contribution.blank?
end
```

app/controllers/task_contributions_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def new
+    raise Pundit::NotAuthorizedError if current_user.nil?
+
+    @task_contribution = TaskContribution.new_for(@task, current_user)
+    authorize @task_contribution
+
+    @task = @task_contribution.suggestion
+    render 'task_contributions/new'
```
MrSerth, line 52: This should be the default `render` action, so that you can probably omit this line.

app/policies/task_contribution_policy.rb
```diff
@@ -0,0 +1,52 @@
...
+    @record.base
+  end
+
+  %i[create? new?].each do |action|
+    define_method(action) do
+      Pundit.policy(@user, base).contribute?
+    end
+  end
+
+  # While discard_changes? and show? look similar, the difference in paraphrases position is relevant
```
MrSerth, line 22: What are you referring to with `paraphrases`?

```suggestion
  # While discard_changes? and show? look similar, the difference in the position of parentheses is relevant
```

app/policies/task_contribution_policy.rb
```diff
@@ -0,0 +1,52 @@
...
+  # discard_changes? can only be executed while the task_contribution is pending by either the owner of the base task or the contrib author
+  # show? the author of the contrib can view it, while it is pending while the owner
```
MrSerth, line 24:
```suggestion
  # show? allows the owner of the base task to view it anytime. Simultaneously, the contributor can view it only while the contrib is still pending
```

app/views/task_contributions/new.html.slim
```diff
@@ -0,0 +1,5 @@
+.header
+  .std-heading
+    i.fa-regular.fa-file-lines
+    =< + t('.new_contribution')
```
MrSerth, line 4: I haven't tested it myself yet, but what is the `+` doing here?

db/migrate/20240610061157_fix_parent_id_format.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class FixParentIdFormat < ActiveRecord::Migration[7.1]
+  def change
+    # rubocop:disable Rails/ReversibleMigration
+    change_column :task_files, :parent_id, :bigint, limit: nil
+    change_column :model_solutions, :parent_id, :bigint, limit: nil
+    change_column :tests, :parent_id, :bigint, limit: nil
+    # rubocop:enable Rails/ReversibleMigration
```
MrSerth, line 9: You could consider marking this migration as [`up_only`](https://apidock.com/rails/ActiveRecord/Migration/up_only) or splitting it in dedicated methods [`up` and `down`](https://edgeguides.rubyonrails.org/active_record_migrations.html#making-the-irreversible-possible).

app/views/task_contributions/new.html.slim
MrSerth, line 4: With #1359, Zuhra added some further nesting and other classes here, which you might want to copy, too:

```suggestion
.header
   .row
     .col-md-12.pb-2.border-bottom
       .std-heading
         i.fa-regular.fa-file-lines
         =< + t('.new_contribution')
```

app/policies/task_contribution_policy.rb
```diff
@@ -19,22 +19,28 @@ def base
...
-  %i[edit? update? destroy?].each do |action|
+  %i[edit? update?].each do |action|
     define_method(action) do
-      record_owner?
+      record_owner? && task_contribution.pending?
     end
   end
 
+  def destroy?
+    false
```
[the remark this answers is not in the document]
MrSerth:
```rb
def no_one
  false
end
```

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
MrSerth: I would propose to hide the `parent_uuid` for contributions, while showing it for clones and other regular tasks.

florian-str: I have implemented such a feature (a list of contributions for a given task), but I'm not really sure where to place the corresponding button, without overcrowding the toolbar at the bottom of the page.

MrSerth: We decided to add another rubric, similar to "Basic Information" or "Additional Information".

MrSerth: Let's go with a third gray box, similar to "Add to collection" or the Download / Export options.

MrSerth:
For the main `/tasks` view, let's set the padding in the buttons:

```css
.nav-container {
  button {
    padding: 10px 40px; // <-- was 10px 60px
    background-color: #f5f5f5;
  }
```

This change needs to be done three times (for tasks, collections, groups).

MrSerth:
I fixed some translations in #1489, which will require a rebase and dealing with the conflicts. For the `app/views/access_request_mailer/send_access_request.html.erb`, only the translation key changed (besides the new file format we fixed in this PR). Please take a brief moment to include the latest changes in your branch 🙏

spec/mailers/task_contribution_mailer_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    it 'sends an email' do
+      expect(contribution_request.to).to include(task_contrib.base.user.email)
+      expect(contribution_request.subject).to include(task_contrib.suggestion.user.name)
+      expect(contribution_request.subject).to include(task_contrib.base.title)
+    end
+
+    it 'contains the correct contents' do
+      expect(contribution_request.body.encoded).to include(task_contrib.suggestion.user.name)
+      expect(contribution_request.body.encoded).to include(task_contrib.base.title)
+      expect(contribution_request.body.encoded).to include('suggested changes')
```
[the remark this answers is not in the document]
florian-str: In this context, I think hard-coding sections of the string would be better to catch issues with the translations.

app/views/tasks/new.html.slim
```diff
@@ -2,4 +2,4 @@
   .std-heading
     i.fa-regular.fa-file-lines
     =< + t('.new_task')
```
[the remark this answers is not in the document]
MrSerth: Unresolving this thread, since the headline doesn't look correct.

<details>
<summary>Current style</summary>

Bildschirmfoto 2024-07-09 um 20 57 05

</details>

<details>
<summary>Desired</summary>

Bildschirmfoto 2024-07-09 um 20 57 12

</details>

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
MrSerth: @florian-str This conversation has been resolved by you. However, I still see the `uuid` and `parent_uuid` for new contributions and edits. What am I missing that resolves this conversation?

<details>
<summary>New</summary>

Bildschirmfoto 2024-07-09 um 20 59 58

</details>

<details>
<summary>Edit</summary>

Bildschirmfoto 2024-07-09 um 21 00 02

</details>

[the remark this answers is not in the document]
MrSerth: On the index page (http://localhost:7500/tasks?visibility=contribution), we have the same issue.

MrSerth: I might have missed a resolution (the conversation on GitHub was resolved, at least), but anonymous users still cannot see the disabled button.

app/views/task_contributions/new.html.slim
MrSerth: Since the headline is not shown correctly (see some comment above), which is caused by these lines, I am unresolving this thread, too.

MrSerth: Actually, I saw that I made a specific (already resolved 🤓) comment about that issue, which is more descriptive. Hence, I am resolving the current discussion and rather unresolving [this one](https://github.com/openHPI/codeharbor/pull/1103/files#diff-e82418b173e8baf2d2e214e67edec3903491ca7c16f9831d04608b893ff96d8f).

MrSerth (commented):
During today's review, I gathered a few more comments. I assume, that many of them are rather small issues with quick fixes, while only a minority of comments are probably a bit more difficult to tackle. While the feature (and this PR) is rather complex, I am sure we're close the finish line regarding the functionality and edge cases implemented! 💪

app/policies/task_contribution_policy.rb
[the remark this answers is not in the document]
MrSerth: Fixing this bug is potentially supported by adding a `has_many` relationship to tasks. A task is either a contribution itself or can have multiple contributions attached to it.

app/views/tasks/new.html.slim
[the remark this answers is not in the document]
florian-str: I wanted to combine all UI related issues into one commit. The issue was resolved but not yet pushed.

I apologize for the confusion.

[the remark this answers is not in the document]
MrSerth: Thanks for clarifying and resolving my confusion. It just seems we are using threaded conversations on GitHub differently, then. 🫣

app/views/tasks/show.html.slim
florian-str: I have added the requested box.

app/views/tasks/show.html.slim
```diff
@@ -160,9 +160,26 @@
...
+            = t("task_contributions.status.#{@task.task_contribution.status}")
+        .row
+          .col-auto.row-label
+            = Task.human_attribute_name('parent')
+            | :
+          .col.row-value
+            = link_to @task.parent.title, task_path(@task.parent)
     legend.toggle-next
-      span.number = '3'
+      span.number = '4'
```
MrSerth, line 182: Does this mean a regular task has the numbers 1,2,4,5,... missing three?

app/views/tasks/show.html.slim
```diff
@@ -160,9 +160,26 @@
             = link_to @task.license.link, @task.license.link
           - else
             = t('.no_license')
-
+    - if @task.contribution?
+      legend.toggle-next.closed
+        span.number = '3'
+        => t('tasks.shared.contribution_info')
+        span.fa-solid.fa-caret-down
+      .show-table.exercise-show style="display:none;"
```
MrSerth, line 168: I found another inline-style that sneaked into the code 🤓

app/views/tasks/show.html.slim
```diff
@@ -535,6 +552,21 @@
...
+      .wrapper
+        - if !@task.contribution? && policy(@task).edit?
+          = link_to t('.button.contribution_list'), task_task_contributions_path(@task), class: 'btn btn-light', 'data-turbolinks' => false
+        - else
+          - if @task.contribution? || @current_user.nil?
+            - contrib_list_unavailable_tooltip = unavailable_tooltip
+          - else
+            - contrib_list_unavailable_tooltip = t('.owner_required_tooltip')
+          div data-bs-toggle='tooltip' title=contrib_list_unavailable_tooltip data-bs-delay=150
+            = link_to t('.button.contribution_list'), task_task_contributions_path(@task), class: 'btn btn-outline-dark disabled', 'data-turbolinks' => false
```
MrSerth, line 568: I would assume you don't need the `'data-turbolinks' => false`, do you? This is only added for some frontend libraries (like select2), but since the target page doesn't (yet) contain such parts, we don't need to disable it.

app/views/tasks/show.html.slim
```diff
@@ -535,6 +552,21 @@
           div data-bs-toggle='tooltip' title=unavailable_tooltip data-bs-delay=150
             = button_tag class: 'btn btn-outline-dark dropdown-toggle disabled', 'data-bs-toggle': 'dropdown' do
               = t('.button.export')
+  .col-md-4.col-sm-6.mb-3.p-0
+    .button-box.border.p-3.h-100
+      .button-description
+        = t('.contribution_list_hint')
+      .wrapper
+        - if !@task.contribution? && policy(@task).edit?
+          = link_to t('.button.contribution_list'), task_task_contributions_path(@task), class: 'btn btn-light', 'data-turbolinks' => false
```
MrSerth, line 561: I would assume you don't need the `'data-turbolinks' => false`, do you? This is only added for some frontend libraries (like select2), but since the target page doesn't (yet) contain such parts, we don't need to disable it.

app/views/tasks/show.html.slim
```diff
@@ -535,6 +552,21 @@
...
+        - else
+          - if @task.contribution? || @current_user.nil?
+            - contrib_list_unavailable_tooltip = unavailable_tooltip
```
MrSerth, line 564: Doesn't this mean that an anonymous user will see a tooltip stating that "the operation is not available for contributions" even when viewing a regular task?

florian-str: No, it would become the `unavailable_tooltip`.

This is a variable defined as
```slim
- if @task.contribution?
        - unavailable_tooltip = t('.task_contribution.disabled.tooltip')
- else
        - unavailable_tooltip = t('.guest.disabled.tooltip')
```

While the nesting is not ideal, it does return the correct message.

MrSerth: Ah, okay. Fine for me, thanks for the clarification!

app/views/tasks/show.html.slim
florian-str: Apparently, it got accidentally removed while resolving a merge conflict.

I have reintroduced the feature.

florian-str: An additional section for Contribution Information has been added.

app/views/task_contribution_mailer/approval_info.html.slim
```diff
@@ -0,0 +1,4 @@
+h3
+  = t('task_contributions.mailer.greeting', user: @contrib_author.name)
+p = t('task_contributions.mailer.approval_info.status_message', task: @task_contrib.base.title)
```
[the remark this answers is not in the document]
florian-str: The task has been enclosed by quotes.

app/views/task_contribution_mailer/rejection_info.html.slim
```diff
@@ -0,0 +1,6 @@
+h3
+  = t('task_contributions.mailer.greeting', user: @contrib_author.name)
+p = t('task_contributions.mailer.rejection_info.status_message', task: @task_contrib.base.title)
```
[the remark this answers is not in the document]
florian-str: The task has been enclosed by quotes.

app/views/tasks/_form.html.slim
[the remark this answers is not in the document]
florian-str: Save task has been replaced by the generic implementation.

spec/mailers/previews/task_contribution_mailer_preview.rb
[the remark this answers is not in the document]
florian-str: The methods have been renamed.

app/controllers/task_contributions_controller.rb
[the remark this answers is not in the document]
florian-str: The implementation of `Task.contribution?` has been changed, to reflect the correct behavior.

app/views/tasks/show.html.slim
florian-str: I have replaced the hard-coded value by an incremented variable. Therefore, this should no longer occur.

app/views/tasks/show.html.slim
florian-str: The data attribute was removed.

app/views/tasks/show.html.slim
florian-str: The data attribute was removed.

app/views/tasks/show.html.slim
florian-str: It is the same as found with the additional information.

The additional information will be fixed by #1527. I would wait until these changes are merged.

app/controllers/tasks_controller.rb
```diff
@@ -208,6 +210,12 @@ def only_authorize_action
     authorize Task
   end
 
+  def redirect_task_contribution
+    return unless @task.present? && @task.contribution?
+
+    redirect_to action: action_name, controller: 'task_contributions', task_id: @task.parent, id: @task.task_contribution
```
MrSerth, line 216: I just got an `ActionController::UrlGenerationError` here. However, I am not sure whether this is some corruption in my local test data or a real issue that could re-occur.

<details>
<summary>Local Reproduction</summary>

I tested with <http://localhost:7500/tasks/856>, which is a contribution belonging to a task that was deleted (probably belonging to <http://localhost:7500/tasks/24>, deleted on `2024-07-09 23:38:44 +0200`.

</details>

Maybe we still ensure to catch those errors?

app/views/tasks/_nested_file_form.html.slim
```diff
@@ -2,21 +2,21 @@
...
+        - display_container = 'display: none'
         .table-header.closed.toggle-next
           => file.object.full_file_name
           span.fa-solid.fa-caret-down
           = file.remove_nested_fields_link :text, class: 'btn btn-main btn-sm float-end' do
             i.fa-solid.fa-xmark.text-danger
             =< t('.remove_file')
-      .file-container.mx-2.my-3 style=display_container
-        = render 'file_form', f:, file:, nested_form_placeholder:
+      .file-container.mx-2.my-3 style="#{display_container}"
```
MrSerth, line 18: This change is currently being discussed as part of #1527, I think. Let's see how we decide there. For now, I would probably not touch it (neither here nor the few occurrences above) if possible.

app/views/tasks/show.html.slim
MrSerth: Alright, fine for me. Before merging this PR, we should probably double-check that no further inline-style is left.

app/controllers/tasks_controller.rb
florian-str: It is an actual issue, which is caused by the fact that deleting a task does not yet modify its associated contributions.

[See this comment for reference](https://github.com/openHPI/codeharbor/pull/1103#discussion_r1671273478)

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
MrSerth: We fixed this issue with 5fecfda92995c0f25a000a5c9ff56c81b5853749.

app/views/tasks/_form.html.slim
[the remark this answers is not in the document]
MrSerth: We fixed this issue with 5fecfda92995c0f25a000a5c9ff56c81b5853749.

app/controllers/task_contributions_controller.rb
MrSerth: This issue has been fixed with 5fecfda92995c0f25a000a5c9ff56c81b5853749.

app/policies/task_contribution_policy.rb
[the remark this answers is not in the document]
MrSerth: We fixed the issue (and are now handling contributions correctly in case a base task is removed). Further, a `has_many` relationship has been added.

app/models/task_file.rb
```diff
@@ -23,9 +37,12 @@ def full_file_name=(full_file_name)
     self.name = File.basename(full_file_name)
   end
 
-  def duplicate
+  def duplicate(set_parent_id: true)
     dup.tap do |file|
       file.attachment.attach(attachment.blob) if attachment.attached?
```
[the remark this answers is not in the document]
MrSerth: We fixed this issue with 5fecfda92995c0f25a000a5c9ff56c81b5853749.

app/controllers/task_contributions_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    if @task_contribution.save(context: :force_validations)
+      TaskContributionMailer.contribution_request(@task_contribution).deliver_later
+      redirect_to [@task, @task_contribution], notice: t('common.notices.object_created', model: TaskContribution.model_name.human)
+    else
+      redirect_to @task, alert: t('common.errors.model_not_saved', model: TaskContribution.model_name.human)
+    end
+  end
+
+  def update
```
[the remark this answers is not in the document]
MrSerth: I created a PR for this issue, since it is unrelated: https://github.com/openHPI/codeharbor/pull/1555

Hence, I am closing this thread.

app/controllers/tasks_controller.rb
MrSerth: Okay, thanks. One could still argue to add a rescue here, but I would leave it in the current state for now. Thus resolving this thread.

MrSerth:
Currently, the linter fails with regard to the associations. While fixing these warnings might be considered a laborious task (especially if everything is working otherwise), it ensures a smooth compatibility with the Rails framework and future updates. Please find my proposal attached to resolve all current warnings.

<details>
<summary>Patch</summary>

```diff
Subject: [PATCH] Refactor associations between Task and TaskContribution
---
Index: app/models/task.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/models/task.rb b/app/models/task.rb
--- a/app/models/task.rb	(revision fd6ccd598e075fd4c7d5534ea11ba6e84c633f7e)
+++ b/app/models/task.rb	(date 1723054227324)
@@ -37,17 +37,21 @@
   has_many :comments, dependent: :destroy
   has_many :ratings, dependent: :destroy
 
-  has_one :task_contribution, dependent: :destroy, inverse_of: :suggestion
-  # We use a custom query to identify task contributions, and especially don't want to find those TaskContributions having the `task_id`.
-  # Reminder: A TaskContribution object with the current task's ID will identify itself as a contribution (which don't have contributions on their own).
-  # TODO: Test relationship (that it contains exactly the required contribs, not too few and not too many).
-  # TODO (?): Test that the `task_id` is not used in the where
-  has_many :task_contributions, ->(task) { unscope(:where).for_task_uuid(task.uuid) }
+  # The `parent` designates the original task that the current task is derived from.
+  belongs_to :parent, class_name: 'Task', foreign_key: :parent_uuid, primary_key: :uuid, optional: true, inverse_of: :children
+  # `Children` are either duplicates of or suggestions for the current task.
+  has_many :children, class_name: 'Task', foreign_key: :parent_uuid, primary_key: :uuid, dependent: :nullify, inverse_of: :parent
+
+  # A `task_contribution` marks a task as a suggestion for a parent task and thus is optional.
+  has_one :task_contribution, dependent: :destroy, inverse_of: :suggestion, required: false
+  # `Contributions` are the change requests with a status (pending, merged, closed) for the current task.
+  has_many :contributions, through: :children, source: :task_contribution, class_name: 'TaskContribution'
+  # `Suggestions` contain the proposals for the current task.
+  has_many :suggestions, through: :contributions, source: :suggestion
 
   belongs_to :user
   belongs_to :programming_language, optional: true
   belongs_to :license, optional: true
-  belongs_to :parent, class_name: 'Task', foreign_key: :parent_uuid, primary_key: :uuid, optional: true, inverse_of: false
 
   accepts_nested_attributes_for :files, allow_destroy: true
   accepts_nested_attributes_for :tests, allow_destroy: true
@@ -211,19 +215,6 @@
     title
   end
 
-  def contributions(user: nil, status: nil, all_states: false)
-    task_filter_set = {parent_uuid: uuid}
-    unless user.nil?
-      task_filter_set[:user] = user
-    end
-    query = TaskContribution.joins(:suggestion)
-      .where(suggestion: task_filter_set)
-    return query if all_states
-    return query.where(status:) unless status.nil?
-
-    query.where(status: :pending)
-  end
-
   def klass
     if contribution?
       TaskContribution
@@ -253,7 +244,7 @@
   def handle_contributions
     # TODO: Write tests for this method
     # TODO: Include in this tests that a rollback occurs if an exception occurs.
-    task_contributions.each do |contribution|
+    contributions.each do |contribution|
       case contribution.status.to_sym
         when :pending
           # We only want to remove the TaskContribution object, but not the suggestion itself.
@@ -275,7 +266,9 @@
 
   def unique_pending_contribution
     if contribution?
-      other_existing_contrib = parent.contributions(user:).where.not(id: task_contribution.id).any?
+      other_existing_contrib = parent.contributions.joins(:suggestion)
+        .where(suggestion: {user:}, status: :pending)
+        .where.not(id: task_contribution.id).any?
       errors.add(:task_contribution, :duplicated) if other_existing_contrib
     end
   end
Index: app/models/task_contribution.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/models/task_contribution.rb b/app/models/task_contribution.rb
--- a/app/models/task_contribution.rb	(revision fd6ccd598e075fd4c7d5534ea11ba6e84c633f7e)
+++ b/app/models/task_contribution.rb	(date 1723054415591)
@@ -1,13 +1,14 @@
 # frozen_string_literal: true
 
 class TaskContribution < ApplicationRecord
+  # The `suggestion` contains the updated version of the task.
   belongs_to :suggestion, class_name: 'Task', inverse_of: :task_contribution, foreign_key: :task_id
+  # The `base` denotes the original version of the task that should be updated.
+  has_one :base, through: :suggestion, source: :parent, required: true
+
   delegate :user, :user=, :access_level=, to: :suggestion
   accepts_nested_attributes_for :suggestion
 
-  # TODO: Test this scope (maybe already done with the test written for task.rb#l45).
-  scope :for_task_uuid, ->(uuid) { joins(:suggestion).where(suggestion: {parent_uuid: uuid}) }
-
   # TODO: Use `_prefix: true`
   enum status: {pending: 0, closed: 1, merged: 2}, _default: :pending
 
@@ -29,14 +30,6 @@
     end
   end
 
-  def base
-    suggestion.parent
-  end
-
-  def base=(task)
-    suggestion.parent = task
-  end
-
   def to_s
     I18n.t('task_contributions.model.contribution_title', task_title: base.title)
   end
Index: app/policies/task_policy.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/policies/task_policy.rb b/app/policies/task_policy.rb
--- a/app/policies/task_policy.rb	(revision fd6ccd598e075fd4c7d5534ea11ba6e84c633f7e)
+++ b/app/policies/task_policy.rb	(date 1723054033804)
@@ -59,9 +59,12 @@
   end
 
   def contribute?(check_other_contrib: true)
-    return false if @user.blank?
+    return false if @user.blank? # Disallow contributing without an active session (anonymous user).
+    return false unless show? && !edit? # Contributing is only allowed if a user cannot make direct changes.
+    return true unless check_other_contrib # Allow action, if existing contributions are irrelevant.
 
-    (show? && !edit?) && (!check_other_contrib || task.contributions(user: @user).none?)
+    # Contributing is only allowed if no other pending contributions exists.
+    task.contributions.joins(:suggestion).where(suggestion: {user: @user}, status: :pending).none?
   end
 
   private
Index: app/views/tasks/show.html.slim
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/views/tasks/show.html.slim b/app/views/tasks/show.html.slim
--- a/app/views/tasks/show.html.slim	(revision fd6ccd598e075fd4c7d5534ea11ba6e84c633f7e)
+++ b/app/views/tasks/show.html.slim	(date 1723053783004)
@@ -623,8 +623,9 @@
             =< t('task_contributions.buttons.decouple')
       - unless @task.contribution?
         - if policy(@task).contribute?(check_other_contrib: false)
-          - if @task.contributions(user: @current_user).any?
-            = link_to edit_task_task_contribution_path(@task, @task.contributions(user: @current_user).first), method: :get, class: 'btn btn-important' do
+          - contributions = @task.contributions.joins(:suggestion).where(suggestion: {user: @current_user}, status: :pending)
+          - if contributions.any?
+            = link_to edit_task_task_contribution_path(@task, contributions.first), method: :get, class: 'btn btn-important' do
               i.fa-regular.fa-edit
               =< t('task_contributions.buttons.edit_contribution')
           - else
Index: app/controllers/task_contributions_controller.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/controllers/task_contributions_controller.rb b/app/controllers/task_contributions_controller.rb
--- a/app/controllers/task_contributions_controller.rb	(revision fd6ccd598e075fd4c7d5534ea11ba6e84c633f7e)
+++ b/app/controllers/task_contributions_controller.rb	(date 1723053430554)
@@ -36,7 +36,7 @@
 
   def index
     authorize @task, :edit?
-    @task_contributions = @task.contributions(all_states: true)
+    @task_contributions = @task.contributions
   end
 
   def show
```

</details>

app/views/tasks/show.html.slim
MrSerth: With my latest push, I've fixed this occurrence and also validated that no other inline styles are currently used.

spec/controllers/task_contributions_controller_spec.rb
```diff
@@ -0,0 +1,288 @@
...
+
+      it 'renders the form again' do
+        post_request
+        expect(response).to render_template('tasks/new')
+      end
+    end
+  end
+
+  describe 'POST #approve_changes' do
+    let(:user) { original_author }
```
MrSerth, line 97: Changing the `user` does not only have an effect on the session (allowing the `POST` to `approve_changes` to pass), but also changes the user who owns the contribution.

Specifically, in my test:
- the `task` is owned by `user`
- the `contribution` is owned by the `user`, too.

The current setup is fine to pass the test, but it is semantically difficult.

florian-str: I have added an additional variable to avoid this effect.

app/models/task_file.rb
```diff
@@ -46,6 +70,13 @@ def extract_text_data
 
   private
 
+  def attach_parent_blob
+    # The comparison of the blob ID is used as a safeguard to ensure updated files in a task contribution are not overwritten unintended.
+    return unless parent_blob_id.to_i == parent.attachment.blob.id
+
+    attachment.attach(parent.attachment.blob)
+  end
+
```
MrSerth, line 79: As discussed: A valid test for this method would be a feature spec (soon called a system spec) with the following steps:

1. Access an existing task with an attached file that is not owned by the current user
2. Click on "Contribute"
3. Perform any change, e.g., for the title
4. Invalidate the task by removing the `language` (or do any other change)
5. Click "Save"
6. Expect that the new contribution was not saved and that errors where found. Further ensure that the attached file name is shown as part of a `div` with the classes `.btn.btn-light.border-right-0.disabled` (`app/views/tasks/_editor.html.slim:23`) and that re-uploading a new file is possible (`app/views/tasks/_editor.html.slim:24-25`).
7. Fix the previous validation error, e.g. by fixing the language
8. Hit "Save" again
9. Expect that the contribution was successfully saved. Expect that it features the new name and still has an attachment.

A second test explicitly checking that the method under test is not used for changed files would be:

1. Access an existing task with an attached file that is not owned by the current user
2. Click on "Contribute"
3. Perform any change, e.g., for the title
4. Replace the attachment of the existing file by selecting a new file for upload. [This post](https://til.hashrocket.com/posts/c790268652-attach-a-file-with-capybara) shows how to do that with Capybara.
5. Invalidate the task by removing the `language` (or do any other change)
6. Click "Save"
7. Expect that the new contribution was not saved and that errors where found. Further ensure that no file name is shown for the attachment and re-selecting the file is required (`app/views/tasks/_editor.html.slim:29-30`). 
8. Fix the previous validation error, e.g. by fixing the language and re-select the file to upload
9. Hit "Save" again
10. Expect that the contribution was successfully saved. Expect that it features the new name and still has the uploaded attachment.

app/controllers/task_contributions_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+      TaskContributionMailer.send_rejection_info(@task_contribution, duplicate).deliver_later
+      redirect_to @task, notice: t('.success')
+    else
+      redirect_to [@task, @task_contribution], alert: t('.error')
+    end
+  end
+
+  def index
+    @task_contributions = @task.contributions.order(created_at: :desc).paginate(page: params[:page], per_page: per_page_param)
+    raise Pundit::NotAuthorizedError unless policy(TaskContribution).index? base: @task
```
MrSerth, line 39:
```suggestion
    raise Pundit::NotAuthorizedError unless policy(@task_contributions).index? base: @task
```

This change will fail the `spec/controllers/task_contributions_controller_spec.rb` for `index`, which will be fixed by adding `let(:user) { original_author }` there.

spec/models/task_spec.rb
```diff
@@ -284,4 +377,219 @@
...
+        apply_contribution
+        expect(contribution.status_merged?).to be(true)
+      end
+
+      it 'transfers the metadata' do
+        apply_contribution
+        expect(task.meta_data).to eq('New meta data')
+        expect(task.submission_restrictions).to eq('New submission restrictions')
+        expect(task.external_resources).to eq('New external resources')
+        expect(task.grading_hints).to eq('New grading hints')
```
MrSerth, line 568: Here is a patch for setting real-world meta data:

<p>

```diff
Subject: [PATCH] Set ProFormA meta data in contributions
---
Index: spec/models/task_spec.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/spec/models/task_spec.rb b/spec/models/task_spec.rb
--- a/spec/models/task_spec.rb	(revision 0d282a9a941e27556ce306e5d2215cf7bd7cce3f)
+++ b/spec/models/task_spec.rb	(date 1726044898071)
@@ -542,12 +542,17 @@
     let(:task) { create(:task) }
     let(:contribution) { create(:task_contribution, base: task) }
 
+    let(:meta_data) { attributes_for(:task, :with_meta_data)[:meta_data] }
+    let(:submission_restrictions) { attributes_for(:task, :with_submission_restrictions)[:submission_restrictions] }
+    let(:external_resources) { attributes_for(:task, :with_external_resources)[:external_resources] }
+    let(:grading_hints) { attributes_for(:task, :with_grading_hints)[:grading_hints] }
+
     before do
       contribution.suggestion.title = 'New title'
-      contribution.suggestion.meta_data = 'New meta data'
-      contribution.suggestion.submission_restrictions = 'New submission restrictions'
-      contribution.suggestion.external_resources = 'New external resources'
-      contribution.suggestion.grading_hints = 'New grading hints'
+      contribution.suggestion.meta_data = meta_data
+      contribution.suggestion.submission_restrictions = submission_restrictions
+      contribution.suggestion.external_resources = external_resources
+      contribution.suggestion.grading_hints = grading_hints
     end
 
     context 'when the contribution is pending' do
@@ -562,10 +567,10 @@
 
       it 'transfers the metadata' do
         apply_contribution
-        expect(task.meta_data).to eq('New meta data')
-        expect(task.submission_restrictions).to eq('New submission restrictions')
-        expect(task.external_resources).to eq('New external resources')
-        expect(task.grading_hints).to eq('New grading hints')
+        expect(task.meta_data).to eq meta_data
+        expect(task.submission_restrictions).to eq submission_restrictions
+        expect(task.external_resources).to eq external_resources
+        expect(task.grading_hints).to eq grading_hints
       end
 
       it 'transfers the labels' do
Index: spec/controllers/task_contributions_controller_spec.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/spec/controllers/task_contributions_controller_spec.rb b/spec/controllers/task_contributions_controller_spec.rb
--- a/spec/controllers/task_contributions_controller_spec.rb	(revision 0d282a9a941e27556ce306e5d2215cf7bd7cce3f)
+++ b/spec/controllers/task_contributions_controller_spec.rb	(date 1726044748295)
@@ -8,7 +8,7 @@
   let(:contribution_user) { create(:user) }
   let(:user) { contribution_user }
   let(:original_author) { create(:user) }
-  let!(:task) { create(:task, user: original_author, access_level: 'public') }
+  let!(:task) { create(:task, :with_meta_data, :with_submission_restrictions, :with_external_resources, :with_grading_hints, user: original_author, access_level: 'public') }
   let(:contrib_task) { build(:task, user: contribution_user, title: 'Modified title', parent_uuid: task.uuid) }
   let(:contribution) { build(:task_contribution, suggestion: contrib_task, base: task, status: :pending) }
 ```

</p>
</details>

MrSerth (commented):
Awesome work! Just a few minor comments left 🚀

spec/models/task_spec.rb
florian-str: I have applied your patch.

app/controllers/task_contributions_controller.rb
florian-str: I have implemented the changes.

app/models/task_file.rb
MrSerth: Awesome, thanks for the first feature spec! 

There is currently a failure on GitHub Actions, which is not related to your code. We are missing the required `libvips` library, which will be included with #1589: [proposed change for `ci.yml` including `libvips`](https://github.com/openHPI/codeharbor/pull/1589/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f).

MrSerth: I extracted the required change with #1684, which we merged on master already. Rebasing will include this change, so that the test should pass.

MrSerth: We added both tests and rebased the current branch onto the latest master, resolving the previous test failures. Hence, this conversation can be resolved.

MrSerth (approved):
What a huge improvement to CodeHarbor! 👏 I super happy about these changes and can't wait to see what our users will do with contributions. From my conversations with teachers, MOOC instructors and university professors, I am confident that this work can ease their workflows and support in better teaching 💪

Thank you so much for your countless and tireless work on this topic! 💐

PR #1104, opened 2023-11-21
Fix small interoperability issues

kkoehn:
Proforma Export:
- fix forcing default usage-of-lms
- simplify simple description strings


MrSerth (approved):
Perfect, thank you! It works like a charm and fixes both issues 👍

PR #1117, opened 2023-11-29
Redirect users after sign in to their desired page

MrSerth:
When a user is not signed in but accesses a restricted page, we currently only show a generic error message and redirect them to the root page. However, it would be a nice quality of life improvement to support them in signing in and then access the desired page immediately. This PR adds that functionality.


florian-str (approved):
It looks good and seems to work reliably.

Issue #781, opened 2022-11-15
Small issue compilation extract: collections and groups


kkoehn:
-collections
  - [ ] Collections can't be deleted manually (the might be deleted, once they have no user attached)
    - it's a feature, rename leave button to delete
  - [ ] Do we still want to use Collections, Groups and Carts / what purpose/differences do they have?
    - remove cart
    - add to collection -> add option on create collection on the fly (to replace cart)
  - [ ] Accepting a non existing collection errors (You get a message with a link)
    - better error message
  - [ ] Own exercises in collections are "locked"
    - check again -> fix
  - [ ] There is no way to see who has access to a collection
    - fix with groups <> collections

- groups
  - [ ] Adding an exercise to a group has a strange flow (you have to go to edit and enter it in a select2 there)
    - leave option in new/edit form
    - add option to add to group in show

PR #1130, opened 2023-12-06
781 small issue compilation extract collections and groups

kkoehn:
Tackles most collection issues from #781 

- [x] add fast collection creation in task/show
- [x] prevent a user joining a collection multiple times (validation, error handling)
- [x] show count of users in collection/show (to preserve anonymity, since we have no collection admin)
- [x] rename leave button to delete, if user is last member of collection
- [x] `view_shared` didn't work at all
- [x] fix  tests
- [x] add tests

Additional tasks after review/discusstion:
- [x] add collection visibility
  - [x] default private => only members can see it
  - [x] optional public

Will be handled in different PR(s):
- [ ] show # of open invitations (since they can be used to get access at any time)
- [ ] consume invitations
- [ ] public collections can be saved by any user (but not edited) (follow)
- [ ] collection can be assigned to group (edit rights)

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -157,6 +157,7 @@ class CollectionsController < ApplicationController
   end
 
   def collection_params
-    params.require(:collection).permit(:title, :task_ids, :visibility_level, :description, collection_tasks_attributes: collection_tasks_params)
+    params.require(:collection).permit(:title, :task_ids, :visibility_level, :description,
+      collection_tasks_attributes: collection_tasks_params)
   end
 end
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -6,8 +6,7 @@ class CollectionsController < ApplicationController
   before_action :load_and_authorize_collection, except: %i[index new create create_ajax]
 
   def index
-    @collections = Collection.includes(:users, tasks: %i[user groups])
-      .where(collection_users: {user: current_user})
+    @collections = Collection.member(current_user).or(Collection.public_access).includes(:users, tasks: %i[user groups])
       .order(id: :asc)
       .paginate(page: params[:page], per_page: per_page_param)
       .load
@@ -162,6 +161,6 @@ class CollectionsController < ApplicationController
   end
 
   def collection_params
-    params.require(:collection).permit(:title, :task_ids, :description, collection_tasks_attributes: collection_tasks_params)
+    params.require(:collection).permit(:title, :task_ids, :visibility_level, :description, collection_tasks_attributes: collection_tasks_params)
   end
 end
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -110,7 +110,7 @@ class CollectionsController < ApplicationController
   end
 
   def save_shared
-    return redirect_to user_messages_path(user), alert: t('.errors.already_member') if @collection.users.include? user
+    return redirect_to user_messages_path(current_user), alert: t('.errors.already_member') if @collection.users.include? current_user
 
     @collection.users << current_user
     if @collection.save
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -65,12 +65,8 @@ class CollectionsController < ApplicationController
   end
 
   def remove_task_ajax
-    if @collection.remove_task(params[:task])
-      redirect_to Task.find_by(id: params[:task]), notice: t('.success_notice')
-    else
-      flash.now[:alert] = t('.error')
-      head :ok
-    end
+    @collection.remove_task(params[:task])
+    redirect_to Task.find_by(id: params[:task]), notice: t('.success_notice')
   end
 
   def remove_all
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -3,7 +3,7 @@
 require 'zip'
 
 class CollectionsController < ApplicationController
-  before_action :load_and_authorize_collection, except: %i[index new create]
+  before_action :load_and_authorize_collection, except: %i[index new create create_ajax]
 
   def index
     @collections = Collection.includes(:users, tasks: %i[user groups])
@@ -36,6 +36,19 @@ class CollectionsController < ApplicationController
     end
   end
 
+  def create_ajax
+    @collection = Collection.new(collection_params)
+    @collection.users << current_user
+    authorize @collection
+
+    if @collection.save
+      redirect_to @collection.tasks.first, notice: t('.success_notice')
+    else
+      flash.now[:alert] = t('.error')
+      head :ok
+    end
+  end
+
   def update
     if @collection.update(collection_params)
       redirect_to collections_path, notice: t('common.notices.object_updated', model: Collection.model_name.human)
@@ -49,6 +62,15 @@ class CollectionsController < ApplicationController
       redirect_to @collection, notice: t('common.notices.object_removed', model: Task.model_name.human)
     else
       redirect_to @collection, alert: t('.cannot_remove_alert')
+    end
+  end
+
+  def remove_task_ajax
+    if @collection.remove_task(params[:task])
+      redirect_to Task.find_by(id: params[:task]), notice: t('.success_notice')
+    else
+      flash.now[:alert] = t('.error')
+      head :ok
     end
   end
 
@@ -93,6 +115,8 @@ class CollectionsController < ApplicationController
   end
 
   def save_shared
+    return redirect_to user_messages_path(user), alert: t('.errors.already_member') if @collection.users.include? user
+
     @collection.users << current_user
     if @collection.save
       redirect_to @collection, notice: t('.success_notice')
@@ -138,6 +162,6 @@ class CollectionsController < ApplicationController
   end
 
   def collection_params
-    params.require(:collection).permit(:title, :description, collection_tasks_attributes: collection_tasks_params)
+    params.require(:collection).permit(:title, :task_ids, :description, collection_tasks_attributes: collection_tasks_params)
   end
 end
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -3,7 +3,7 @@
 require 'zip'
 
 class CollectionsController < ApplicationController
-  before_action :load_and_authorize_collection, except: %i[index new create create_ajax]
+  before_action :load_and_authorize_collection, except: %i[index new create]
 
   def index
     @collections = Collection.member(current_user).or(Collection.public_access).includes(:users, tasks: %i[user groups])
@@ -24,27 +24,16 @@ class CollectionsController < ApplicationController
   def edit; end
 
   def create
-    @collection = Collection.new(collection_params)
-    @collection.users << current_user
+    @collection = Collection.new(collection_params.merge(users: [current_user]))
     authorize @collection
 
-    if @collection.save
+    if @collection.tasks.any?
+      # Redirect back to task#show as collections with tasks can only be created from that view
+      create_collection_with_task
+    elsif @collection.save
       redirect_to collections_path, notice: t('common.notices.object_created', model: Collection.model_name.human)
     else
       render :new
-    end
-  end
-
-  def create_ajax
-    @collection = Collection.new(collection_params)
-    @collection.users << current_user
-    authorize @collection
-
-    if @collection.save
-      redirect_to @collection.tasks.first, notice: t('.success_notice')
-    else
-      flash.now[:alert] = t('.error')
-      head :ok
     end
   end
 
@@ -57,16 +46,12 @@ class CollectionsController < ApplicationController
   end
 
   def remove_task
+    redirect_target = params[:return_to_task] ? Task.find(params[:task]) : @collection
     if @collection.remove_task(params[:task])
-      redirect_to @collection, notice: t('common.notices.object_removed', model: Task.model_name.human)
+      redirect_to redirect_target, notice: t('common.notices.object_removed', model: Task.model_name.human)
     else
-      redirect_to @collection, alert: t('.cannot_remove_alert')
+      redirect_to redirect_target, alert: t('.cannot_remove_alert')
     end
-  end
-
-  def remove_task_ajax
-    @collection.remove_task(params[:task])
-    redirect_to Task.find_by(id: params[:task]), notice: t('.success_notice')
   end
 
   def remove_all
@@ -131,6 +116,14 @@ class CollectionsController < ApplicationController
   end
 
   private
+
+  def create_collection_with_task
+    if @collection.save
+      redirect_to @collection.tasks.first, notice: t('collections.create.success_notice')
+    else
+      flash.now[:alert] = t('collections.create.error')
+    end
+  end
 
   def share_message
     user = User.find_by(email: params[:user])
--- a/app/views/collections/_form.html.slim
+++ b/app/views/collections/_form.html.slim
@@ -9,7 +9,7 @@
         = f.label :description, Collection.human_attribute_name('description'), class: 'form-label'
         = f.text_area :description, class: 'form-control', rows: 10
       .field-element
-        = f.label :visibility_level,  Collection.human_attribute_name("visibility_level"), class: 'form-label collection-edit-visibility'
+        = f.label :visibility_level, Collection.human_attribute_name("visibility_level"), class: 'form-label collection-edit-visibility'
         .radio-switch
           = f.radio_button :visibility_level, :public, value: :public, checked: true
           = label_tag 'collection_visibility_level_public', t('collections.visibility.public'), class: 'radio-left'
--- a/app/views/collections/_form.html.slim
+++ b/app/views/collections/_form.html.slim
@@ -8,6 +8,14 @@
       .field-element
         = f.label :description, Collection.human_attribute_name('description'), class: 'form-label'
         = f.text_area :description, class: 'form-control', rows: 10
+      .field-element
+      = f.label :visibility_level,  Collection.human_attribute_name("visibility_level"), class: 'form-label collection-edit-visibility'
+      .radio-switch
+        = f.radio_button :visibility_level, :public, value: :public, checked: true
+        = label_tag 'collection_visibility_level_public', t('collections.visibility.public'), class: 'radio-left'
+        = f.radio_button :visibility_level, :private, value: :private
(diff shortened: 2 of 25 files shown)


MrSerth:
> @MrSerth could you take a quick look at the changes (esp. the ui) to see if you have any suggestions?

Sure -- happy to do so. I think, the code mostly looks fine. Probably, one of the first things I would mention is to move all locales from the HTML, controller, ... to the respective locale files. Since we are planning to translate CodeHarbor to German in the future, we should already pay attention there.

Regarding the UI, I like the new option to create a collection. For the input / button, I would probably take a look at a [bootstrap input group with a button](https://getbootstrap.com/docs/5.3/forms/input-group/#button-addons) to make that look nice (and add some more padding to the last element of the list of existing collections). 

Other than that, I was wondering whether we should (in addition to the existing method) allow the assignment of a task to a collection through the tasks#edit route, similar to groups. There, a select2 (similar to labels) could be used, either showing existing groups or allowing to create a new one. This method (select2) could also be suitable for the tasks#show route, but I am not yet sure whether I prefer it.

When viewing collections, I was wondering whether there is some "user management": I am now seeing that I've shared my collection with X other users, but can I see whom I gave access or remove them?

kkoehn:
Reflection (more or less) of meeting on dec 20th

> Sure -- happy to do so. I think, the code mostly looks fine. Probably, one of the first things I would mention is to move all locales from the HTML, controller, ... to the respective locale files. Since we are planning to translate CodeHarbor to German in the future, we should already pay attention there.

done
> Regarding the UI, I like the new option to create a collection. For the input / button, I would probably take a look at a [bootstrap input group with a button](https://getbootstrap.com/docs/5.3/forms/input-group/#button-addons) to make that look nice (and add some more padding to the last element of the list of existing collections).

good idea, it looks a lot cleaner now
image

> Other than that, I was wondering whether we should (in addition to the existing method) allow the assignment of a task to a collection through the tasks#edit route, similar to groups. There, a select2 (similar to labels) could be used, either showing existing groups or allowing to create a new one. This method (select2) could also be suitable for the tasks#show route, but I am not yet sure whether I prefer it.

I would prefer to keep the Collection addition in the show-view, since it should be easy and fast to add a Task to a Collection. I am not a fan of the group-handling either - it just feels confusing.

> When viewing collections, I was wondering whether there is some "user management": I am now seeing that I've shared my collection with X other users, but can I see whom I gave access or remove them?

If we add user management to Collection they basically become groups, therefore I would keep Collections a bit more simple and maybe a little anarchic. It would be nice if a member of the Collection could change its visibility between `private` and `public`, to enable other users to "follow" or "subscribe" a Collection without becoming an actual member.

MrSerth:
Thanks for tackling continuously working on this PR and addressing some of the points we discussed in late December. I think, the code and the functionality is fine. There are some optical things, but given upcoming rebrush, I probably wouldn't pay too much attention. The only exception might be the new Collections#show Route, where the three icons (within the orange box) all have other sizes, weights, and maybe the two lines of text are too close to each other (or the second line should be lighter...) I haven't tried anything, but it might be a bit too crowded: 

Bildschirmfoto 2024-01-12 um 21 48 29

Other than checking the functionality (which seems to be fine), I haven't made a (code) review -- leaving that open for @Dome-GER.

kkoehn:
> Thanks for tackling continuously working on this PR and addressing some of the points we discussed in late December. I think, the code and the functionality is fine. There are some optical things, but given upcoming rebrush, I probably wouldn't pay too much attention. The only exception might be the new Collections#show Route, where the three icons (within the orange box) all have other sizes, weights, and maybe the two lines of text are too close to each other (or the second line should be lighter...) I haven't tried anything, but it might be a bit too crowded:
> Bildschirmfoto 2024-01-12 um 21 48 29

I used longer titles during testing, there it didn't feel that cramped. Since I'm planning to add even more information there I tried to clean it up a bit:
image
What do you think?

MrSerth:
> What do you think?

Ah, much better! I like that, thanks! 🎨🖌️

Dome-GER:
Could you please resolve the conflicts? 😇

app/controllers/collections_controller.rb
```diff
@@ -3,11 +3,10 @@
 require 'zip'
 
 class CollectionsController < ApplicationController
-  before_action :load_and_authorize_collection, except: %i[index new create]
+  before_action :load_and_authorize_collection, except: %i[index new create create_ajax]
```
Dome-GER, line 6: Having a separate action for responding to AJAX is a bit uncommon, I'd say. Usually, one would allow an action to respond to both. An alternative would be to route it to a dedicated controller (e.g., scoped controller in an `Ajax` module). The choice depends on the amount of code / logic to be re-used, of course.

How is this handled in other places in your code, so is this consistent?

app/controllers/collections_controller.rb
```diff
@@ -3,11 +3,10 @@
...
 
   def index
-    @collections = Collection.includes(:users, tasks: %i[user groups])
-      .where(collection_users: {user: current_user})
+    @collections = Collection.member(current_user).or(Collection.public_access).includes(:users, tasks: %i[user groups])
```
Dome-GER, line 9: It's hard to tell whether this is related to the AJAX action or not. Usually, I try to keep the changes separate, but probably not needed here.

app/controllers/collections_controller.rb
```diff
@@ -36,6 +35,19 @@ def create
     end
   end
 
+  def create_ajax
+    @collection = Collection.new(collection_params)
+    @collection.users << current_user
+    authorize @collection
+
+    if @collection.save
+      redirect_to @collection.tasks.first, notice: t('.success_notice')
```
Dome-GER, line 44: Is a redirect the expected behavior for an AJAX call? 🤔 I'd expect a JSON response, no?

app/views/collections/_form.html.slim
```diff
@@ -8,6 +8,14 @@
       .field-element
         = f.label :description, Collection.human_attribute_name('description'), class: 'form-label'
         = f.text_area :description, class: 'form-control', rows: 10
+      .field-element
+        = f.label :visibility_level,  Collection.human_attribute_name("visibility_level"), class: 'form-label collection-edit-visibility'
```
Dome-GER, line 12: How about a real label here?

app/views/collections/show.html.slim
```diff
@@ -1,10 +1,26 @@
...
+          =< t('.visibility.public')
+        - else
+          i.fa-solid.fa-lock.lock-icon
+          =< t('.visibility.private')
+      span
+        i.fa-solid.fa-users
+        - if @collection.users.length == 1
+          =< t('.no_other_user')
+        - else
+          =< t('.num_of_other_users', count: @collection.users.length - 1)
```
Dome-GER, line 21: Is this a deliberate choice for `length` since we need the collection users anyway?
I was just wondering, since further down you call `count`.

app/views/collections/show.html.slim
```diff
@@ -49,22 +65,22 @@
...
-            .input-group
-              = text_field_tag 'user', nil, form: 'share', class: 'form-control'
-              = submit_tag t('.button.share'), form: 'share', class: 'btn btn-light', type: 'button'
         - if policy(@collection).edit?
+          .btn-group
+            = button_tag type: 'button', 'data-bs-toggle': 'dropdown', aria: {haspopup: "true", expanded: "false"}, class: 'btn btn-important nav-btn-exercise dropdown-toggle split' do
+              => t('.button.share')
+              span.caret
+            .dropdown-menu#share-menu
+              .dropdown-header style="padding-left: 2px;"
```
Dome-GER, line 78: What is your "policy" regarding inline styles? 🤔

app/views/tasks/show.html.slim
```diff
@@ -478,13 +478,21 @@
             = button_tag class: 'btn btn-light dropdown-toggle', 'data-bs-toggle': 'dropdown' do
               = t('.button.add_to_collection')
             ul.scrollable.dropdown-menu[role="menu"]
-              - if !current_user.collections.empty?
-                - current_user.collections.reverse.each do |collection|
-                  li
-                    = link_to collection.title, add_to_collection_task_path(collection: collection), method: 'post', class: 'dropdown-item'
-              - else
+              - current_user.collections.reverse.each do |collection|
```
Dome-GER, line 481: I assume this is using `reverse` since collections have a default order by creation date, and we want to list the most recent first. Could be made explicit (but I saw that this code has been moved, and it was this way before).

app/views/tasks/show.html.slim
```diff
@@ -478,13 +478,21 @@
...
+                  - else
+                    = link_to "<span class=\"collection-checkmark\"></span>#{collection.title}".html_safe, add_to_collection_task_path(collection: collection), method: :post, class: 'dropdown-item'
+
+              li
+                hr.dropdown-divider
+              li
+                = form_with model: Collection.new, url: create_ajax_collections_path do |f|
+                  .create-collection.input-group
+                    = f.text_field :title, class: 'form-control', placeholder: t('.create_collection_placeholder')
+                    = f.hidden_field :task_ids, value: @task.id
```
Dome-GER, line 494: `ids` vs. `id` is confusing for me. But I don't know the domain.

config/routes.rb
```diff
@@ -32,12 +32,17 @@
...
+      post :create_ajax
+    end
+
     member do
       get :download_all
       get :view_shared # ???
 
       patch :remove_all
       patch :remove_task
+      patch :remove_task_ajax
```
Dome-GER, line 45: You can restrict routes as follows, assuming they shall only be available for AJAX calls:

```suggestion
      patch :remove_task_ajax, constraints: ->(r) { r.xhr? }
```

db/schema.rb
```diff
@@ -82,6 +82,7 @@
     t.datetime "updated_at", null: false
     t.string "title"
     t.string "description", default: "", null: false
+    t.integer "visibility_level", limit: 2, default: 0, null: false, comment: "Used as enum in Rails"
```
Dome-GER, line 85: In the migration, `limit` is 1, but here it is 2. 🤔 

And why is the schema timestamp not updated?

spec/factories/collection_user.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :collection_user do
```
Dome-GER, line 4: Is this a useful factory? I'd rather expect to see the association between a user and collection be created in the test setup explicitly.

app/views/collections/show.html.slim
kkoehn: No, I have not thought about the use of `length` here - The `count` further down was already used before. If I see it correctly `size` should fit them both better.

app/views/collections/show.html.slim
kkoehn: We don't really have a policy, but I try to avoid it normally

app/views/tasks/show.html.slim
kkoehn: Good idea!

app/views/tasks/show.html.slim
kkoehn: I understand the confusion, but I think it makes sense:
A `Collection` `has_many` `Tasks` and here we want to create a collection with one `Task` already added

db/schema.rb
kkoehn: The changed limit is interesting, setting limit for integers specifies the number of bytes that the number uses. Since Postgres does not have a `tinyint` type it reverts to 2, so it can fit a `smallint`.

the timestamp probably went MIA during the merge

spec/factories/collection_user.rb
kkoehn: I use it in `collection_user_spec.rb` to validate the presence of validations. I think it makes sense and is useful, even if only for that one case.

app/controllers/collections_controller.rb
Dome-GER: So this will be extracted or are you planning to keep it here, now?

app/controllers/collections_controller.rb
```diff
@@ -25,11 +24,13 @@ def new
...
 
   def create
-    @collection = Collection.new(collection_params)
-    @collection.users << current_user
+    @collection = Collection.new(collection_params.merge(users: [current_user]))
     authorize @collection
+    create_with_task = @collection.tasks.any?
 
-    if @collection.save
+    if create_with_task
```
Dome-GER, line 31: I'd rather add an explanatory code comment here instead of making it a descriptive variable.
Benefit: A code comment can focus on the _WHY_. Why is the redirect behavior different if there are any tasks?

app/views/collections/_form.html.slim
Dome-GER: You are using `human_attribute_name` here. This is a helper that makes the attribute more readable, but it's not a real label. You could also just define a new locale as we do for other attributes.

Dome-GER, line 12:
```suggestion
        = f.label :visibility_level, Collection.human_attribute_name("visibility_level"), class: 'form-label collection-edit-visibility'
```

config/routes.rb
Dome-GER: This is a leftover now, isn't it?

Dome-GER (commented):
Thanks for incorporating the feedback. I didn't look at the specs in detail.

MrSerth:
@kkoehn: Would you mind rebasing this PR, please? Sorry for the delay in the review; I am waiting for it, too.

Note to self: Once done, I need to set some collections to public.

MrSerth:
The failing specs are related to missing translations (i.e., for German). That should be something easy to fix :)

app/views/collections/_form.html.slim
Dome-GER: I just didn't know. So this was a deliberate decision, fine for me.

app/views/tasks/show.html.slim
```diff
@@ -478,13 +478,21 @@
...
-              - if !current_user.collections.empty?
-                - current_user.collections.reverse.each do |collection|
-                  li
-                    = link_to collection.title, add_to_collection_task_path(collection: collection), method: 'post', class: 'dropdown-item'
-              - else
+              - current_user.collections.order(created_at: :desc).each do |collection|
                 li
-                  = link_to t('.button.add_collection'), new_collection_path(current_user)
+                  - if collection.tasks.include?(@task)
+                    = link_to "<span class=\"collection-checkmark\">✓</span>#{collection.title}".html_safe, remove_task_ajax_collection_path(collection, task: @task, return_to_task: true), method: :patch, class: 'dropdown-item', data: { confirm: t('.remove_task_from_collection_warning') }
```
Dome-GER, line 484: This still looks a bit sketchy to me.

Probably, I'd go for `link_to ... do`, which is better readable and meant for such use cases.

Dome-GER (approved):

config/locales/de/views/collections.yml
```diff
@@ -29,6 +30,16 @@ de:
...
+      no_other_user: Kein anderer Benutzer hat Zugriff auf diese Aufgabensammlung
+      num_of_other_users:
+        one: Ein anderer Benutzer hat Zugriff auf diese Aufgabensammlung
+        other: "%{count} andere Benutzer haben Zugriff auf diese Aufgabensammlung"
       shared_by: Geteilt von
       temporarily_disabled: Vorübergehend deaktiviert
       type_hint: Geben Sie die E-Mail-Adresse des/der Benutzer:in ein
+      visibility:
+        private: Diese Aufgabensammlung ist Privat
+        public: Diese Aufgabensammlung ist Öffentlich
```
MrSerth, line 42:
```suggestion
        private: Diese Aufgabensammlung ist privat
        public: Diese Aufgabensammlung ist öffentlich
```

config/locales/de/views/collections.yml
```diff
@@ -29,6 +30,16 @@ de:
         deletion_warning: Sie sind der/die letzte Benutzer:in in dieser Aufgabensammlung. Wenn Sie gehen, wird diese Aufgabensammlung dauerhaft gelöscht! Sind Sie sicher?
       locked_exercise_hint: 'Hinweis: Für Aufgaben mit einem Schloss-Symbol müssen Sie zuerst Zugriff beantragen, um sie anzeigen zu können.'
       noExercisesAdded: Keine Aufgabe hinzugefügt
+      no_other_user: Kein anderer Benutzer hat Zugriff auf diese Aufgabensammlung
+      num_of_other_users:
+        one: Ein anderer Benutzer hat Zugriff auf diese Aufgabensammlung
+        other: "%{count} andere Benutzer haben Zugriff auf diese Aufgabensammlung"
```
MrSerth, line 36: In the German translation, we are using "gendergerechte Sprache" (with `:`), since CodeHarbor is mainly targeted at schools and universities. Please double check those occurrences, i.e., "Benutzer". There are other examples in this file to look at.

config/locales/de/views/tasks.yml
```diff
@@ -98,6 +99,7 @@ de:
       no_license: Keine
       no_model_solution_present: Keine Musterlösungen vorhanden
       no_tests: Keine Tests enthalten
+      remove_task_from_collection_warning: Sind Sie sicher, dass sie diese Aufgabe aus der Sammlung entfernen wollen?
```
MrSerth, line 102:
```suggestion
      remove_task_from_collection_warning: Sind Sie sicher, dass Sie diese Aufgabe aus der Sammlung entfernen wollen?
```

app/views/tasks/show.html.slim
kkoehn: You are right, this looks way better now

MrSerth: Nice, thanks for taking another look here. Personally, I think the `link_to ... do` solution now implemented is much cleaner and better to maintain 💪

app/views/tasks/show.html.slim
```diff
@@ -478,13 +478,26 @@
...
-                    = link_to collection.title, add_to_collection_task_path(collection: collection), method: 'post', class: 'dropdown-item'
-              - else
+              - current_user.collections.order(created_at: :desc).each do |collection|
                 li
-                  = link_to t('.button.add_collection'), new_collection_path(current_user)
+                  - if collection.tasks.include?(@task)
+                    = link_to remove_task_collection_path(collection, task: @task, return_to_task: true), method: :patch, class: 'dropdown-item', data: { confirm: t('.remove_task_from_collection_warning') } do
+                      span.collection-checkmark
+                        = '✓'
+                        = collection.title
```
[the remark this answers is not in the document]
kkoehn: You are right, there was a space missing

Dome-GER (approved):

Issue #1124, opened 2023-12-01
Add description to collections


MrSerth:
It would be awesome if we could add a description to collections: The description could provide some further context (such as the original course where exercises were first offered, further information, ...). The description should be viewable from the collections#index page, supporting users in finding the desired collection.

PR #1134, opened 2023-12-11
Rewriting authorization with Pundit

### LINKED ISSUE #1020: Switch from CanCanCan to Pundit
So far, we used [CanCanCan](https://github.com/CanCanCommunity/cancancan) for authorization management. This worked so far, but I also see some disadvantages (incl. missing checks for some actions, difficult ability definition, missing scopes, a difference to CodeOcean, ...). Therefore, we should evaluate and potentially switch to [Pundit](https://github.com/varvet/pundit).

Mathis-Z:
Closes #1020
I think scopes should be added as a next step. (for TaskController#index for example)
I am looking forward to your feedback :)


app/controllers/users_controller.rb
```diff
@@ -15,15 +13,17 @@ class UsersController < ApplicationController
     if current_user
       redirect_to({id: current_user.id}, alert: t('common.errors.not_found_error'))
     else
-      redirect_to root_path, alert: t('common.errors.not_authorized')
+      redirect_to root_path, alert: t('common.errors.not_found_error')
```
MrSerth, line 16: Previously, we somewhat decided by design to use `not_authorized` here. The intention was to provide as few information as possible to those who are just trying to access invalid resources. Then, just trying around with different URLs won't give you any clue whether the user exists or not. However, this mechanism wasn't working here correctly, since any user logged in would receive an 404. Ideally, we would only show admins a 404 (and otherwise allow access), and all other users (logged in or anonymous) just receive a "generic" error. Those users logged in should be, that was the original idea at that time, redirect to their own profile instead.

app/controllers/users_controller.rb
```diff
@@ -15,15 +13,17 @@ class UsersController < ApplicationController
...
     end
   end
 
   def index
     @users = User.all.paginate(page: params[:page], per_page: per_page_param)
+    authorize @users
   end
```
MrSerth, line 23: Not your fault, but we don't have an index route or view. Hence, we can probably drop this here.

app/models/task.rb
```diff
@@ -111,15 +111,15 @@ def author?(user)
...
 
   def updateable_by?(user)
-    user.nil? ? false : user.can?(:update, self)
+    user.nil? ? false : Pundit.policy(user, self).update?
   end
 
   def destroyable_by?(user)
-    user.nil? ? false : user.can?(:destroy, self)
+    user.nil? ? false : Pundit.policy(user, self).destroy?
   end
```
MrSerth, line 123: In the next slice (after we merged this PR), we should probably convert those methods to real policies as well.

app/policies/application_policy.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  private
+
+  def user_required?
+    true
+  end
+
+  def require_user!
+    raise Pundit::NotAuthorizedError unless @user
```
MrSerth, line 28: If this exception is thrown, the error message is somewhat incomplete:

```
Pundit::NotAuthorizedError: not allowed to  this NilClass
```

This is due to the default message:

```rb
        message = options.fetch(:message) { "not allowed to #{query} this #{record.class}" }
```

Therefore, we could overwrite the message here it make it more accurate for fellow devs:

```rb
    raise Pundit::NotAuthorizedError.new("my custom error message") unless @user
```

app/policies/application_policy.rb
```diff
@@ -0,0 +1,60 @@
...
+    require_user! if user_required?
+  end
+
+  def method_missing(_method_name, *_args)
+    false
+  end
+
+  def respond_to_missing?(method_name, include_private = false)
+    RESOURCE_ACTIONS.include?(method_name) || super
+  end
```
MrSerth, line 19: That's an interesting part; I am not quite sure I get the motivation correctly. I assume you wanted to provide a "fallback", in case something goes wrong. Having this fallback to be false makes sense, absolutely. 

Then, you are also overwriting the `respond_to_missing?`, which seems to be good practice when overwriting `method_missing` (I just learned that -- thanks!). Nevertheless, I now feel there is some mismatch between`.respond_to? :foo`, `.method :foo`, and `.foo`.
[Read more](https://blog.marc-andre.ca/2010/11/15/methodmissing-politely/)

```rb
[13] pry(main)> ap = ApplicationPolicy.new("bar", "foobar")
[13] pry(main)> ap.foo
=> false
[14] pry(main)> ap.method :foo
NameError: undefined method `foo' for class `ApplicationPolicy'
from (pry):14:in `method'
[15] pry(main)> ap.respond_to? :foo
=> false
```

app/policies/comment_policy.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def task_authorized?
+    if @record.respond_to?(:map)
+      @record.map(&:tasks).uniq.all? {|task| Pundit.policy(@user, task).show? }
+    else
+      Pundit.policy(@user, comment.task).show?
+    end
+  end
+
+  public
```
MrSerth, line 22: Is there any specific reason you started with the private methods above and then switched again to public? Normally, it's just the other way around in Ruby (omitting `public`).

app/views/comments/load_comments.js.erb
```diff
@@ -1,4 +1,4 @@
-id = <%= @task.id %>
```
MrSerth, line 1: What is your intention behind this change? Didn't we just ensure that the `@task` is properly authorised?

app/policies/task_policy.rb
```diff
@@ -0,0 +1,50 @@
...
+
+class TaskPolicy < ApplicationPolicy
+  def task
+    @record
+  end
+
+  def user_required?
+    false
+  end
+  private :user_required?
```
MrSerth, line 11: I would probably just make a `private` block as usual, even if it just contains this method.

app/policies/task_policy.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class TaskPolicy < ApplicationPolicy
+  def task
+    @record
+  end
```
MrSerth, line 6: I was shortly considering whether we should make this method private. However, since the `@record` variable is an `attr_reader`, the value is already accessibly publicly (on each `ApplicationPolicy` instance), so that this additional method in each policy shouldn't have any impact. Otherwise, one could argue that we don't want other callees to access it that way and hence make it private....
For simplicity, I would probably keep it.

app/policies/task_policy.rb
```diff
@@ -0,0 +1,50 @@
...
+      task.access_level_public?
+    end
+  end
+
+  %i[index? new? import_start? import_confirm? import_uuid_check? import_external?].each do |action|
+    define_method(action) { everyone }
+  end
+
+  %i[create?].each do |action|
+    define_method(action) { record_owner? || admin? } # should import_uuid_check and import_external be public?
```
MrSerth, line 26: > should import_uuid_check and import_external be public?

Previously, it was that way. However, in both actions, we are identifying users with their account link token anyway. So maybe it's just time to reflect that in the policy and perform a "proper" check.

app/policies/task_policy.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    record_owner? || task.in_same_group?(@user) || admin?
+  end
+
+  def edit?
+    update?
+  end
+
+  def destroy?
+    record_owner? || task.in_same_group_admin?(@user) || admin?
```
MrSerth, line 48: Wouldn't it be a good idea to add an early return for `@user.blank?` here as well? I haven't tested it, but from reading the code, I would say that your solution will work correctly, but maybe it's more "explicit" to state it directly (similar to `update?`).

MrSerth (commented):
Awesome work, I am really amazed by your pace and all the nice changes you did. Below is already the first batch of my review; some of the comments are well-suited for future work (similar to the scopes you mentioned, I like them as well).

app/policies/task_policy.rb
MrSerth: Currently, the import is broken, since we `require_user!` in the controller and are not setting the `current_user` accordingly.

app/controllers/messages_controller.rb
```diff
@@ -1,25 +1,28 @@
 # frozen_string_literal: true
 
 class MessagesController < ApplicationController
+  before_action :load_and_authorize_message, only: %i[destroy]
   before_action :set_user
-  before_action :set_message, only: %i[destroy]
   before_action :set_option, only: %i[index]
 
-  def index
+  def index # rubocop:disable Metrics/AbcSize
```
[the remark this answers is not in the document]
Mathis-Z: I created a ticket for this: #1141

app/models/task.rb
Mathis-Z: I created a ticket so we remember this: #1142

app/controllers/users_controller.rb
Mathis-Z: It should work as you described now.

app/policies/application_policy.rb
Mathis-Z: Good idea. I changed that.

app/policies/application_policy.rb
Mathis-Z: We need every policy to not throw a MethodNotFound for the standard actions like index?, show? etc. because we use those in views/application/_breadcrumbs_and_title.html.slim. 
We could just return false for those in the ApplicationPolicy but then every policy would have those public instance methods even when a controller does not have an action like that. This causes a problem in the policy specs because then we cannot write 
```
context 'when current_user is admin' do
      let(:current_user) { create(:admin) }

      it { is_expected.to permit_all_actions }
end
```
This would fail for the users controller because there is no index route for users which means even admins cannot access this. We would have to write something like the following instead:
```
context 'when current_user is admin' do
      let(:current_user) { create(:admin) }

      it { is_expected.to forbid_only_actions(%i[index ...]) }
end
```
This makes the specs harder to read which I wanted to avoid. The solution I came to was this little meta programming stunt. Because the index?, show?, ... actions are technically not defined the pundit_matcher permit_all_actions is happy and the breadcrumbs partial is also happy because it gets a return value for the standard actions. Maybe you have another idea, though?

app/policies/comment_policy.rb
Mathis-Z: I wanted to declare the helper methods before the actual authorization because I figured it was more readable. Since the policy is not that long anyway I switched the order like you suggested.

app/views/comments/load_comments.js.erb
Mathis-Z: That was an oversight by me. I undid that.

app/policies/task_policy.rb
Mathis-Z: I skipped require_user! for those actions because we do not want to throw an error in import_uuid_check and we authorize the task creation in import_external anyway. Could you please test if the import works now?

MrSerth: Sounds good. However, the check still fails:

Bildschirmfoto 2023-12-18 um 11 29 09

Now, we still need to change the policy (or, better, configure the `current_user` correctly).

app/policies/application_policy.rb
MrSerth: Ah, I see. Overall, the solution you've come up with is quite clever, and I like it :+1:. It fits great for both the view (breadcrumbs_and_title) as well as the specs. The only thing might be some more complicated "debugging" if something is not working or someone forgot to specify the respective method on their own. This could be something optional to include as a debug (or trace) log output. Otherwise, I think that's fine.

app/controllers/comments_controller.rb
```diff
@@ -1,15 +1,13 @@
...
-  before_action :set_task
-  before_action :set_comment, only: %i[edit update destroy]
-  before_action :new_comment, only: :create
+  before_action :load_and_authorize_task
+  before_action :load_and_authorize_comment, only: %i[edit update destroy]
+  skip_before_action :require_user!, only: %i[index]
 
   def index
     @comments = Comment.where(task: @task).paginate(page: params[:page], per_page: per_page_param).order(created_at: :desc)
+    authorize @comments, :index?
```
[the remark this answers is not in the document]
Mathis-Z: In create, update, etc. we render the index route on success which means authorize @comments would be called indirectly from create/update/... which calls the policy methods create?/update?/... and those cannot handle an ActiveRecord:Relation object.

app/policies/task_policy.rb
Mathis-Z: It should work now. Could you test again please?

app/views/comments/load_comments.js.erb
MrSerth: Thanks for reverting that. I'm fine with that change, but in general would rather like to have the previous change be removed from Git (i.e., amending the commit that introduced the change). No need to do it here, but maybe for future occasions...

app/policies/task_policy.rb
MrSerth: Yes, it works now 🎉

app/controllers/comments_controller.rb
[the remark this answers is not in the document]
MrSerth: Ah, I see. This is rather ... unexpected, but probably I reviewed that 🥲. Alright, I've got it now and think it's the best solution for the time being. Shall we add a comment, hinting on the usage as part of the create/update/... routes?

app/policies/task_policy.rb
```diff
@@ -0,0 +1,57 @@
...
+  def show?
+    if @user.present?
+      admin? || task.access_level_public? || task.in_same_group?(@user) || task.author?(@user)
+    else
+      task.access_level_public?
+    end
+  end
+
+  %i[import_uuid_check? import_external?].each do |action|
+    define_method(action) { true } # allow unauthenticated access for API routes
```
MrSerth, line 17: Since we are authenticated with a `current_user` in the action (and by moving that to a `before_filter` that is performed before `#only_authorize_action`), we should also be able to switch that to `everyone` and get rid of the comment.

MrSerth (commented):
Awesome, thanks! Here's my next round of review. 

Many thanks as well for increasing the test coverage recently. There are [a few minor lines not completely covered yet](https://app.codecov.io/gh/openHPI/codeharbor/pull/1134) (but touched), which might make sense in this PR (due to the nature of authorizations and their security implication). For example, this includes:
- the new JSON response in `import_confirm`,
- the `not_authorized` error in the UsersController for anonymous users (and the index page)
- or the `method_missing` in the `application_policy` (which, by the way, is a good thing!). Do you think we can add some smart mechanism to discover that the `method_missing` is not called by any other test, to ensure we don't forget anything? Or is that not needed?

spec/policies/message_policy_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    context 'when user is admin' do
+      let(:user) { create(:admin) }
+
+      it { is_expected.to permit_all_actions }
+    end
+
+    context 'when user is recipient' do
+      let(:user) { recipient }
+
+      it { is_expected.to permit_only_actions(%i[index new destroy]) }
```
MrSerth, line 28: Why is this user not to reply? Or, phrasing this question differently: If I read the specs correctly, no one except an admin is allowed to reply. Why? Maybe we are just missing some specs here...

spec/policies/task_file_policy_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    context 'when task is from user' do
+      let(:task_user) { user }
+
+      it { is_expected.to permit_only_actions(%i[download_attachment extract_text_data]) }
+    end
+
+    context 'when task is in same group' do
+      let(:user) { create(:user) }
+      let(:group_memberships) { [build(:group_membership, :with_admin), build(:group_membership, user:, role: :confirmed_member)] }
+      let(:groups) { create_list(:group, 1, group_memberships:) }
```
MrSerth, line 37: Potentially, we could use this mechanism in the `group_policy_specs.rb` as well, rather than the `before` block there.

spec/policies/task_file_policy_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+        it { is_expected.to permit_only_actions(%i[download_attachment extract_text_data]) }
+      end
+    end
+  end
+
+  context 'with a user and public task' do
+    let(:user) { create(:user) }
+    let(:access_level) { :public }
+
+    # why only permit downloading attachments when the text could be extracted after downloading anyway?
```
MrSerth, line 59: Good question. Originally, `extract_text` is designed to be part of an update operation, thus requiring write access to the task (file). It got introduced by following the ProFormA specification, which differentiates between embedded and attached text (and binary) files. Therefore, this action is designed to convert attached text files to embedded text.

However, when digging deeper into the implementation, it seems like the actual action `extract_text` is just returning the file content for an AJAX operation (which, in turn, will use the response to set the file content in the frontend, followed by a regular "Update" operation triggered through the user). This, in turn, raises the question why extracting is not permitted, but downloading is. If you ask me, both operations might behave similar (in the current implementation, where extracting text is mainly a frontend operation).

spec/policies/task_policy_spec.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe TaskPolicy do
+  subject { described_class.new(user, task) }
+
+  let(:user) { nil }
```
MrSerth, line 8: Here (and in some other policy specs above): We might move this line to the `context 'without a user'`.

spec/policies/user_policy_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+require 'rails_helper'
+
+RSpec.describe UserPolicy do
+  subject { described_class.new(current_user, user) }
+
+  let(:current_user) { nil }
+  let(:user) { create(:user) }
+
+  context 'with a current_user' do
```
MrSerth, line 11: Here (and, similar in some other policy specs): Wouldn't we need to test the policy without a user as well (i.e., ensuring that it doesn't allow anything)?

MrSerth (commented):
Finally, here's my last round of feedback. Well done!

app/controllers/tasks_controller.rb
```diff
@@ -3,17 +3,20 @@
...
 
 class TasksController < ApplicationController # rubocop:disable Metrics/ClassLength
-  load_and_authorize_resource except: %i[import_external import_uuid_check]
+  before_action :load_and_authorize_task, except: %i[index new create import_start import_confirm import_uuid_check import_external]
+  before_action :only_authorize_action, only: %i[import_start import_confirm import_uuid_check import_external]
 
   before_action :handle_search_params, only: :index
   before_action :set_search, only: [:index]
   skip_before_action :verify_authenticity_token, only: %i[import_external import_uuid_check]
+  skip_before_action :require_user!, only: %i[show import_external import_uuid_check]
```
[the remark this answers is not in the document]
Mathis-Z: If we re-enable the require_user! before action and get an API request without a user we would return an error but not a JSON. I am not sure how we could change that?

app/controllers/users_controller.rb
```diff
@@ -13,17 +11,16 @@ class UsersController < ApplicationController
...
     end
   end
 
-  def index
-    @users = User.all.paginate(page: params[:page], per_page: per_page_param)
-  end
-
   def show
     @user = User.find(params[:id])
+    authorize @user
```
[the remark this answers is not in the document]
Mathis-Z: Yes, I will do that.

app/controllers/comments_controller.rb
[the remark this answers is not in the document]
Mathis-Z: I will add a comment :+1:

app/policies/task_policy.rb
Mathis-Z: That would throw an error in case an API request without a user comes in. This would again not return a JSON but instead the pundit NotAuthorizedError. I dont know how we would get around that.

spec/policies/comment_policy_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+require 'rails_helper'
+
+RSpec.describe CommentPolicy do
+  subject { described_class.new(user, comment) }
+
+  let(:user) { nil }
+  let(:comment_user) { create(:user) }
+  let(:task_user) { create(:user) }
+  let(:access_level) { :private }
```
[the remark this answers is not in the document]
Mathis-Z: I am not sure which line you mean but I think all of them are used. Am I missing something? The :user is used in the subject definition, the :comment_user in the comment and the :task_user and :access_level are used in the :task which is needed for the comment. We could inline the :task_user in the creation of :task since we dont change :task_user in the specs.

spec/policies/message_policy_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+require 'rails_helper'
+
+RSpec.describe MessagePolicy do
+  subject { described_class.new(user, message) }
+
+  let(:user) { nil }
+  let(:recipient) { create(:user) }
+  let(:sender) { create(:user) }
+  let(:message) { create(:message, sender:, recipient:) }
+  let(:role) { 'member' }
```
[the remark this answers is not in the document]
Mathis-Z: That is true. I will remove that :)

spec/policies/message_policy_spec.rb
Mathis-Z: Yes we are missing some specs here. The thing is that the reply permission for a message does not mean you are allowed to reply **to** this message but that you are allowed to reply **with** this message. Since you cannot send replies with yourself as recipient the user is not given the reply permission here. I will write another test covering the reply action for common users.

spec/policies/task_file_policy_spec.rb
Mathis-Z: I will change that.

spec/policies/task_file_policy_spec.rb
Mathis-Z: So should I keep it as is or allow extracting text for all users that can download the file?

spec/policies/task_policy_spec.rb
Mathis-Z: I moved the line and also removed it from the policies that dont have a 'without a user' context anyway.

spec/policies/user_policy_spec.rb
Mathis-Z: I thought we can omit those because the application_policy_spec.rb already tests that and for the policies that override the standard behavior we have these tests. However since it does not really cost anything and its just a couple lines I added a test for all policies.

app/controllers/tasks_controller.rb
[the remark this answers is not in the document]
MrSerth: We already have a JSON handler as part of the `render_error` method. Hence, we just need to ensure to call that. Unfortunately, external users of the API are not necessarily calling a JSON format (i.e., clients such as CodeOcean).

In order to achieve the desired behaviour, we just [define a default in the `routes.rb`](https://guides.rubyonrails.org/routing.html#defining-defaults). When doing so, I correctly got the desired behaviour -- besides the fact that CodeOcean just uses the whole body instead of parsing the JSON file (which I would consider to be part of a follow-up ticket, since it needs to be checked in CodeHarbor and CodeOcean for any of these actions). 

Bildschirmfoto 2024-01-02 um 10 32 50

app/policies/task_policy.rb
MrSerth: Same as above. In the `ApplicationController`, we catch any occurrence of the `Pundit::NotAuthorizedError`:

https://github.com/Mathis-Z/codeharbor/blob/499405041763852022acc6ac10718eaaa4e8b490/app/controllers/application_controller.rb#L17

Then, the `render_error` method is called:

https://github.com/Mathis-Z/codeharbor/blob/499405041763852022acc6ac10718eaaa4e8b490/app/controllers/application_controller.rb#L61-L63

Which, in turn, either redirects for any format except JSON or renders a JSON:

https://github.com/Mathis-Z/codeharbor/blob/499405041763852022acc6ac10718eaaa4e8b490/app/controllers/application_controller.rb#L73-L89

spec/policies/comment_policy_spec.rb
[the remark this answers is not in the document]
MrSerth: Indeed, all those variables are used throughout the test file. 

Actually, I haven't noticed the `task_user` thing, where we could, indeed, inline the assignment. But I wanted to double-check with the `access_level` defined in line 11 (formerly) / 10 (as of yesterday). Let me explain shortly:

The test file has two main contexts, depending on the fact whether a user can access or cannot access the task. In the first line of each context, you define an `access_level`. Hence, in those two contexts, the `access_level` is given through these "inner" definitions. Besides these two contexts, there is no other test. Hence, the "outer" `access_level` is never used to create a task (but only the two "inner" ones). As a consequence, we can omit the "outer" `access_level`. Due to the nature of RSpec, the order of `let` definitions is not important, allowing to use a name in "outer" context, even though it is only defined in (all) "inner" contexts.

spec/policies/message_policy_spec.rb
MrSerth: Ah, I see. Thanks for explaining and adding another spec there!

spec/policies/task_file_policy_spec.rb
MrSerth: Let's allow extracting text for anyone who can download the file. When doing so, please add a short comment to that, explaining why extracting is allowed. Potentially, you can also link to my previous comment (but that's not required, if you find a well and short explanation 😉).

spec/policies/user_policy_spec.rb
MrSerth: Perfect, thank you! In general, I would agree with your optimisation and agree that testing the super class / method would be enough. However, for this security context, I'd rather opt to duplicate a few lines in tests to make it very explicit, since this will raise awareness if anything changes unexpectedly.

app/controllers/tasks_controller.rb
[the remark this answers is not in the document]
Mathis-Z: Alright, it was a bit tricky to get the order of before_actions right because I have to set the api user before require_user! is executed. Since the before_action require_user! is defined in the application controller I had to override it. I will push that in a moment.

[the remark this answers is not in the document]
MrSerth: Ah, of course! Besides re-defining a `before_action` with the same name (what you currently do), you can also change the order of the actions and moving `:set_user_for_api_request` to the very first spot. This is done by using `prepend_before_action` rather than `before_action`. [This blog post](https://reganchan.ca/blog/ordering-of-filters-in-rails-controllers/) highlights these differences and the inheritance.

Since prepending will always work for any action to be defined, this is probably the way to go. 

Finally, we can probably drop the comment about "unauthenticated access" in the `TaskPolicy`, can't we?

app/policies/application_policy.rb
Mathis-Z: I added debug output like you suggested.

app/controllers/tasks_controller.rb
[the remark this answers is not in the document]
Mathis-Z: Yes the comment can be dropped. I also switched to prepend_before_action :+1:

Mathis-Z:
I removed the not_authorized json response from TasksController::import_confirm because there is currently no case in which this error could be triggered. If the user is nil then the before_action require_user! will catch that and redirect. If the user is not nil it will always be allowed to create the task. I thought about stubbing the policy to write a test but I don't think that would make sense because we would rather have this case generate an error in Sentry so we can fix it.

Mathis-Z:
Regarding the line in UsersController not yet covered by tests I think we don't need a test for that. The line never gets reached anyway because if there is no current_user the before_action require_user! will already have redirected. We might consider removing the line though.

MrSerth:
> I removed the not_authorized json response from TasksController::import_confirm because there is currently no case in which this error could be triggered. If the user is nil then the before_action require_user! will catch that and redirect. If the user is not nil it will always be allowed to create the task. I thought about stubbing the policy to write a test but I don't think that would make sense because we would rather have this case generate an error in Sentry so we can fix it.

I am fine with that :+1:. If, in the current case, we argue that this case cannot happen, we should remove it. When we learn later (through Sentry) that some specific edge case was not considered, we'll add a regression test and fix the behaviour.

MrSerth:
> Regarding the line in UsersController not yet covered by tests I think we don't need a test for that. The line never gets reached anyway because if there is no current_user the before_action require_user! will already have redirected. We might consider removing the line though.

Right, I agree. Then, let's drop this line (together with the conditional) and always redirect as we currently do.

MrSerth:
Great stuff, many thanks for your hard work here!

I think, we're almost good to go. I see these three points left:
- Since I've merged #1124 first, I accidentally created a conflict 🙈, which we should resolve first.
- Let's address the RecordNotFound exception in the UsersController, as mentioned [previously](https://github.com/openHPI/codeharbor/pull/1134#issuecomment-1881943116).
- Ideally add a final test for the MessagesController#new, since there was no existing one yet. Since this is a sub-resource nested below a user, we probably need to test cases (one with a successful and one with an unsuccessful authorise) and maybe one for an admin (besides the fact that it's not making use of the user ID given in the path).

Then, I am more than happy to merge and deploy these changes 💪

Mathis-Z:
There is no case in which the authorization in MessagesController#new can fail because the policy allows this for 'everyone'. I added a simple test though.

app/views/collections/_collections.html.slim
```diff
@@ -16,29 +16,14 @@
...
-              = render_markdown(collection.description)
-          .more-btn-wrapper
-            = button_to collection, class: 'more-btn' do
-              .more-tag
-                = t('common.button.show_more')
-                span.fa-solid.fa-caret-down.ms-1
-              .less-tag.d-none
-                = t('common.button.show_less')
-                span.fa-solid.fa-caret-up.ms-1
-          br
```
MrSerth, line 34: Aren't you removing the collection description here, adding in the previous PR #1124? To me, this seems like a mistake caused by Git...

MrSerth (commented):
> There is no case in which the authorization in MessagesController#new can fail because the policy allows this for 'everyone'. I added a simple test though.

Thanks for adding a test, that looks good! Regarding the failing authorization I (somehow) expected: For sure, a `new` template is usually renderable by everyone. However, in the messages controller, we perform the `set_user` method before any action. Shouldn't this be somehow protected, i.e., when accessing an URL with the ID of someone else as a regular non-admin user? Maybe I don't get the usage of the `@user` there, it mainly seems to be used for the construction of routes, but without actually being useful. Do you think we should keep the nesting of the messages controller as part of the user resources? 

Sorry to raise this discussion again here, it's something present for quite a while already, but obviously never draw my attention before...

MrSerth (approved):
👏 :shipit: 🚀

PR #1143, opened 2023-12-13
Translate CodeHarbor to German

MrSerth:
This PR adds the current status of the translation of CodeHarbor to German.

As specified in the first commit, the translations provided here were not checked manually yet and therefore contain some inconsistencies, such as the salutation ("Du" vs "Sie") or the gender form. Before merging, we should unify those (probably "Sie" and gendered nouns with `:` [e.g., "Empfänger:in"]). Further, I omitted the "external" translations, since we can probably borrow them from some other location without the need to review them.

The other changes beside the locale files are probably enough to enable an additional language for CodeHarbor and allow users to switch. These parts of the code originate from CodeOcean and are only adjusted where necessary.

### DIFF
--- a/app/assets/javascripts/tasks_index.coffee
+++ b/app/assets/javascripts/tasks_index.coffee
@@ -1,5 +1,4 @@
 ready = ->
-  initializeSelect2()
   initCollapsable($('.description'), '95px')
   window.addEventListener 'resize', -> initCollapsable($('.description'), '95px')
   initializeDynamicHideShow()
@@ -7,7 +6,6 @@ ready = ->
   initializeIndexComments()
   initializeInputFieldEnterCallback()
 
-$(document).on('turbolinks:load', ready)
 
 initializeDynamicHideShow = ->
   $('body').on 'click', '.more-btn-wrapper', (event) ->
@@ -103,13 +101,7 @@ intializeAdvancedFilter = ->
       $drop.removeClass('fa-caret-down').addClass('fa-caret-up')
 
   if $('#order_param')
-#    $('#' + order.value).addClass('active')
     $('#order_created').addClass('active')
-
-  #  $('#order_rating').click ->
-  #    $('#order_rating').addClass('active')
-  #    $('#order_created').removeClass('active')
-  #    document.getElementById('order_param').value = 'order_rating'
 
   $('#order_created').click ->
     $('#order_created').addClass('active')
@@ -144,3 +136,7 @@ initializeInputFieldEnterCallback = ->
       return
     $('.search-submit-button-tag').click();
     event.preventDefault();
+
+
+$(document).on('turbolinks:load', ready)
+$(document).on('select2:locales:loaded', initializeSelect2)
--- a/app/assets/javascripts/tasks_index.coffee
+++ b/app/assets/javascripts/tasks_index.coffee
@@ -16,10 +16,12 @@ initializeDynamicHideShow = ->
 
 initializeSelect2 = ->
   $('.defaultSelect2').select2
+    language: I18n.locale
     minimumResultsForSearch: 10
     width: '100%'
 
   $('.language-box').select2
+    language: I18n.locale
     width: '100%'
     multiple: true
     closeOnSelect: false
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -97,10 +97,35 @@ class ApplicationController < ActionController::Base
     end
   end
 
-  def switch_locale(&) # rubocop:disable Metrics/AbcSize
-    session[:locale] = sanitize_locale(params[:custom_locale] || params[:locale] || session[:locale])
-    locale = session[:locale] || http_accept_language.compatible_language_from(I18n.available_locales) || I18n.default_locale
+  def sanitized_locale_param
+    sanitize_locale(params[:locale])
+  end
+
+  def sanitized_session_locale
+    sanitize_locale(session[:locale])
+  end
+
+  def sanitized_user_preferred_locale
+    return if current_user.nil?
+
+    sanitize_locale(current_user.preferred_locale)
+  end
+
+  def choose_locale
+    sanitized_locale_param ||
+      sanitized_session_locale ||
+      sanitized_user_preferred_locale ||
+      http_accept_language.compatible_language_from(I18n.available_locales) ||
+      I18n.default_locale
+  end
+
+  def switch_locale(&)
+    locale = choose_locale
+    session[:locale] = locale
     Sentry.set_extras(locale:)
+    if current_user.present? && locale != sanitized_user_preferred_locale
+      current_user.update(preferred_locale: locale)
+    end
     I18n.with_locale(locale, &)
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,7 +27,7 @@ class ApplicationController < ActionController::Base
   def flash_to_headers
     return unless request.xhr?
 
-    response.headers['X-Message'] = flash_message
+    response.headers['X-Message'] = ERB::Util.url_encode(flash_message)
     response.headers['X-Message-Type'] = flash_type.to_s
 
     flash.discard # don't want the flash to appear when you reload page
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -61,7 +61,7 @@ class GroupsController < ApplicationController
   def request_access
     flash[:notice] = t('.success_notice')
     @group.admins.each do |admin|
-      send_access_request_message(admin, @group)
+      send_message(admin, :group_requested)
 
       AccessRequestMailer.send_access_request(current_user, admin, @group).deliver_now
     end
@@ -77,9 +77,9 @@ class GroupsController < ApplicationController
 
   def grant_access
     @group.grant_access(@user)
-    send_grant_access_messages(@user, @group)
+    send_message(@user, :group_accepted)
 
-    Message.where(sender: @user, recipient: current_user, param_type: 'group', param_id: @group.id).destroy_all
+    Message.where(sender: @user, recipient: current_user, param_type: :group_requested, param_id: @group.id).destroy_all
     redirect_to @group, notice: t('.success_notice')
   end
 
@@ -90,9 +90,9 @@ class GroupsController < ApplicationController
 
   def deny_access
     @group.users.delete(@user)
-    send_deny_access_message(@user, @group)
+    send_message(@user, :group_declined)
 
-    Message.where(sender: @user, recipient: current_user, param_type: 'group', param_id: @group.id).destroy_all
+    Message.where(sender: @user, recipient: current_user, param_type: :group_requested, param_id: @group.id).destroy_all
     redirect_to @group, notice: t('.success_notice')
   end
 
@@ -108,30 +108,12 @@ class GroupsController < ApplicationController
 
   private
 
-  def send_access_request_message(admin, group)
+  # type is one of [:group_requested, :group_accepted, :group_declined]
+  def send_message(recipient, type)
     Message.create(sender: current_user,
-      recipient: admin,
-      text: t('groups.send_access_request_message.message', user: current_user.name, group: group.name),
-      param_type: 'group',
-      param_id: group.id,
-      sender_status: 'd')
-  end
-
-  def send_deny_access_message(user, group)
-    Message.create(sender: current_user,
-      recipient: user,
-      text: t('groups.send_deny_access_message.message', user: current_user.name, group: group.name),
-      param_type: 'group_declined',
-      param_id: group.id,
-      sender_status: 'd')
-  end
-
-  def send_grant_access_messages(user, group)
-    Message.create(sender: current_user,
-      recipient: user,
-      text: t('groups.send_grant_access_messages.message', user: current_user.name, group: group.name),
-      param_type: 'group_accepted',
-      param_id: group.id,
+      recipient:,
+      param_type: type,
+      param_id: @group.id,
       sender_status: 'd')
   end
 
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -122,6 +122,7 @@ class GroupsController < ApplicationController
       recipient: user,
       text: t('groups.send_deny_access_message.message', user: current_user.name, group: group.name),
       param_type: 'group_declined',
+      param_id: group.id,
       sender_status: 'd')
   end
 
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -47,7 +47,12 @@ import 'jquery-ui/themes/base/sortable.css'
 // I18n locales
 import { I18n } from "i18n-js";
 import locales from "../../tmp/locales.json";
-Object.keys(locales).forEach(locale => import(`select2/dist/js/i18n/${locale}`));
+
+Promise.all(
+    Object.keys(locales).map(locale => import(`select2/dist/js/i18n/${locale}`))
+).then(() =>
+    document.dispatchEvent(new Event("select2:locales:loaded"))
+);
 
 // Fetch user locale from html#lang.
 // This value is being set on `app/views/layouts/application.html.slim` and
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -47,6 +47,7 @@ import 'jquery-ui/themes/base/sortable.css'
 // I18n locales
 import { I18n } from "i18n-js";
 import locales from "../../tmp/locales.json";
+Object.keys(locales).forEach(locale => import(`select2/dist/js/i18n/${locale}`));
 
 // Fetch user locale from html#lang.
 // This value is being set on `app/views/layouts/application.html.slim` and
--- a/app/mailers/access_request_mailer.rb
+++ b/app/mailers/access_request_mailer.rb
@@ -5,13 +5,17 @@ class AccessRequestMailer < ApplicationMailer
     @user = user
     @admin = admin
     @group = group
-    mail(to: @admin.email, subject: "#{user.name} wants to access your Group '#{group.name}'")
+    I18n.with_locale(@admin.preferred_locale || I18n.default_locale) do
+      mail(to: @admin.email, subject: t('groups.access_request_mailer.subject', user: @user.name, group: @group.name))
+    end
   end
 
   def send_contribution_request(author, exercise, user)
     @author = author
     @exercise = exercise
     @user = user
-    mail(to: @author.email, subject: "#{user.name} wants to contribute to your Exercise '#{exercise.title}'")
+    I18n.with_locale(@author.preferred_locale || I18n.default_locale) do
+      mail(to: @author.email, subject: t('tasks.access_request_mailer.subject', user: @user.name, task: @exercise.title))
+    end
   end
 end
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Message < ApplicationRecord
-  validates :text, presence: true
+  validates :text, presence: true, unless: -> { %w[group_requested group_accepted group_declined collection].include?(param_type) }
 
   belongs_to :sender, class_name: 'User', inverse_of: :sent_messages
   belongs_to :recipient, class_name: 'User', inverse_of: :received_messages
@@ -18,7 +18,7 @@ class Message < ApplicationRecord
 
   def text # rubocop:disable Metrics/AbcSize
     case param_type
-      when 'group'
+      when 'group_requested'
         I18n.t('groups.send_access_request_message.message', user: sender.name, group: Group.find(param_id).name)
       when 'group_accepted'
         I18n.t('groups.send_grant_access_messages.message', user: sender.name, group: Group.find(param_id).name)
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -16,6 +16,21 @@ class Message < ApplicationRecord
     self.recipient_status = 'd' if recipient == user
   end
 
+  def text # rubocop:disable Metrics/AbcSize
+    case param_type
+      when 'group'
+        I18n.t('groups.send_access_request_message.message', user: sender.name, group: Group.find(param_id).name)
+      when 'group_accepted'
+        I18n.t('groups.send_grant_access_messages.message', user: sender.name, group: Group.find(param_id).name)
+      when 'group_declined'
+        I18n.t('groups.send_deny_access_message.message', user: sender.name, group: Group.find(param_id).name)
+      when 'collection'
+        I18n.t('collections.share_message.text', user: sender.name, collection: Collection.find(param_id).title)
+      else
+        super
+    end
+  end
+
   def self.parent_resource
     User
   end
--- a/config/locales/de/external/devise.yml
+++ b/config/locales/de/external/devise.yml
@@ -1,0 +1,148 @@
+---
+de:
+  activerecord:
+    attributes:
+      user:
+        confirmation_sent_at: Bestätigung gesendet am
+        confirmation_token: Bestätigungs-Token
+        confirmed_at: Bestätigt am
+        created_at: Erstellt am
+        current_password: Bisheriges Passwort
+        current_sign_in_at: Aktuelle Anmeldung am
+        current_sign_in_ip: IP der aktuellen Anmeldung
+        email: E-Mail
+        encrypted_password: Verschlüsseltes Passwort
+        failed_attempts: Fehlversuche
+        last_sign_in_at: Letzte Anmeldung am
+        last_sign_in_ip: IP der letzten Anmeldung
+        locked_at: Gesperrt am
+        password: Passwort
+        password_confirmation: Passwortbestätigung
+        remember_created_at: '"Angemeldet bleiben" erstellt am'
+        remember_me: Angemeldet bleiben
+        reset_password_sent_at: E-Mail zum Zurücksetzen des Passworts gesendet am
+        reset_password_token: Token zum Zurücksetzen des Passworts
+        sign_in_count: Anzahl der Anmeldungen
+        unconfirmed_email: Unbestätigte E-Mail-Adresse
+        unlock_token: Entsperrungs-Token
+        updated_at: Aktualisiert am
+    models:
+      user:
+        one: Benutzer:in
+        other: Benutzer:innen
+  devise:
+    confirmations:
+      confirmed: Ihre E-Mail-Adresse wurde erfolgreich bestätigt.
+      new:
+        resend_confirmation_instructions: Bestätigungsanleitung erneut senden
+      send_instructions: In wenigen Minuten erhalten Sie eine E-Mail zur Bestätigung Ihrer Registrierung.
+      send_paranoid_instructions: Falls Ihre E-Mail-Adresse in unserer Datenbank existiert, erhalten Sie in wenigen Minuten eine E-Mail zur Bestätigung Ihrer Registrierung.
+    failure:
+      already_authenticated: Sie sind bereits angemeldet.
+      inactive: Ihr Account ist noch nicht aktiv.
+      invalid: "%{authentication_keys} oder Passwort ist ungültig."
+      last_attempt: Sie haben noch einen Versuch, bevor Ihr Account gesperrt wird.
+      locked: Ihr Account ist gesperrt.
+      not_found_in_database: "%{authentication_keys} oder Passwort ist ungültig."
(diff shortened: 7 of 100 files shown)


Mathis-Z:
I encountered a problem with the flash messages. Apparently there is an encoding problem for German characters "äöü". The flash messages are transferred using HTTP headers which are supposed to be ASCII encoded. I am not sure how we should solve this. One possibility would be to encode flash messages with base64 and decode them on the frontend. However that is not exactly ideal because we could not use the normal flash messages anymore but replace them with a custom wrapper doing the base64 encoding.
Do you have another idea?

MrSerth:
Thanks for describing the flash message encoding issue. I think one of your solutions is fine (e.g., the URL encoding). The problem here is that the mechanism chosen with the `X-Message` is already a custom and non-Rails solution only required for some of the AJAX actions.

Mathis-Z:
How do we want to translate the mails users receive when somebody requests access to join their group? The problem I see is that we do not know the preferred locale of the receiver at the time we send the mail. Maybe we should store the last used locale for users? We could also use this to automatically set the last used locale when a user signs in.

MrSerth:
> How do we want to translate the mails users receive when somebody requests access to join their group? 

Generally speaking, I see two possibilities here: Either, sending the messages in both languages ("German version below") or storing the desired locale in the user profile (this is the mechanism we chose for openHPI). If you want to go the extra mile, storing it in the profile and updating it on each locale change would hence be the "better" option.

Mathis-Z:
I would go with the second option. I can probably take some code from CodeOcean so this should not be too difficult.

app/models/message.rb
```diff
@@ -16,6 +16,21 @@ def mark_as_deleted(user)
...
+        I18n.t('groups.send_grant_access_messages.message', user: sender.name, group: Group.find(param_id).name)
+      when 'group_declined'
+        I18n.t('groups.send_deny_access_message.message', user: sender.name, group: Group.find(param_id).name)
+      when 'collection'
+        I18n.t('collections.share_message.text', user: sender.name, collection: Collection.find(param_id).title)
+      else
+        super
+    end
+  end
+
```
Mathis-Z, line 33: Maybe you have another idea here? When we store the preferred locale for a user in the database we could also use this to translate the messages before they get written to the database and then remove this part here.

Mathis-Z:
I pushed my current state. Except for the translations of the emails and saving the preferred locale for users this should be ready.

app/models/message.rb
MrSerth: Indeed, we could store the translated message already. However, I would argue that this approach has some drawbacks, since the message is then "fixed". For example, the message won't be updated to another language when the user switches the locale. Further, we cannot update the message either, e.g., when rephrasing the message text. Hence, I would probably keep your current solution, since it is more flexible.

MrSerth:
Awesome, thanks! I was just checking the locales and have a few suggestions as attached: [Locale_Feedback.patch](https://github.com/openHPI/codeharbor/files/14355887/Locale_Feedback.patch)

Otherwise, I noticed the following few things:

- The tooltips when hovering over the configuration elements in tests are not yet localised.
  Hover

- We still need to consider a way to use gender-neutral language
- Many translations in the `users.yml` are just taken from Devise, we can potentially omit these and use the external Devise translations. This can also be extracted to a dedicated issue.
- The user's role (on show) is not translated, since we just display the database value.
- The account link documentation is not up-to-date (at `/home/account_link_documentation`); that's a separate issue for sure.

config/locales/de/external/devise.yml
```diff
@@ -28,8 +28,8 @@ de:
         updated_at: Aktualisiert am
     models:
       user:
-        one: Benutzer
-        other: Benutzer
+        one: Benutzer:in
+        other: Benutzer:in
```
MrSerth, line 32: Isn't this the plural form?

config/locales/de/models.yml
```diff
@@ -8,20 +8,20 @@ de:
         name: Name
         push_url: Push-URL
       account_link_user:
-        user: Benutzer
+        user: Benutzer:in
       collection:
         description: Beschreibung
         title: Titel
-        users: Benutzer
+        users: Benutzer:in
```
MrSerth, line 15: Without checking, this sounds like a plural form, too.

config/locales/de/models.yml
```diff
@@ -121,17 +121,17 @@ de:
         one: Account-Link
         other: Account-Links
       account_link_user:
-        one: Account-Link-Benutzer
-        other: Account-Link-Benutzer
+        one: Account-Link-Benutzer:in
+        other: Account-Link-Benutzer:in
```
MrSerth, line 125: As for all other occurrences of "other" (won't comment on each): Plural?

config/locales/de/views/users.yml
```diff
@@ -1,82 +1,34 @@
...
-        greeting: Sehr geehrte/r %{name},
-        p1: Wir kontaktieren Sie, um Ihnen mitzuteilen, dass ihr Passwort für CodeHarbor geändert wurde.
+        p1: Wir möchten Sie darüber informieren, dass Ihr Passwort geändert wurde.
       reset_password_instructions:
-        action: Ändere mein Passwort
-        greeting: Sehr geehrte/r %{name},
-        p1: Jemand hat einen Link angefordert, um Ihr Passwort für CodeHarbor zu ändern. Sie können das mit dem Link unten tun.
-        p2: Wenn Sie das nicht waren, ignorieren Sie bitte diese E-Mail.
-        p3: Ihr Passwort wird nicht geändert, bis Sie über den Link oben ein neues erstellen.
+        p1: Jemand hat einen Link zum Ändern Ihres Passworts angefordert. Klicken Sie auf den folgenden Link, um Ihr Passwort zu ändern.
```
MrSerth, line 13: I assume that those `p1` and `p2` translations can be replaced by Devise, too.

MrSerth (commented):
Just a quick review....

config/locales/de/external/devise.yml
Mathis-Z: Oops... I updated that.

config/locales/de/views/users.yml
Mathis-Z: They are slightly different from the Devise translations so I kept them but I can replace them too if you like.

app/controllers/rails_admin_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 class RailsAdminController < ApplicationController
   # RailsAdmin does not include translations. Therefore, we fallback to English locales
-  skip_around_action :switch_locale
+  skip_before_action :set_locale
```
Mathis-Z, line 5: Not sure if thats okay but the around_action did not work with the new set_locale code. Do you know if the around_action is required for something?

app/controllers/groups_controller.rb
```diff
@@ -122,6 +122,7 @@ def send_deny_access_message(user, group)
       recipient: user,
       text: t('groups.send_deny_access_message.message', user: current_user.name, group: group.name),
```
Mathis-Z, line 123: Previously I set the locale to English here to make sure the message was in English in the database. This did not work however. I think Rails called the text getter of Message before writing the data to the database. Since the message gets translated in the text getter it would then not necessarily be in English in the database. I thought about adding a parameter "translate: false" to the getter but I think it is not that important to have the messages in a unified language in the database since they always get translated anyways.

app/controllers/application_controller.rb
```diff
@@ -97,11 +97,36 @@ def mnemosyne_trace
...
+    sanitize_locale(current_user.preferred_locale)
+  end
+
+  def choose_locale
+    (sanitized_locale_param ||
+      sanitized_session_locale ||
+      users_preferred_locale ||
+      http_accept_language.compatible_language_from(I18n.available_locales) ||
+      I18n.default_locale).to_sym
+  end
```
Mathis-Z, line 120: Since I added a not-NULL constraint to the preferred_locale-column we dont really need the browser locale and the default locale here. I wanted that constraint because I thought it would be nice to always have something valid in there. On the other hand we then cannot know whether the preferred locale of the user is actually :en or if the user just has not chosen his/her locale yet. I think the only case where this is not ideal is when a user has :de set as their preferred locale in their browser settings and the user first signs up. In this case we would take the default locale :en from the database and think this was the preferred locale of the user. However as soon as the user selects :de as the locale everything works again.

app/mailers/access_request_mailer.rb
```diff
@@ -5,13 +5,17 @@ def send_access_request(user, admin, group)
...
   end
 
   def send_contribution_request(author, exercise, user)
     @author = author
     @exercise = exercise
     @user = user
-    mail(to: @author.email, subject: "#{user.name} wants to contribute to your Exercise '#{exercise.title}'")
+    I18n.with_locale @author.preferred_locale do
+      mail(to: @author.email, subject: t('tasks.access_request_mailer.subject', user: @user.name, task: @exercise.title))
+    end
```
Mathis-Z, line 19: This seems to be unused. Should we remove this or is this used by Florian for the contributions?

app/controllers/application_controller.rb
MrSerth: Using a not-NULL constraint is fine for me, makes sense. The other behavior, with a fallback to English on user sign up, however, is a little unexpected. While I thought we could explicitly store the "current" locale on sign up, that's not easily possible for SAML sign ins. This case might need dedicated handling in the omniauth_controller to work similar. The latter further requires that we also receive the user's current cookie on the omniauth callback, which I am currently not 100% sure about (since a user's request is originating from a third-party domain). 

In short: I would actually prefer if we could store the user's language on sign ups, too.

app/controllers/groups_controller.rb
MrSerth: Wouldn't it be possible to store the I18n key and the hash for parameters (if required) to the database? Or, just store some form of enum, so that we have a mapping from the enum to the locale key separately? 

I think, always storing the English locale isn't the way to go, then I would rather use the "real" translation. However, just some abstract identifier might be even better. The idea behind the enum would be that we can still change the translations and their keys, without invalidating the database entries (because the further indirection through the mapping can be updated in code more easily).

app/controllers/rails_admin_controller.rb
MrSerth: The `around_action` is not strictly required, but I am wondering why it wasn't working? That shouldn't make much of a difference...

app/javascript/application.js
```diff
@@ -11,6 +11,7 @@
 import 'jquery';
 import 'jquery-ujs'
 import 'select2';
+import 'select2/dist/js/i18n/de'
```
MrSerth, line 14: Do you think it is possible and smart to use the available (and imported) I18n library below to load that dynamically?

app/mailers/access_request_mailer.rb
MrSerth: Yes, it's currently unused, but a similar mailer is required by Florian. At the moment, I would suggest keeping your update and leaving the removal / update to the contributions PR.

config/locales/de/views/tasks.yml
```diff
@@ -0,0 +1,116 @@
...
+        overwrite: Überschreiben
+        show_task: Aufgabe anzeigen
+    import_dialog_content:
+      path: 'Pfad: %{path}'
+      task_with_uuid: 'Aufgaben-UUID: %{task_uuid}'
+    index:
+      button:
+        choose_file: Datei auswählen
+        import_tasks: Aufgaben importieren
+        import_zip: ZIP-Datei importieren
```
MrSerth, line 48: In some German locales above, we used the form "Zip". I would suggest unifying that, potentially using "ZIP" (maybe check the English locales as well).

config/locales/de/views/users.yml
```diff
@@ -0,0 +1,43 @@
...
+        p1: 'Vielen Dank für Ihre Registrierung bei CodeHarbor! Klicken Sie auf den folgenden Link, um Ihren Account zu bestätigen:'
+      email_changed:
+        p1: Wir möchten Sie informieren, dass Ihre E-Mail-Adresse zu %{unconfirmed_email} geändert wurde.
+        p2: Wir möchten Sie informieren, dass Ihre E-Mail-Adresse zu %{email} geändert wird.
+      password_change:
+        p1: Wir möchten Sie darüber informieren, dass Ihr Passwort geändert wurde.
+      reset_password_instructions:
+        p1: Jemand hat einen Link zum Ändern Ihres Passworts angefordert. Klicken Sie auf den folgenden Link, um Ihr Passwort zu ändern.
+      unlock_instructions:
+        p1: Ihr CodeHarbor-Account wurde aufgrund einer zu großen Anzahl von erfolglosen Anmeldeversuchen gesperrt.
```
MrSerth, line 15: At least those lines should be a real "duplicate", aren't they? At least I remember copying them over from the German Devise locale.

app/controllers/groups_controller.rb
Mathis-Z: I think we don't need to worry about that because currently the text from the database gets overwritten in Message#text anyways. An end user never sees the text from the database. We could also leave the text column empty for group messages.

app/controllers/application_controller.rb
Mathis-Z: In this case I would remove the not-NULL constraint and use the browser locale or the default locale when the other options are all nil.

app/controllers/application_controller.rb
```diff
@@ -97,11 +97,36 @@ def mnemosyne_trace
...
+
+  def set_locale
+    locale = choose_locale
+    I18n.locale = locale
+    session[:locale] = locale
     Sentry.set_extras(locale:)
-    I18n.with_locale(locale, &)
+    if current_user.present? && locale != users_preferred_locale
```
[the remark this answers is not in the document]
Mathis-Z: Yes, I will fix that.

app/javascript/application.js
Mathis-Z: I don't know. I keep getting weird errors when trying to. I tried the following:
```
import { I18n } from "i18n-js";
import locales from "../../tmp/locales.json";

locales.forEach(locale => import(`select2/dist/js/i18n/${locale}`));
```
also
```
const available_locales = ['en', 'de'];
for(let i = 0; i < available_locales.length; i++) {
  import(`select2/dist/js/i18n/${available_locales[i]}`);
}
```
and even this:
```
const locale = 'de'
import('select2/dist/js/i18n/' + locale);
```
however this works:
```
const locale = 'de'
import('select2/dist/js/i18n/' + 'de');
```

The error I am getting is:
```
ERROR in ./node_modules/select2/dist/js/i18n/build.txt 3:14
Module parse failed: Assigning to rvalue (3:14)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
| af.js
> ----------------
| af.js
| 

webpack 5.90.2 compiled with 1 error in 408 ms
```

So I think I am misunderstanding something about dynamic imports or about how webpack handles them.

config/locales/de/views/users.yml
Mathis-Z: We cannot remove them without removing the English locales too so I just went ahead and removed both.

MrSerth: Alright, thanks!

app/controllers/groups_controller.rb
MrSerth: Alright, thanks, that makes sense. And with the `param_type`, we already have the indirection I thought of.

One final question on this topic: Wouldn't it then make sense to prevent storing the translation at all, if we never use it? Then, we would just set the `text` column to `NULL` on those occasions (in the collections and groups controller). 

By the way, in the groups controller, we can reduce the number of "redundant" message methods by passing the `param_type` as an argument. For the `Message` model, we can modify the presence validator for `text` by appending `, unless: -> { %w[group group_accepted group_declined collection].include?(param_type) }`.

app/controllers/application_controller.rb
MrSerth: Alright, then we use the automatically-derived language as long as the use hasn't made a choice themselves. Fine for me :+1:.

app/controllers/application_controller.rb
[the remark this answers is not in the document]
MrSerth: Ah, right, because sanitising a locale will also convert it to a symbol. I missed that. However, in that case, the additional call `to_sym` in `choose_locale` is unnecessary, isn't it? The first three (`sanitized...`) will always return a symbol or `nil`. The fourth option will return a value of the list we pass as argument, and the list only contains of symbols. Similarly, the last option also is a symbol.

app/javascript/application.js
MrSerth: Uh, that's a though one. I think, you didn't get that wrong. The dynamic import is, per se, working correctly as far as I can tell. However, it's too "smart" and including too many files 😕

The issue here is the select2 package and how webpack prepares the file(s) for dynamic inclusion. In short, webpack prepares each language file separately as a chunk, so that it can be loaded (if needed) by the browser. That's desired, since it prevents browsers from loading (network-wise) and interpreting (locally) the files not required. Still, some preprocessing is required for these files to work as expected.

Now, webpack encounters issues when accessing the `/node_modules/select2/dist/js/i18n/build.txt` file. That's because the file is processed as a JavaScript file (it might be, potentially, included through our `application.js`), but you'll likely agree that this doesn't make sense (if you open the file, you'll just find a list of those file names, such as `en.js`, `de.js`, ...). If you ask me, the presence of the file is a build error of the upstream repo, and potentially can be reported there as well. Nonetheless, we need to tell webpack to handle that file differently.

I tried various things, but the "best" solution I came up with so far is to modify the webpack config. To do so, place the following code snippet in `config/webpack/webpack.config.js`, lines 26+

```js
        rules: [
            erb,
            {
                test: /select2\/dist\/js\/i18n\/build\.txt$/,
                type: 'asset/resource',
                generator: {
                    emit: false,
                },
            },
        ]
```

Of course, this can also be extracted to a dedicated "loader", similar to how we handle the "erb" thing. With this rule, we will add a dedicated handler for the `build.txt` file, and prevent it from being emitted altogether. 

In summary, the aforementioned config change will allow you to use the dynamic loading as expected. Since I am not a true webpack expert, that's all I can offer (otherwise, you'd need to ask in #dev-help).

<hr>

While testing this integration, I further noticed some race condition (at least locally): Sometimes, the translation for German was not working as expected. I suspect some race condition, based on loading the file (with the translation) and firing the `turbolinks:ready` event, which will initialise the Select2 element. Since I was testing with tasks#new, I modified `app/assets/javascripts/labels.coffee.erb` to wait another second:

```coffee
# Previous code
# $(document).on 'turbolinks:load', ready

# Modified code for testing:
$(document).on 'turbolinks:load', () ->
  setTimeout(ready, 1000);
```

MrSerth:
> About the commit squashing: I would squash all of my own commits and leave yours as they are if you are okay with that.

Well, generally, a similar approach is okay. However, I would suggest not squashing too much. What do you think if we mainly squash those that contain changes to the translations and keep those separate, that include (further, relevant) code changes? Specifically, I thought of:

- "Fixing bug in task validation error display"
- "Encoding X-Message header urlsafe to support non-ascii characters in flash messages"
- "Storing preferred user locale in database" + "Removing not-NULL constraint and default value for user preferred_locale"
- either separate or together with the previous one: "Translating access request mails using recipients preferred locale" + "Translating hardcoded subject in access request mails"

Of course, this is just a suggestion and you do not need to comply with them. My suggestion is mainly based on the idea that each commit should contain something like a self-contained, small, shippable increment. For that, we can also include my first commit "Provide initial German translation" in the squashing and merge it with the rephrasing performed later. The idea to keep some other separate is that I'd argue each of them contains a dedicated "feature" or a "bug fix", or that some "explanation" in the commit might be helpful for further debugging.

app/controllers/rails_admin_controller.rb
MrSerth: I was trying the `around_action` once again, but couldn't find an issue: [around_action.patch](https://github.com/openHPI/codeharbor/files/14466846/around_action.patch)
Bildschirmfoto 2024-03-02 um 00 21 27

Maybe I am missing some action, some email that is sent or some other integration?

config/settings/development.yml
```diff
@@ -2,7 +2,7 @@ codeocean:
   url: http://localhost:7000
 omniauth:
   samltestid:
-    enable: true
+    enable: false
```
MrSerth, line 5: This change seems to be unrelated, see #1281.

app/controllers/rails_admin_controller.rb
Mathis-Z: It works for me as well now. The problem I had before was that no page on CodeHarbor would load/display anything even though the console reported that the view had been rendered. Not sure what caused that but apparently it had something to do with my setup. I am applying your patch.

app/controllers/application_controller.rb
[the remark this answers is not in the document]
Mathis-Z: That's true. I will remove that.

app/javascript/application.js
Mathis-Z: The dynamic import now works but I cannot recreate the race condition. What exactly is happening? Is the translation English even though it should be German?

config/settings/development.yml
Mathis-Z: I accidentally committed that.

Mathis-Z:
I now squashed the commits in a way that I think makes sense. What do you think?

app/javascript/application.js
MrSerth: Yes, the translation is still in English besides the remaining page being in German. There is no error in the console on the language is correctly loaded:

Bildschirmfoto 2024-03-09 um 18 03 16

It just required me five tries to get the desired behavior (German locale with German select2). For this, I was "just" reloading the page multiple times and trying over and over again:

Bildschirmfoto 2024-03-09 um 18 04 53

MrSerth (commented):
> I now squashed the commits in a way that I think makes sense. What do you think?

Looks perfectly fine; I think you found a good level of keeping dedicated units while still grouping relevant parts together.

> I was wondering whether we want to pass the locale parameter across requests like the [rails guide](https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests) suggests [...]

Mh, we could do that, of course. Nevertheless, I am actually not much in favour of always appending the locale, since it could also negatively affect the readability of URLs. Hence, and based on the previous experience on openHPI, the locale switching for users signing out should be much of a problem. In my test, openHPI seems to behave the same way, still showing the logout notification in my account language English, then rendering the remaining page in German as detected through my browser:

Bildschirmfoto 2024-03-09 um 18 28 15

Currently, CodeHarbor shows the same behavior 👍 

Bildschirmfoto 2024-03-09 um 18 30 15

Hence, I would just go with the simplicity and keep that 🙂

app/controllers/application_controller.rb
```diff
@@ -97,10 +97,36 @@ def mnemosyne_trace
...
+    sanitized_locale_param ||
+      sanitized_session_locale ||
+      sanitized_user_preferred_locale ||
+      http_accept_language.compatible_language_from(I18n.available_locales) ||
+      I18n.default_locale
+  end
+
+  def switch_locale(&)
+    locale = choose_locale
+    I18n.locale = locale
```
MrSerth, line 124: This line should be removed, since we are using the `I18n.with_locale` below.

app/models/message.rb
```diff
@@ -16,6 +16,21 @@ def mark_as_deleted(user)
     self.recipient_status = 'd' if recipient == user
   end
 
+  def text # rubocop:disable Metrics/AbcSize
+    case param_type
+      when 'group_requested'
```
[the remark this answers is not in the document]
Mathis-Z: I forgot to search for all occurrences. Sorry for that, I fixed it now.

config/locales/en/models.yml
```diff
@@ -52,6 +52,19 @@ en:
...
+        xml_id: Files XML ID
+      task/model_solutions:
+        xml_id: Model solutions XML ID
+      task/model_solutions/files:
+        attachment: Attachment of a file in a model solution
+        name: Name of a file in a model solution
+        xml_id: XML ID of a file in a model solution
+      task/tests:
+        title: Tests title
+        xml_id: Tests XML ID
```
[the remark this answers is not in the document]
Mathis-Z: I thought that was grammatically correct but apparently I was wrong. I fixed that.

db/migrate/20240312165451_rename_group_param_type.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class RenameGroupParamType < ActiveRecord::Migration[7.1]
+  def change
+    reversible do |dir|
+      dir.up do
+        Message.where(param_type: :group).find_each {|message| message.update(param_type: :group_requested) }
+      end
+
+      dir.down do
+        Message.where(param_type: :group_requested).find_each {|message| message.update(param_type: :group) }
```
MrSerth, line 11: Just thinking aloud: Do we want to be notified about potential validation errors during migration? In this case, we would need to use `update!`. Otherwise, a `Message` currently stored in the database could be invalid (e.g., because the corresponding group was deleted and the message wasn't too), and update would just return `false` without failing the migration.

MrSerth (commented):
Thanks for your final touches on this PR. The only part we need to check is the localisation of select2, otherwise I think that's fine :+1:

Mathis-Z:
Just after pushing the commit I noticed there was another bug it. I will update it soon.

Mathis-Z:
Actually I cannot figure it out. Maybe you see the problem? I pushed my current state. The problem is that only the label select2 gets initialized and the others are not.

app/assets/javascripts/tasks_index.coffee
```diff
@@ -8,18 +7,23 @@ ready = ->
...
 $(document).on('turbolinks:load', ready)
+window.select2_locales_loaded.then(initializeSelect2)
+console.log("task_index select2_locales_loaded then registered")
 
 initializeDynamicHideShow = ->
   $('body').on 'click', '.more-btn-wrapper', (event) ->
     event.preventDefault()
     toggleHideShowMore $(this)
 
 initializeSelect2 = ->
```
MrSerth, line 18: > Actually I cannot figure it out. Maybe you see the problem? I pushed my current state. The problem is that only the label select2 gets initialized and the others are not.

Here's the issue: You're first setting up the promise chain by defining that the function `initializeSelect2` should be called. However, at that time, the function `initializeSelect2` is not yet known, since it only gets defined thereafter.

app/javascript/application.js
```diff
@@ -48,6 +48,18 @@ import 'jquery-ui/themes/base/sortable.css'
...
 import locales from "../../tmp/locales.json";
 
+// create a promise that only gets resolved when turbolinks:load fired and all select2 locales have been loaded
+let select2_locales_loaded = new Promise((resolve, reject) => {
+  document.addEventListener('turbolinks:load', resolve);
+});
+Object.keys(locales).forEach(locale =>
+    select2_locales_loaded = Promise.all([select2_locales_loaded, import(`select2/dist/js/i18n/${locale}`)])
+);
+window.select2_locales_loaded = select2_locales_loaded;
```
MrSerth, line 58: Constructing multiple promises here (just by merging two promises together in each iteration) seems to be a bit complicated. Do you think we can simplify that a little?

```js
Promise.all(
  Object.keys(locales).map(locale => import(`select2/dist/js/i18n/${locale}`))
).then(() =>
  window.dispatchEvent(new Event("select2_locales_loaded"))
);
```

MrSerth, line 58: While I preferred the previous version (with emitting an event) a little bit more, the following question arised: 

Is there any specific reason you chose the `window` here (and in all other places), rather than the `document`? I would probably go with `document`, that shouldn't change much technically, just semantically:

```
document.dispatchEvent(new Event("select2_locales_loaded"))
```

And, I also like the colon `:` notation, but that's just taste.

app/assets/javascripts/tasks_index.coffee
Mathis-Z: Right, I was not aware that this is a thing in JS :)

app/javascript/application.js
Mathis-Z: That looks much better, thanks.

Mathis-Z: No, there is no specific reason I used window so I switched to document. Also I am not sure what you mean by colon notation but I am guessing you mean "select2:locales:loaded" instead of "select2_locales_loaded"?

MrSerth: Indeed, thanks!

MrSerth (commented):
Great, thank you very much for your endurance and perfection to finalise this PR! 💯 I've just rebased this branch on the current master, localised two additional strings I discovered and tackled the remaining comment. Now, I am happy to approve and merge it right away 👍

PR #1184, opened 2024-01-14
781 - collections consume invitation

kkoehn:
Currently a user can use an invitation as many times as they want, which does not feel intuitive.
This PR also adds an overview of how many open invitations a collection has

depends on #1130

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -85,8 +85,7 @@ class CollectionsController < ApplicationController
   end
 
   def share
-    message = share_message
-    if @collection.users.exclude?(share_message.recipient) && message.save
+    if @collection.users.exclude?(share_message.recipient) && share_message.save
       redirect_to collection_path(@collection), notice: t('.success_notice')
     else
       redirect_to collection_path(@collection), alert: t('common.errors.something_went_wrong')
@@ -102,7 +101,9 @@ class CollectionsController < ApplicationController
     return redirect_to user_messages_path(user), alert: t('.errors.already_member') if @collection.users.include? user
 
     @collection.users << user
-    Message.received_by(user).find_by(param_type: 'collection', param_id: @collection.id).destroy
+    message = Message.received_by(user).find_by(param_type: 'collection', param_id: @collection.id)
+    message.mark_as_deleted(user)
+    message.save
 
     redirect_to @collection, notice: t('.success_notice')
   end
@@ -128,8 +129,12 @@ class CollectionsController < ApplicationController
   end
 
   def share_message
-    user = User.find_by(email: params[:user])
-    Message.new(sender: current_user, recipient: user, param_type: 'collection', param_id: @collection.id)
+    @share_message ||= Message.new(
+      sender: current_user,
+      recipient: User.find_by(email: params[:user]),
+      param_type: 'collection',
+      param_id: @collection.id
+    )
   end
 
   def push_exercises
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -96,16 +96,16 @@ class CollectionsController < ApplicationController
     render :show
   end
 
-  def save_shared
-    user = current_user
-    return redirect_to user_messages_path(user), alert: t('.errors.already_member') if @collection.users.include? user
+  def save_shared # rubocop:disable Metrics/AbcSize
+    return redirect_to user_messages_path(current_user), alert: t('.errors.already_member') if @collection.users.include? current_user
 
-    @collection.users << user
-    message = Message.received_by(user).find_by(param_type: 'collection', param_id: @collection.id)
-    message.mark_as_deleted(user)
-    message.save
-
-    redirect_to @collection, notice: t('.success_notice')
+    ActiveRecord::Base.transaction(requires_new: true) do
+      @collection.users << current_user
+      message = Message.received_by(current_user).find_by(param_type: 'collection', param_id: @collection.id)
+      message.mark_as_deleted(current_user)
+      message.save!
+      redirect_to @collection, notice: t('.success_notice')
+    end
   end
 
   def leave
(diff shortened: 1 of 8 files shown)


app/controllers/collections_controller.rb
```diff
@@ -95,14 +97,15 @@ def view_shared
   end
 
   def save_shared
-    return redirect_to user_messages_path(current_user), alert: t('.errors.already_member') if @collection.users.include? current_user
+    user = current_user
```
Dome-GER, line 100: Why is the `current_user` assigned here? 🤔

app/controllers/collections_controller.rb
```diff
@@ -95,14 +97,15 @@ def view_shared
...
-      redirect_to @collection, notice: t('.success_notice')
-    else
-      redirect_to users_messages_path, alert: t('common.errors.something_went_wrong')
-    end
+    @collection.users << user
+    message = Message.received_by(user).find_by(param_type: 'collection', param_id: @collection.id)
+    message.mark_as_deleted(user)
+    message.save
+
+    redirect_to @collection, notice: t('.success_notice')
```
Dome-GER, line 108: So the success is defined now? Saving the message can still fail, no?

app/views/collections/show.html.slim
```diff
@@ -13,12 +13,19 @@
...
+      - if action_name != 'view_shared'
+        span
+          i.fa-solid.fa-users
+          - if @collection.users.size == 1
+            =< t('.no_other_user')
+          - else
+            =< t('.num_of_other_users', count: @collection.users.size - 1)
+        span
+          i.fa-regular.fa-envelope
+          - if @num_of_invites == 0
```
Dome-GER, line 25:
```suggestion
          - if @num_of_invites.zero?
```

I tend to prefer `zero?` (similar to `empty?`) for readability over `== 0` unless `0` is meant as some math value, i.e. coordinate or whatever.

Dome-GER (approved):
Looks good, merely minor suggestions that I consider non-blocking.

app/controllers/collections_controller.rb
kkoehn: We need `current_user` 5 times in this method and with that the Metrics/AbcSize-Cop triggers. An assignment beforehand alleviates that issue.

app/controllers/collections_controller.rb
```diff
@@ -94,14 +96,19 @@ def view_shared
...
 
-    @collection.users << current_user
-    if @collection.save
-      redirect_to @collection, notice: t('.success_notice')
+    ActiveRecord::Base.transaction(requires_new: true) do
+      @collection.users << current_user
+      message = Message.received_by(current_user).find_by(param_type: 'collection', param_id: @collection.id)
+      message.mark_as_deleted(current_user)
+      message.save!
+    rescue ActiveRecord::ActiveRecordError
```
Dome-GER, line 107: I tend to be more specific and rather list the errors I'd expect here. If something else is raised, we will notice in the monitoring.

spec/controllers/collections_controller_spec.rb
```diff
@@ -78,6 +83,19 @@
       expect(response.body).to include(I18n.t('collections.show.no_other_user'))
     end
 
+    context 'when there are pending invites' do
+      before do
+        2.times do
+          create(:message, sender: user, recipient: create(:user), param_type: 'collection', param_id: collection.id, text: 'Invitation')
```
Dome-GER, line 89: With factory_bot, `create_list` would be the built-in way to achieve this ;)

Dome-GER (approved):

app/controllers/collections_controller.rb
kkoehn: I don't really expect any errors here. I just want to make sure, that the database is in a consistent state. The user-feedback was just a bonus. As far as I see, any error is unexpected here and should end up in the monitoring. Should I just remove the `rescue` then?

Dome-GER: I'd say so.

spec/controllers/collections_controller_spec.rb
kkoehn: Is there a way to make sure that `create(:user)` is called each message individually?

Dome-GER: Sorry, I don't get the question. What's the purpose?

kkoehn: I want to create two messages, and we have a uniqueness validation for the message if it is an invitation, ensuring that each recipient can only receive one invitation per collection:
```ruby
validates :param_id, uniqueness: {scope: %i[recipient_id param_type]}, if: -> { param_type == 'collection' }
```
Therefore each message would need its own user. If I replace the current `create` with a `create_list` the `create(:user)` is evaluated first and both messages have the same user, which fails the validation.

app/controllers/collections_controller.rb
kkoehn: Added a transaction to ensure database-consistency

app/controllers/collections_controller.rb
kkoehn: done

MrSerth:
I just stumbled across the collection statistics, which I really like. Still, I have minor clarification questions: 

1. Currently, we show all four lines (visibility, member count, invitation count, favorite count) to anyone if the collection is public. I was wondering whether we should restrict some of the information (i.e., member count and/or pending invitation count) to members/admins only? What do you think, @kkoehn?

  Bildschirmfoto 2024-05-26 um 11 18 47

  ... and actually, I noticed that we hide this information for "view_shared" already:

  Bildschirmfoto 2024-05-26 um 11 25 05

2. The text "There are 2 pending invitations" does include **accepted** invitations as well, not only **pending** ones:
  
  Bildschirmfoto 2024-05-26 um 11 27 21

Can we fix these minor things, please? Sorry for the late review here 😇.

PR #1191, opened 2024-01-17
Making the collection task list sortable

### LINKED ISSUE #1125: Sort collections
Currently, collections are just a container for tasks. However, it would be useful to have all elements within a collection sorted, so that a didactical suggestion is implied with the order of tasks.

Potentially, we could use [SortableJS](https://github.com/SortableJS/Sortable) for this task (as it is also used in CodeOcean).

Mathis-Z:
I also removed the possibility to remove tasks at the collection#index and collection#show route.

Tasks can now be sorted and their order is applied on all views. When there is no explicit order set by the user then tasks will be sorted alphabetically. As soon as the user reorders the tasks this alphabetical ordering is overwritten and tasks added later will not be inserted in the task order alphabetically but instead new tasks just get appended as the last elements.
I chose to always apply the order whenever the tasks are retrieved from the database because I thought this would be convenient and the performance hit should be negligible.

Closes #1125

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -11,9 +11,9 @@ class Collection < ApplicationRecord
   has_many :users, through: :collection_users
 
   has_many :collection_tasks, lambda {
-                                includes(:task).order(rank: :desc).order('task.title ASC')
+                                order(rank: :desc).order('collection_tasks.created_at ASC')
                               }, dependent: :destroy, inverse_of: :collection
-  has_many :tasks, -> { order(rank: :desc, title: :asc) }, through: :collection_tasks
+  has_many :tasks, through: :collection_tasks
 
   accepts_nested_attributes_for :collection_tasks, allow_destroy: true
 
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -10,8 +10,12 @@ class Collection < ApplicationRecord
   has_many :collection_users, dependent: :destroy
   has_many :users, through: :collection_users
 
-  has_many :collection_tasks, dependent: :destroy
-  has_many :tasks, through: :collection_tasks
+  has_many :collection_tasks, lambda {
+                                includes(:task).order(rank: :desc).order('task.title ASC')
+                              }, dependent: :destroy, inverse_of: :collection
+  has_many :tasks, -> { order(rank: :desc, title: :asc) }, through: :collection_tasks
+
+  accepts_nested_attributes_for :collection_tasks, allow_destroy: true
 
   def add_task(task)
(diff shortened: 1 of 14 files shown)


kkoehn (changes requested):
Thanks for proposing this new feature, it makes sense to propose an order for a collection.
I have one fundamental request for implementation though: Could you try to refactor your implementation to use `nested_form_fields`? 
That way we should be able to avoid the extra `reorder_tasks` method and you wouldn't leave the edit view only because you deleted a task. 
For inspiration look at `views/tasks/_form` line 98ff and the [docs of the gem](https://github.com/ncri/nested_form_fields). 
Note that you might need to refactor your view to not use a `table`-tag, as that might conflict with magic of the gem.

Mathis-Z:
I agree that the behavior is different in that case. I wanted to sort alphabetically to automatically sort cases where a collection has tasks named like "Java 1: Hello World", "Java 2: Variables", etc.
However sorting by "added to collection date" should work too if we assume that these tasks get added in the right order.

app/models/collection.rb
```diff
@@ -10,9 +10,13 @@ class Collection < ApplicationRecord
   has_many :collection_users, dependent: :destroy
   has_many :users, through: :collection_users
 
-  has_many :collection_tasks, dependent: :destroy
+  has_many :collection_tasks, lambda {
+                                order(rank: :desc).order('collection_tasks.created_at ASC')
+                              }, dependent: :destroy, inverse_of: :collection
   has_many :tasks, through: :collection_tasks
```
Mathis-Z, line 16: I noticed that it is not necessary to sort the tasks as well.  When the tasks get queried they automatically go through the sorting of the join table. The only caveat is that you have to explicitly say which columns to sort on if there are multiple with the same name (in this case tasks and collection_tasks both have the created_at column).

app/views/collections/_form.html.slim
```diff
@@ -20,7 +20,7 @@
               = f.nested_fields_for :collection_tasks do |ct|
                 - task = ct.object.task
                 - if task.present?
-                  .row[data-id=ct.object.task_id]
+                  .row[data-id=ct.object.task.id]
```
kkoehn, line 23: Sorry for not being clear - Since you already assigned task right above, I thought it would be nice to use it.
```suggestion
                  .row[data-id=task.id]
```

kkoehn (changes requested):
Looks really good now, I only have a few small remarks left

kkoehn (approved):
I think this is great now - thanks for taking the additional time to implement the `nested_form_fields`! I really like it.

PR #1196, opened 2024-01-20
Fixing #1186: Viewing tasks is no longer possible as anonymous user

### LINKED ISSUE #1186: Viewing tasks is no longer possible as anonymous user
Today, I've noticed that viewing a public tasks without being logged in is no longer possible. This is probably a regression introduced by #1134.

Consider the task having ID 89: 
- Viewing the IEEE LOM format is possible: https://codeharbor.openhpi.de/bridges/lom/tasks/89
- Downloading the ProFormA XML is _NOT_ possible: https://codeharbor.openhpi.de/tasks/89/download
- Viewing the task in the browser is _NOT_ possible: https://codeharbor.openhpi.de/tasks/89

Mathis-Z:
Closes #1186

The problem was that the RatingPolicy and UserPolicy cannot be created without a current_user. I avoided that by introducing the extra check for current_user. Another possibility would be to set user_required? to false in the RatingPolicy and UserPolicy but I did not want to touch the policies just for this single view.

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -10,7 +10,7 @@ class TasksController < ApplicationController # rubocop:disable Metrics/ClassLen
   before_action :set_search, only: [:index]
   prepend_before_action :set_user_for_api_request, only: %i[import_uuid_check import_external]
   skip_before_action :verify_authenticity_token, only: %i[import_uuid_check import_external]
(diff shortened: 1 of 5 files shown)


Mathis-Z:
:+1: I fixed that and also updated the tests.

kkoehn (approved):

PR #1218, opened 2024-01-31
Issue #1142: Moving methods from task model to policy

### LINKED ISSUE #1142: Refactor task methods showable_by?, updateable_by?, ... with Pundit
After PR #1134 is merged the task methods showable_by?, updateable_by?, destroyable_by? and lom_showable_by? should be replaced with their pundit equivalents. The task method can_access might be replaced too.

Mathis-Z:
Closes #1142

I also moved/refactored some other methods because they were only used in the task policy
- replaced showable_by?, updateable_by?, destroyable_by?, lom_showable_by? with their pundit equivalents
- replaced can_access with a new pundit manage? 
(I am not sure if we could just replace it with update? because previously can_access implied that a user can also destroy the task. Replacing can_access with destroy? did not seem very readable though because we actually want to test if a user has full control over the task and not if he/she can destroy it. Because destroy? is the highest permission possible can_access and destroy? are equivalent but for users that dont know that this might be confusing to read. This is why I thought a manage? permission could be used in these cases where we want to test for full control over the task)
- removed/replaced author? because it was only used in the task policy
- moved and renamed in_same_group?, in_same_group_member? and in_same_group_admin? to the policy because they are only used there. However I could also see why it would make sense to have these in the model instead.

I hope you agree that these changes make sense.

### DIFF
--- a/app/views/groups/show.html.slim
+++ b/app/views/groups/show.html.slim
@@ -62,14 +62,12 @@
     .row-value
       - if @group.tasks.present?
         - @group.tasks.each do |task|
-          .table-list
+          .table-list.group-tasks
             .btn-group
+              = link_to task.title, task_path(task), class: 'btn btn-light'
               - if policy(@group).remove_task?
-                = link_to task.title, task_path(task), class: 'btn btn-light', style: "border-right: none"
-                = link_to remove_task_group_path(task: task), method: :patch, class: 'btn btn-light', style: "border-left: none; margin-right: 5px;"  do
+                = link_to remove_task_group_path(task: task), method: :patch, class: 'btn btn-light' do
                   i.fa-solid.fa-xmark style=("color:gray;")
-              - else
-                = link_to task.title, task_path(task), class: 'btn btn-light'
       - else
         = t('.has_no_tasks')
 
--- a/app/views/groups/show.html.slim
+++ b/app/views/groups/show.html.slim
@@ -64,9 +64,12 @@
         - @group.tasks.each do |task|
           .table-list
             .btn-group
-              = link_to task.title, task_path(task), class: 'btn btn-light', style: "border-right: none"
-              = link_to remove_task_group_path(task: task), method: :patch, class: 'btn btn-light', style: "border-left: none; margin-right: 5px;"  do
-                i.fa-solid.fa-xmark style=("color:gray;")
+              - if policy(@group).remove_task?
+                = link_to task.title, task_path(task), class: 'btn btn-light', style: "border-right: none"
+                = link_to remove_task_group_path(task: task), method: :patch, class: 'btn btn-light', style: "border-left: none; margin-right: 5px;"  do
+                  i.fa-solid.fa-xmark style=("color:gray;")
+              - else
+                = link_to task.title, task_path(task), class: 'btn btn-light'
       - else
         = t('.has_no_tasks')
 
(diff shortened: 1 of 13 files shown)


kkoehn (changes requested):
Thanks for refactoring and cleaning up this code.
I found two more places, where using the new policies would make sense:
- `/tasks/_tasks` line 65 shows the "Delete"-button even if the user is not allowed to `destroy`
- `groups/show` line 68 (remove task from group) does not have any check currently

Could you please also fix those problems?

Mathis-Z:
I fixed the two unauthorized buttons. I wonder if we should always hide unauthorized buttons or if it makes sense to keep them visible (maybe greyed out) to show the user that this functionality exists and that he just cannot access it at the moment.

app/views/groups/show.html.slim
```diff
@@ -64,9 +64,12 @@
...
             .btn-group
-              = link_to task.title, task_path(task), class: 'btn btn-light', style: "border-right: none"
-              = link_to remove_task_group_path(task: task), method: :patch, class: 'btn btn-light', style: "border-left: none; margin-right: 5px;"  do
-                i.fa-solid.fa-xmark style=("color:gray;")
+              - if policy(@group).remove_task?
+                = link_to task.title, task_path(task), class: 'btn btn-light', style: "border-right: none"
+                = link_to remove_task_group_path(task: task), method: :patch, class: 'btn btn-light', style: "border-left: none; margin-right: 5px;"  do
+                  i.fa-solid.fa-xmark style=("color:gray;")
+              - else
+                = link_to task.title, task_path(task), class: 'btn btn-light'
```
kkoehn, line 72: I would suggest solving that in css instead of adding the `link_to task` to both cases with minimal different styling.
For example you could add a class to the `.table-list` and do 
```SCSS
.table-list.group-tasks {
  .btn-group.btn:not(:last-child) {
    border-right: none;
  }
}
```

app/views/groups/show.html.slim
```diff
@@ -64,9 +64,12 @@
         - @group.tasks.each do |task|
           .table-list
             .btn-group
-              = link_to task.title, task_path(task), class: 'btn btn-light', style: "border-right: none"
-              = link_to remove_task_group_path(task: task), method: :patch, class: 'btn btn-light', style: "border-left: none; margin-right: 5px;"  do
-                i.fa-solid.fa-xmark style=("color:gray;")
+              - if policy(@group).remove_task?
+                = link_to task.title, task_path(task), class: 'btn btn-light', style: "border-right: none"
+                = link_to remove_task_group_path(task: task), method: :patch, class: 'btn btn-light', style: "border-left: none; margin-right: 5px;"  do
```
kkoehn, line 69: I would remove the whole custom styling for the remove-button, because I actually like the divider there:

```suggestion
                = link_to remove_task_group_path(task: task), method: :patch, class: 'btn btn-light' do
```

image

kkoehn (commented):
Looks great, but I found one more small thing in the `group/show` view.

> I wonder if we should always hide unauthorized buttons or if it makes sense to keep them visible (maybe greyed out) to show the user that this functionality exists and that he just cannot access it at the moment.

I'd say it depends, but most of the time, hiding should be fine.

app/views/groups/show.html.slim
Mathis-Z: I agree this was not the most elegant solution :+1:

kkoehn (approved):
Looks great now - thanks

PR #1223, opened 2024-02-01
781 - Collections: add favorites

kkoehn:
Add the ability to mark a Collection as favorite
Also adds corresponding filters on `collection/index`

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -118,6 +118,17 @@ class CollectionsController < ApplicationController
     end
   end
 
+  def toggle_favorite
+    if current_user.favorite_collections.include? @collection
+      current_user.favorite_collections.delete @collection
+      flash_message = t('.favorite.removed')
+    else
+      current_user.favorite_collections << @collection
+      flash_message = t('.favorite.added')
+    end
+    redirect_to @collection, notice: flash_message
+  end
+
   private
 
   def create_collection_with_task
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -4,12 +4,17 @@ require 'zip'
 
 class CollectionsController < ApplicationController
   before_action :load_and_authorize_collection, except: %i[index new create]
+  before_action :set_option, only: %i[index]
 
   def index
-    @collections = Collection.member(current_user).or(Collection.public_access).includes(:users, tasks: %i[user groups])
-      .order(id: :asc)
-      .paginate(page: params[:page], per_page: per_page_param)
-      .load
+    @collections = case @option
+                     when 'favorites'
+                       Collection.member(current_user).or(Collection.public_access).favorites(current_user)
+                     when 'public'
+                       Collection.public_access
+                     else
+                       Collection.member(current_user)
+                   end.includes(:users, tasks: %i[user groups]).order(id: :asc).paginate(page: params[:page], per_page: per_page_param)
 
     authorize @collections
   end
@@ -137,6 +142,10 @@ class CollectionsController < ApplicationController
     else
       flash.now[:alert] = t('collections.create.error')
     end
+  end
+
+  def set_option
+    @option = params[:option] || 'mine'
   end
 
   def share_message
--- a/app/policies/collection_policy.rb
+++ b/app/policies/collection_policy.rb
@@ -13,8 +13,8 @@ class CollectionPolicy < ApplicationPolicy
     everyone
   end
 
-  def show?
-    admin? || collection.users.include?(@user) || collection.visibility_level_public?
+  %i[show? toggle_favorite?].each do |action|
+    define_method(action) { admin? || collection.users.include?(@user) || collection.visibility_level_public? }
   end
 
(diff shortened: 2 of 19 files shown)


app/controllers/collections_controller.rb
```diff
@@ -118,6 +123,17 @@ def leave
     end
   end
 
+  def toggle_favorite
```
Dome-GER, line 126: I'm more a fan of individual controllers following RESTful resource style, e.g. `POST /favorites` here.
But you have if different style in this project anyway.

app/policies/collection_policy.rb
```diff
@@ -13,8 +13,8 @@ def new?
     everyone
   end
 
-  def show?
-    admin? || collection.users.include?(@user) || collection.visibility_level_public?
+  %i[show? toggle_favorite?].each do |action|
```
Dome-GER, line 16: I also prefer duplication to meta programming for simplicity, but that's also rather a personal aspect.

app/views/collections/_collections.html.slim
```diff
@@ -13,6 +13,8 @@
   .card.my-4
     .card-header
       h6
+        - if current_user.favorite_collections.include?(collection)
+          i.fa-solid.fa-star>
```
Dome-GER, line 17: What is the trailing `>` good for?

app/views/collections/show.html.slim
```diff
@@ -26,6 +28,13 @@
             =< t('.no_invites')
           - else
             =< t('.num_of_invites', count: @num_of_invites)
+        span
+          - if @collection.user_favorites.length == 0
```
Dome-GER, line 32:
```suggestion
          - if @collection.user_favorites.length.zero?
```

Nice syntactic sugar and supporting understanding (comparison of numbers vs. "presence" check). Same as mentioned in the other MR before. ;)

More high-level question: Why making a difference here? You are using `count`, so the locale for zero favorites can easily cover the `.no_favorites` locale while the `fa-*` class can be set dynamically based on the condition.

db/migrate/20240410131313_add_collection_user_favorites.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddCollectionUserFavorites < ActiveRecord::Migration[7.1]
+  def change
+    create_table :collection_user_favorites, id: :uuid, force: :cascade do |t|
```
Dome-GER, line 5: Very good choice to make sure we have proper database constraints in place. 👏

app/views/collections/_collections.html.slim
kkoehn: It adds a trailing whitespace. In this case:
without `>`
image
with '>'
image

app/controllers/collections_controller.rb
kkoehn: I see what you mean, but would still keep it, since it matches the current style as you said.

app/policies/collection_policy.rb
kkoehn: Normally I agree and tend to avoid meta programming, but in case of these policies I like that you can easily see which actions belong together. Especially if you have a lot of actions (like in line 20-21 here)

app/views/collections/show.html.slim
kkoehn: You are completetly right. I also refactored the "invitations" and "other users" to use `count` correctly

app/views/collections/_collections.html.slim
Dome-GER: I see. I know what it does for expressions, e.g. `=>` but have not seen it used for classes.

app/controllers/collections_controller.rb
```diff
@@ -4,12 +4,17 @@
...
   before_action :load_and_authorize_collection, except: %i[index new create]
+  before_action :set_option, only: %i[index]
 
   def index
-    @collections = Collection.member(current_user).or(Collection.public_access).includes(:users, tasks: %i[user groups])
-      .order(id: :asc)
-      .paginate(page: params[:page], per_page: per_page_param)
-      .load
+    @collections = case @option
+                     when 'favorites'
```
Dome-GER, line 11: By the way, this could also be constants, e.g. `Collection::OPTION::FAVORITES`, like you did previously (don't remember where I saw this).

Dome-GER (approved):

app/controllers/collections_controller.rb
kkoehn: I see but you mean, but don't remember using it anywhere and couldn't find it - I would like to keep it as it is for now

MrSerth:
Just some feedback on the German locale: 

- The text chosen for the ribbon seems to be too long, causing the "New Collection" button to wrap in a new line.
- When I don't have a favorite collection (or any other public collection) yet, the error message is misleading. For both, another explanation would be more suitable.

Bildschirmfoto 2024-05-26 um 11 05 00
Bildschirmfoto 2024-05-26 um 11 07 26

@kkoehn Would you mind having another look here again, please?

PR #1248, opened 2024-02-18
Disallowing group admins to destroy tasks they do not own

### LINKED ISSUE #1213: Broken destroy policy for tasks
Currently you can edit tasks when you share a group with them but you can only destroy them if you are also the admin of that group.
However this can be circumvented by creating a second group which you are now admin of and then you can add the task to your group. This gives you the right to destroy the task which also removes it from the first group even though you should not be allowed to do that.

I think the easiest thing to do here would be to disallow group members to add group tasks to other groups. In the task form view we could add a tooltip saying something like: "You cannot add this task to another group. Duplicate it instead"

Mathis-Z:
Closes #1213

I thought about whether we would need a button like "Remove from my groups" for group admins that removes the task from all groups the current user is an admin in or if we wanted the destroy button to behave like that instead of actually destroying the task. I decided against that because I think a "Remove from my groups" button might be a bit confusing given that "my groups" in this case only means groups the current user is admin in and also includes groups the current user did not originally create. Also this button would not add functionality because group admins can already remove tasks from their groups.
The second option of changing the behavior of the destroy button might also be a bit confusing because the button would not actually do what it says.

### DIFF
--- a/app/policies/task_policy.rb
+++ b/app/policies/task_policy.rb
@@ -46,9 +46,7 @@ class TaskPolicy < ApplicationPolicy
   end
 
   def destroy?
-    return false if @user.blank?
-
-    record_owner? || task_in_group_with_admin?(@user) || admin?
+    record_owner? || admin?
   end
 
   def manage?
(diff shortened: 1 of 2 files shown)


kkoehn (approved):
Small and concise - very good, thanks!

PR #1250, opened 2024-02-19
allow filenames of imported files to contain spaces and symbols

### LINKED ISSUE #1092: Allow ProFormA ZIP files with spaces (and other special characters)
During my testing of the interoperability with different ProFormA versions and file providers, I was (accidentally) using a file name with a space. That caused an exception in CodeHarbor, since the desired file could not be processed (after a successful, initial validation). 

The error is caused by this line:

https://github.com/openHPI/codeharbor/blob/acf10d4a347082a3a25c1816fd75812b6364dff8/app/services/proforma_service/proforma_task_from_cached_file.rb#L16

As seen in the screenshot, a "normalised" path with underscores is compared to my user input containing spaces. Therefore, no file is found (result of the `.find{}` is `nil`, so that the `[:task]` access is returning an error.

Bildschirmfoto 2023-11-17 um 11 15 32

We should fix this error and also investigate whether other special characters in file names are causing issues.

Here is my example: [task 2.0 ohne p.zip](https://github.com/openHPI/codeharbor/files/13390460/task.2.0.ohne.p.zip)

kkoehn:
fixes #1092 

This removes the normalization of filenames introduced in https://github.com/openHPI/codeharbor/pull/288/commits/9a40f3f6839dd1897fd92025f37016f54281bb98. I did not find a reference to a specific problem I tried to fix back then.

Symbols were also problematic, because at one point [ActiveStorage::Blob#filename](https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-filename) was called, which implicitly returns a sanitized string. To avoid that I concatenated `base` and `extension` by hand.


app/services/proforma_service/cache_import_file.rb
```diff
@@ -25,7 +25,7 @@ def execute
     private
 
     def file_data_hash(task, import_file, proforma_task)
-      {path: proforma_task[:path].tr(' ', '_'),
+      {path: proforma_task[:path],
```
Dome-GER, line 28: There is also `Filenam#sanitized`, which may be more useful. But hard to tell without domain knowledge.

Dome-GER (approved):
Good to see that this is covered by specs properly. Just a small idea to look into.

PR #1261, opened 2024-02-21
Include unittest generator using openAI gem

Melhaya:
- the aim is to use the openAI GPT model to generate unittests given a task description.
- An access key is required in order to run the generator successfully
- If the key doesn't exist, the button doesn't appear.
- The aim is to make it more dynamic in the future

This is in draft until the new features are added.


.gitignore
Melhaya, line 1: remove this line

app/services/task_service/gpt_generate_tests.rb
Melhaya, line 1: further clean up of file needed

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -0,0 +1,63 @@
...
+    def initialize(task:)
+      super()
+      @task = task
+      @message = task.description
+      @language = @task.programming_language.language || 'Python'
+    end
+
+    def execute
+      file_content = gpt_response
+      test_file = TaskFile.new(content: file_content, name: 'test.py', used_by_grader: true, visible: false, xml_id: SecureRandom.uuid)
```
MrSerth, line 14: The filename (or, to be more specific, the file extension) might not suit non-Python programming languages. Consider a Java, R, or Ruby program that ends up with a `.py` file.

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -0,0 +1,63 @@
...
+      testing_framework = TestingFramework.find_or_initialize_by(name: 'Pytest', version: 6)
```
MrSerth, line 15: Similar to above, the test framework might not be suitable.

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -0,0 +1,63 @@
...
+      test = Test.new(task: @task, title: 'AI-generated Test', xml_id: SecureRandom.uuid, files: [test_file], testing_framework:)
```
MrSerth, line 16: This ('AI-generated Test') seems to be a translatable string, should we localise it?

config/settings/development.yml
```diff
@@ -10,3 +10,5 @@ omniauth:
     enable: false
 oai_pmh:
   admin_mail: admin@example.org
+open_ai:
+  access_token: ~ #ENTER ACCESS TOKEN HERE
```
MrSerth, line 14: Usually, we place at least a space between the comment and the hashtag. Maybe, we could also add a reference to https://platform.openai.com/api-keys, since users can generate an API key there.

This comment applies to all files in this directory.

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -0,0 +1,63 @@
...
+      @task.tests << test
+      @task.save!
```
MrSerth, line 18: The first line (`@task.tests << test`) will probably save the task already; do we need the second one?

app/services/task_service/gpt_generate_tests.rb
MrSerth, line 1: Can we identify whether ChatGPT refused to generate a test? In my example, I got the following "test" file:

Bildschirmfoto 2024-04-17 um 21 19 33

MrSerth, line 1: Here's another response, that is more likely expected to be rendered in a markdown context (which is unexpected):

Bildschirmfoto 2024-04-17 um 21 20 04
Bildschirmfoto 2024-04-17 um 21 20 51

MrSerth (commented):
As mentioned today, I made a first review. During my review, I focussed on the user-facing functionality and did not review the specs (leaving those to @kkoehn).

spec/services/task_service/gpt_generate_tests_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    it 'creates a test file and test instance related to the task' do
+      expect { gpt_generate_tests_service.execute }.to change(TaskFile, :count).by(1).and change(Test, :count).by(1)
+
+      test_file = TaskFile.last
+      test = Test.last
+
+      expect(test_file.content).to eq('def test_script(): assert True')
+      expect(test_file.name).to eq('test.py')
+      expect(test.title).to eq('AI-generated Test')
+      expect(task.tests).to include(test)
```
kkoehn, line 45: I would sum those expectations up in one by using the `have_attributes`-matcher (Used many times in this project. One example: `spec/services/proforma_service/convert_proforma_task_to_task_spec.rb:318`)

spec/services/task_service/gpt_generate_tests_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+  end
+
+  describe '#gpt_response' do
```
kkoehn, line 49: Since this method should be private (see other comment). This example could be part of the perform example.

app/services/task_service/gpt_generate_tests.rb
Melhaya: I created a mapping for the language type to match it with the file extension. Any other languages I should consider?

```
      file_extensions = {'Python' => 'py', 'JavaScript' => 'js', 'Ruby' => 'rb', 'Java' => 'java'}
      testing_frameworks = {'Python' => 'Pytest', 'JavaScript' => 'Jest', 'Ruby' => 'RSpec', 'Java' => 'JUnit'}

      framework_name = testing_frameworks[@language]
      framework_version = {'Pytest' => 6, 'Jest' => 27, 'RSpec' => 4, 'JUnit' => 5}
      file_extension = file_extensions[@language]
```

app/services/task_service/gpt_generate_tests.rb
Melhaya: see above comment :)

app/services/task_service/gpt_generate_tests.rb
Melhaya: I created a localized message for it.

en
```
tests:
    generated_test: AI-generated Unit Test
```

de
```
tests:
   generated_test: 'AI-generierter Unit-Test'
```

app/services/task_service/gpt_generate_tests.rb
Melhaya: From a quick search, it seems that both commands perform slightly different things therefore both are required. at least that's what I understood.

app/services/task_service/gpt_generate_tests.rb
Melhaya: could you provide me with the task description you used?

spec/services/task_service/gpt_generate_tests_spec.rb
Melhaya: Hey :) I am not sure what you mean here. could you explain more? I have changed the method above to private as you mentioned.

spec/services/task_service/gpt_generate_tests_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    it 'assigns language from programming language' do
+      expect(gpt_generate_tests_service.instance_variable_get(:@language)).to eq('Ruby')
+    end
+  end
+
+  describe '#execute' do
+    let(:programming_language) { create(:programming_language, language: 'Python') }
+    let(:task) { create(:task, description: 'Create a Python script.', programming_language:) }
+    let(:gpt_generate_tests_service) { described_class.new(task:) }
```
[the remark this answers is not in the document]
Melhaya: Also here I am not sure what I need to do 😬
should I add a similar line to this `subject(:convert_zip_to_tasks) { described_class.call(zip_file:) }` here?

app/services/task_service/gpt_generate_tests.rb
MrSerth: Sounds good. Please store this information in the database, so that we can adjust it there.

app/services/task_service/gpt_generate_tests.rb
MrSerth: Thanks, I would go with a database entry, too.

app/services/task_service/gpt_generate_tests.rb
MrSerth: Let's use "KI-generierter Unit-Test" in German :+1:

app/services/task_service/gpt_generate_tests.rb
MrSerth: Sure, usually they do, but not necessarily in this example. However, it is unexpected that line 17 already saves the task as well, this is what I meant. Feel free to use a debugger to observe the behavior and place a breakpoint before executing line 18.

app/services/task_service/gpt_generate_tests.rb
MrSerth: I think, I was using this one:

```md
* `Download` and save the zip folder where you'd like it to be (no need to unzip).
* In R/Rstudio, run `codeoceanR::rt_create()`, see lesson 1.4 and [instructions][1].



[1]: https://github.com/openHPI/codeoceanR#procedure-for-each-exercise
```

For example, it is used [here](https://codeharbor.openhpi.de/tasks/758).

Melhaya: Yeah, this is not a nice example to provide to the model. you are referencing a lesson it has no access to and a link it can't understand. 
This could be a type of example the model can't comprehend and therefore would refuse to create a unit test for.

Melhaya: I have extracted the text from the markdown context and it should appear as desired now. :)

app/services/task_service/gpt_generate_tests.rb
Melhaya: Not sure what that means. I have just pushed the new changes and adjustments. Could you please clarify more about storing in the database and how to do it? 
Everything else should be tackled, covered, and tested locally. Waiting for you to check them and resolve conflicts as discussed :)

app/services/task_service/gpt_generate_tests.rb
MrSerth: Sure, I fully agree. Still, I was wondering whether there is some other way to handle this response rather then creating a "fake" file with this file content (i.e., showing some flash message etc).

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -6,18 +6,38 @@ def initialize(task:)
...
+      file_extension = file_extensions[@language]
+
       file_content = gpt_response
-      test_file = TaskFile.new(content: file_content, name: 'test.py', used_by_grader: true, visible: false, xml_id: SecureRandom.uuid)
-      testing_framework = TestingFramework.find_or_initialize_by(name: 'Pytest', version: 6)
-      test = Test.new(task: @task, title: 'AI-generated Test', xml_id: SecureRandom.uuid, files: [test_file], testing_framework:)
+      test_file = TaskFile.new(content: file_content, name: "test.#{file_extension}", used_by_grader: true, visible: false,
+        xml_id: SecureRandom.uuid)
+      testing_framework = TestingFramework.find_or_initialize_by(name: framework_name, version: framework_version[framework_name])
+      test = Test.new(task: @task, title: I18n.t('common.tests.generated_test'), xml_id: SecureRandom.uuid, files: [test_file],
```
MrSerth, line 28: Probably, the translation currently located at `'common.tests.generated_test'` is better located at `'tests.model.generated_test'` (and `model.yml`), similarly to `tasks.model.copy_of_task`?

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -37,13 +57,14 @@ def gpt_response
...
+      raw_response = response.dig('choices', 0, 'message', 'content')
 
-    def client
-      @client ||= OpenAI::Client.new
+      # check for ``` in the response and extract the text between the first set
+      if raw_response.include?('```')
+        raw_response[/```(.*?)```/m, 1].lines[1..]&.join&.strip
+      else
+        raw_response
+      end
```
MrSerth, line 67: I am eager to hear Karol's opinion here, but one could consider extracting this part to a dedicated method. Nothing required from my side, just optional.

app/services/task_service/gpt_generate_tests.rb
MrSerth: Thanks, I resolved some already :+1:

Regarding the database persistency: You'll likely need to create a new model (i.e., `FileExtension`) or extend the current one `ProgrammingLanguage`. Both have advantages and disadvantages, I would say. For the testing framework, a new column is probably enough (a reference to the programming language). 

To persist data, we are using [`ActiveRecord`](https://guides.rubyonrails.org/active_record_basics.html) as a layer to integrate with the database. Adding a new column (or table) is done with an [Active Record Migration](https://guides.rubyonrails.org/active_record_migrations.html). For the reference (testing framework to the programming language), you'll need a `reference` (see section 3.6 in the guide). Please add a foreign key and maybe an index, too. For the other problem, the migration guide will also point out how to create a generic column (data type would be `:string`) or a table.

So far, this is just as a first reference to get started...

app/services/task_service/gpt_generate_tests.rb
Melhaya: I will handle this by creating a notice to the user instead of creating a file. that's a good idea :)

Melhaya: Not sure how to test it though. When I write this:


```
    context 'when task description is invalid' do
      let(:task) { build(:task, description: '', programming_language:) }

      it 'raises InvalidTaskDescription' do
        expect { gpt_generate_tests_service }.to raise_error(InvalidTaskDescription)
      end
    end
```

Nothing is raised. Not sure how to cause this error to be activated. Any help?

app/services/task_service/gpt_generate_tests.rb
Melhaya: Yes, I outcommented it and everything still worked fine. Will remove this line. Thanks

app/services/task_service/gpt_generate_tests.rb
Melhaya: I inserted the test in the #gpt_response section and now I can detect the raised error.

Melhaya: I was able to handle it by adjusting the test case in '#gpt_response' as follows:

```
    it 'raises InvalidTaskDescription when response does not contain backticks' do
      allow(mock_client).to receive(:chat).and_return({'choices' => [{'message' => {'content' => 'Sample response'}}]})
      expect { gpt_generate_tests_service.send(:gpt_response) }.to raise_error(InvalidTaskDescription)
    end
```

app/services/task_service/gpt_generate_tests.rb
Melhaya: Might need help doing this.. Will try to do it till then.

app/services/task_service/gpt_generate_tests.rb
Melhaya: I have moved the translation to the dedicated files as suggested. I did add an extra translation for 'invalid task description' in `tasks.yml` so I will not resolve this until that is also checked and confirmed to be correct. Thanks.

app/services/task_service/gpt_generate_tests.rb
kkoehn: I don't see the need to extract that part, but would refactor the if-else into a guard-clause. (since the else-branch now contains a raise)

spec/services/task_service/gpt_generate_tests_spec.rb
kkoehn: I wouldn't test this method directly, but through the normal execution path (`Service.call`) (see other comment)

spec/services/task_service/gpt_generate_tests_spec.rb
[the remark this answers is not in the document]
kkoehn: yes exactly. The method you want to test should be the `subject`. In this case it could look like this:

```ruby
describe '.call' do
    subject(:gpt_generate_tests) { described_class.call(task:) }
    let(:programming_language) { create(:programming_language, language: 'Python') }
    let(:task) { create(:task, description: 'Create a Python script.', programming_language:) }
```
an expectation could look like this
```ruby
expect { gpt_generate_tests }.to ....
```

spec/services/task_service/gpt_generate_tests_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      expect(test_file).to have_attributes(
+        content: "def test_script():\n  assert true",
+        name: 'test.py'
+      )
+    end
+
+    it 'creates a test instance related to the task' do
+      expect { gpt_generate_tests_service.execute }.to change(Test, :count).by(1)
+      test = Test.last
+      expect(test).to have_attributes(title: I18n.t('tests.model.generated_test'))
```
kkoehn, line 56: You could directly access the tests of the task like this:
`task.reload.tests.last` the `reload` reloads the object from the database so you can see your changes.
With this you could also omit the `include` check.

This can also be applied to the example above.

app/services/task_service/gpt_generate_tests.rb
MrSerth: The error handling works as expected, thanks! The only remaining issue I have is the color of the alert you show, it's green usually indicating a success. [See alert examples here](https://getbootstrap.com/docs/4.0/components/alerts/#examples), that we map as follows:

https://github.com/openHPI/codeharbor/blob/b71c67b683bf77adca9d5c649128b40ce40817e5/app/views/application/_flash.html.slim#L6

Regarding your test: It works fine, and as expected. However, while calling a private method with `send` works, it is usually discouraged (even in tests), since you're ignoring the responsibilities configured through the visibility level.

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -0,0 +1,83 @@
...
+  class GptGenerateTests < ServiceBase
+    def initialize(task:)
+      super()
+      @task = task
+      @message = task.description
+      if @task.programming_language&.language
+        @language = @task.programming_language.language
+      else
+        raise MissingLanguageError
+      end
```
MrSerth, line 13: We can probably simplify this:

```suggestion
      @language = @task.programming_language&.language
      raise MissingLanguageError if @language.blank?
```

app/services/task_service/gpt_generate_tests.rb
Melhaya: What do you suggest I write instead?

app/services/task_service/gpt_generate_tests.rb
Melhaya: I will switch to alert instead of notice. thanks :) 

I am not sure how else to call it then?

MrSerth: As mentioned by @kkoehn in https://github.com/openHPI/codeharbor/pull/1261#discussion_r1574983426, you should rather call a public method in the service and still expect the same exception.

app/services/task_service/gpt_generate_tests.rb
MrSerth: Ah, don't mind. You are not changing the task at all, so that this method is fine.

app/services/task_service/gpt_generate_tests.rb
Melhaya: I also switched the notice here to alert.

spec/services/task_service/gpt_generate_tests_spec.rb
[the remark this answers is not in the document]
Melhaya: I am not sure if this is what you meant.

I have added this function as public:

```
    def process_gpt_response
      gpt_response
    end
```

and adjusted the tests as follows as you suggested:


```
  describe '#gpt_response' do
    subject(:gpt_generate_tests_service) { described_class.new(task:) }

    let(:programming_language) { build(:programming_language, language: 'Ruby') }
    let(:task) { build(:task, description: 'Sample Task', programming_language:) }
    let(:mock_client) { instance_double(OpenAI::Client) }

    before do
      allow(OpenAI::Client).to receive(:new).and_return(mock_client)
    end

    it 'raises InvalidTaskDescription when response does not contain backticks' do
      allow(mock_client).to receive(:chat).and_return({'choices' => [{'message' => {'content' => 'Sample response'}}]})
      expect { gpt_generate_tests_service.process_gpt_response }.to raise_error(InvalidTaskDescription)
    end

    it 'extracts code from response when enclosed in backticks' do
      allow(mock_client).to receive(:chat).and_return({'choices' => [{'message' => {'content' => "```ruby\ndef test_sample\n  assert true\nend\n```"}}]})
      expect(gpt_generate_tests_service.process_gpt_response).to eq("def test_sample\n  assert true\nend")
    end
  end
```

Is this what you were referring to? :)

spec/services/task_service/gpt_generate_tests_spec.rb
Melhaya: for `test_file = TaskFile.last` what should be the alternative in this case?

app/services/task_service/gpt_generate_tests.rb
Melhaya: I have handled this and provided a comment here: https://github.com/openHPI/codeharbor/pull/1261#discussion_r1618707934

Is this what you both meant?

spec/services/task_service/gpt_generate_tests_spec.rb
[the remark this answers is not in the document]
MrSerth: We discussed that in person today.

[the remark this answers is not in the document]
Melhaya: I had a meeting with Sebastian and this should be fixed now. please take a look whenever possible to resolve the comment. thanks a lot for your feedback and help 😊

spec/services/task_service/gpt_generate_tests_spec.rb
Melhaya: This is the alternative used: `test_file = task.reload.tests.last.files.first`

Melhaya:
Dear @MrSerth and @kkoehn, Thank you both for your feedback and help so far in merging this PR. 
Currently, all of your issues and comments should be resolved except for storing certain information in a database as mentioned here: https://github.com/openHPI/codeharbor/pull/1261#discussion_r1609016956

If you agree, then please resolve the related comments and hopefully I can add the store the relevant data in the database and we can finally merge this feature into CodeHarbor. 🤞🏼😊

app/services/task_service/gpt_generate_tests.rb
MrSerth: Yes, looks good. Thanks!

app/services/task_service/gpt_generate_tests.rb
Melhaya: Hey @MrSerth, I am working on the ActiveRecords part of the PR (With the help and expertise of AI :robot_face: ) 

When I extended the `ProgrammingLanguage` module to include file_extension, testing_framework, and framework_version, manyyy tests failed because they didn't initialize these new parameters. So I was wondering if it makes sense to add a condition to make these new variables only necessary in the GPT context. otherwise, we will have to change all task initialization in the repo to include the new variables which I don't think is what we want to do :smile: 

here is the code:

```
# frozen_string_literal: true

class ProgrammingLanguage < ApplicationRecord
  attr_accessor :validation_context

  has_many :tasks, dependent: :destroy

  validates :file_extension, presence: true, if: :gpt_context?
  validates :testing_framework, presence: true, if: :gpt_context?
  validates :framework_version, presence: true, if: :gpt_context?

  def language_with_version
    "#{language} #{version}"
  end

  private

  def gpt_context?
    validation_context == :gpt
  end
end

```
When I did this, tests worked.

Is this in any way in the direction you were asking about?

MrSerth: Thanks for working on this topic further. I still have a few concerns, i.e., on reusing the existing `TestingFramework` model, the `attr_accessor :validation_context` added, or how these values are permanently stored. To resolve these uncertainties, a pair-programming session would probably be helpful. 

Unfortunately, I cannot offer this service at the moment.

Melhaya: It would be very helpful to have this session together. Feel free to set up a suitable time for both of us to resolve your concerns :)

MrSerth: Sure, I fully get that. However, I really cannot make it and am sorry for that 😕. I fear, you'll need to find another solution here 🍀🤞.

spec/services/task_service/gpt_generate_tests_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    context 'when the response does not contain backticks' do
+      before do
+        allow(mock_client).to receive(:chat).and_return({'choices' => [{'message' => {'content' => 'Python script should assert true without any code block.'}}]})
+      end
+
+      it 'raises InvalidTaskDescription when response does not contain backticks' do
+        expect { gpt_generate_tests_service.execute }.to raise_error(InvalidTaskDescription)
+      end
+    end
+  end
```
kkoehn, line 80: This is want I meant:

```suggestion
  describe '#call' do
    subject(:gpt_generate_tests) { described_class.call(task:) }

    let(:programming_language) { create(:programming_language, language: 'Python', file_extension: 'py', testing_framework: 'Pytest', framework_version: '6') }
    let(:task) { create(:task, description: 'Create a Python script.', programming_language:) }
    let(:mock_client) { instance_double(OpenAI::Client) }

    before do
      allow(OpenAI::Client).to receive(:new).and_return(mock_client)
      programming_language.validation_context = :gpt
    end

    context 'when the response includes valid code blocks' do
      before do
        allow(mock_client).to receive(:chat).and_return('choices' => [{'message' => {'content' => "```Python\ndef test_script():\n  assert true```"}}])
        gpt_generate_tests
      end

      it 'creates a test file related to the task' do
        test_file = task.reload.tests.last.files.first
        expect(test_file).to have_attributes(
          content: "def test_script():\n  assert true",
          name: 'test.py'
        )
      end

      it 'creates a test instance related to the task' do
        test = task.reload.tests.last
        expect(test).to have_attributes(title: I18n.t('tests.model.generated_test'))
      end

      it 'associates the correct testing framework with the test' do
        test = task.reload.tests.last
        expect(test.testing_framework).to have_attributes(name: 'Pytest', version: '6')
      end
    end

    context 'when the response does not contain backticks' do
      before do
        allow(mock_client).to receive(:chat).and_return({'choices' => [{'message' => {'content' => 'Python script should assert true without any code block.'}}]})
      end

      it 'raises InvalidTaskDescription when response does not contain backticks' do
        expect { gpt_generate_tests }.to raise_error(InvalidTaskDescription)
      end
    end
  end
```

spec/services/task_service/gpt_generate_tests_spec.rb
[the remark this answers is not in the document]
kkoehn: A lot of steps in the right direction, I added a new suggestions for the last steps I had in mind.

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -0,0 +1,78 @@
...
+    def execute
+      programming_language = @task.programming_language
+      programming_language.validation_context = :gpt
+
+      file_extension = programming_language.file_extension
+      testing_framework_name = programming_language.testing_framework
+      framework_version = programming_language.framework_version
+
+      file_content = gpt_response
+      test_file = TaskFile.new(content: file_content, name: "test.#{file_extension}", used_by_grader: true, visible: false,
```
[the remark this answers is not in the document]
kkoehn: Wouldn't it be possible to outsource all those decisions to the AI as well? I could imagine modifying the prompt so that it also returns a fitting filename (with extension).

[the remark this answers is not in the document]
Melhaya: I discussed this with Karol and there are many challenges to automating the correct filename creation. I think this should be resolved for the current PR and this challenge can be tackled in a future PR. 

For now, instructors can rename the test file accordingly.

app/services/task_service/gpt_generate_tests.rb
MrSerth: I am resolving this thread. Now, there is a file extension in the `ProgrammingLanguage` mode, and since the `TestingFramework` is optional, we don't need it.

app/services/task_service/gpt_generate_tests.rb
MrSerth: The testing framework is no longer set at all, and since it is optional, we can resolve this thread.

app/services/task_service/gpt_generate_tests.rb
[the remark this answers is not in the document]
MrSerth: Okay, we can proceed with this approach. In this case, wouldn't it be suitable to suggest users to double check (and potentially correct) the file name chosen (or the test in general)? This could be integrated into the success message, I think.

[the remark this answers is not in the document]
Melhaya: Yes, I can add that to the success message.

One extra future work thing I don't think should be too hard to add is the possibility for users to input their own API key for the feature to appear. not sure how I can proceed with doing this? Because currently, the feature is hidden for everyone without a key.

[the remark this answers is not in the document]
Melhaya: Does this make sense:

successful_generation: Unit test generated successfully. Please check the generated test and provide the appropriate filename to it

successful_generation: Unit-Test erfolgreich generiert. Bitte prüfen Sie den generierten Test und geben Sie ihm den entsprechenden Dateinamen

db/schema.rb
```diff
@@ -325,7 +326,7 @@
     t.datetime "locked_at"
     t.string "preferred_locale"
     t.boolean "password_set", default: true, null: false
-    t.integer "status_group", limit: 1, default: 0, null: false, comment: "Used as enum in Rails"
+    t.integer "status_group", limit: 2, default: 0, null: false, comment: "Used as enum in Rails"
```
MrSerth, line 329: This change is unexpected here 🤔

spec/controllers/tasks_controller_spec.rb
```diff
@@ -1041,4 +1041,76 @@
...
+
+    context 'when GptGenerateTests is successful' do
+      before do
+        allow(TaskService::GptGenerateTests).to receive(:call)
+        post :generate_test, params: {id: task.id}
+      end
+
+      after do
+        Settings.open_ai.access_token = nil
+      end
```
MrSerth, line 1063: Rather than repeating this `after` block multiple times, wouldn't it work to move it below setting the access token and remove the duplications?

```rb
before do
   sign_in task_user
   Settings.open_ai.access_token = 'access_token'
 end
```

MrSerth (commented):
Thank you so much for continuously working on this PR! All my major concerns have been resolved 🥳 There are a few comments left, but I am confident we are getting close to merging this PR soon!

app/services/task_service/gpt_generate_tests.rb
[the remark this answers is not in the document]
MrSerth: > One extra future work thing I don't think should be too hard to add is the possibility for users to input their own API key for the feature to appear. not sure how I can proceed with doing this? Because currently, the feature is hidden for everyone without a key.

Yes, that's a good future work. For now, and for this PR, I would suggest to proceed with the global key. Then, we can ship small releases and make a follow-up PR.

Adding the API key per user would probably involve another database column per user and a simple edit view as part of the profile. Then, we need to adapt the corresponding policy and use this user-specific key.

<hr>
Sounds good! For the German translation, I got a minor suggestion. Based on that, I updated the English locale as well, but this is optional (we can keep your translation there, too).

```yml
successful_generation: Unit test generated successfully. Please check the generated test and assign an appropriate filename

successful_generation: Unit-Test erfolgreich generiert. Bitte überprüfen Sie den generierten Test und vergeben Sie einen passenden Dateinamen
```

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -0,0 +1,73 @@
...
+          messages:,
+          temperature: 0.7, # Lower values insure reproducibility
+        }
+      )
+
+      # parse out the response
+      raw_response = response.dig('choices', 0, 'message', 'content')
+
+      # check for ``` in the response and extract the text between the first set
+      raise InvalidTaskDescription unless raw_response.include?('```')
```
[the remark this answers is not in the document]
Melhaya: with the current prompt, if the model responds correctly, then it will include the code markdown which we can then extract. However, if the task description is unclear, it will only include text. I have tried this locally a few times and this seemed to be the best way to get a consistent output from the model and detect when the description was not sufficient to yield a code output.

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+module TaskService
+  class GptGenerateTests < ServiceBase
+    def initialize(task:)
+      super()
+      @task = task
+      @message = task.description
+      @language = @task.programming_language&.language
```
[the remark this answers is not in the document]
Melhaya: I went with the more likely option.

app/views/tasks/show.html.slim
```diff
@@ -531,6 +531,13 @@
           div data-bs-toggle='tooltip' title=t('.guest.disabled.tooltip') data-bs-delay=150
             = button_tag class: 'btn btn-outline-dark dropdown-toggle disabled', 'data-bs-toggle': 'dropdown' do
               = t('.button.export')
+  - if policy(@task).generate_test?
```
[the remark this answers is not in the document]
Melhaya: I can't see the image you attached. Could you please provide the code change you did of which you created this box?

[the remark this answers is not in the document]
Melhaya: I could also remove that box completely and only include the one button at the bottom :)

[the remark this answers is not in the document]
MrSerth: > I can't see the image you attached. 

The two images are attached in the collapsable "details" part, which is hidden by default. Clicking on the line with the black triangle reveals it.

<details>
<summary>Code Mockup</summary>

I was editing in the DOM viewer, thus I only hav the HTML. This is not necessary nice, just wanted to get the visual result (and I could have used a drawing tool, too).

```html
<div class="m-0 mb-3 p-0" style="
    height: 115px;
"><div class="border button-box h-100 m-0 mb-4 p-3" style="
    background-color: antiquewhite;
    height: 40px;
"><div class="button-description" style="
">Beta Feature: Generate unit tests for this Task.<br>As part of our research ...
</div><div class="wrapper"><div class="btn-group dropdown float-end" style="
    width: 200px;
"><button name="button" type="submit" class="btn btn-light" data-bs-toggle="dropdown" control-id="ControlID-1" style="
    /* width: 100px; */
">Generate Unit Tests</button><ul class="scrollable dropdown-menu" role="menu"><li><hr class="dropdown-divider"></li><li><form action="/collections" accept-charset="UTF-8" method="post"><input type="hidden" name="authenticity_token" value="lRcflGrerJYE57qJJCGOlhhznawS1lep2_QLgdGtrotZ7UnAIkgwhhNEfU-AKjcipmzBt--HYEthT65W-rfwIw" autocomplete="off"><div class="create-collection input-group"><input class="form-control custom-input" placeholder="Collection name" type="text" name="collection[title]" id="collection_title"><input value="359" autocomplete="off" type="hidden" name="collection[task_ids]" id="collection_task_ids"><input type="submit" name="commit" value="Create new Collection" class="btn btn-light submit" data-disable-with="Create Collection"></div></form></li></ul></div></div></div></div>
```

</details>

> I could also remove that box completely and only include the one button at the bottom :)

As you like.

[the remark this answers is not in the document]
Melhaya: Yea I know. The first image shows. the second one is broken. But thanks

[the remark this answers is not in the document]
MrSerth: Works on my machine 😅 

Bildschirmfoto 2024-06-28 um 11 58 34

<hr>


343725671-24eaf944-ddd9-4cee-8060-c326a98d357a

db/schema.rb
Melhaya: This I did with @kkoehn. Any idea why this changed? and should I change it back to 1?

app/services/task_service/gpt_generate_tests.rb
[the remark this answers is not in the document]
Melhaya: Sorry, I misunderstood what you were asking before. 
The model's explanation varries everytime in term of size and content. Showing directly what the explanation of the model is, might cause more confusion sometimes. the output would need to be analysed and refined before it can be shown back to the user showing a clearer explanation and guides to improve. Another area of future work I think that can be useful.

db/schema.rb
MrSerth: Yes, please. Since you haven't touched the corresponding migration, we should introduce such a change here.

app/services/task_service/gpt_generate_tests.rb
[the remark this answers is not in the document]
MrSerth: Okay, fine for me. Would you please mind to create a follow-up ticket where you collect all the ideas we gathered during the current implementation / review? For example, I am thinking of:
- per-user tokens for ChatGPT
- showing the model's error message to users
- improving the file name (extension + name)
- Thinking of further "promotions" (optional)
- Maybe have this view as part of the "edit" view (since this could allow edits before they are persisted). However, this also "hides" the feature more which is not good
- Check a need for rate limiting
- Improved "billing" overview (i.e., on a per-user base)
- Testing the current implementation as part of a user study
- ... some further ideas we had before 😅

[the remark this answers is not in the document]
Melhaya: Sure I can do that. I already had a couple of notes I took earlier from the comments provided by you and Karol.

db/schema.rb
Melhaya: Just to clarify, I should return it back to 1, correct?

MrSerth: Yes, please ensure that your pull request does not change this line at all.

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
Melhaya: They are all visible now :D thanks.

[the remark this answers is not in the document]
MrSerth: Since you dropped the grey box right now (and we'll have the follow-up ticket in a bit), I would suggest to resolve this conversation here (please do so if you agree). Feel free to re-use some of the images in the follow-up ticket, if you like.

MrSerth (approved):
I've tested the changes again locally, and squashed the branch (hence the force-push). Further, I've ensured that our seed data and our production database contain the correct file extensions. With these changes, I am now considering this PR ready to be merged.

db/schema.rb
kkoehn: @MrSerth we had that issue appearing in the past (ie https://github.com/openHPI/codeharbor/pull/1130#discussion_r1486723764).
I'm not sure how you got the `limit: 1` in there (if you were not editing the schema by hand).
This issue will probably pop up on any future schema generation

MrSerth: Thanks for pointing that out, I missed that (because my schema usually contains many unwanted changes). I took a moment to fix all errors potentially present in the schema and submitted PR #1495.

PR #1281, opened 2024-03-01
Switch SAML development provider

MrSerth:
The previously-used samltest.id is no longer in operation, so we switch to MockSAML.


kkoehn (approved):
Looks good and works, thanks for fixing this!
:+1:

PR #1289, opened 2024-03-12
Fix Kramdown issues for task descriptions

### LINKED ISSUE #1095: Quotes in ProFormA Description are escaped
Importing a ProFormA task containing quotes in the description generally works. However, those quotes are shown / stored escaped in CodeHarbor:

Bildschirmfoto 2023-11-17 um 13 15 46

The XML just contains regular quotes:

```xml
<description><![CDATA[<p>
Schreiben Sie ein Programm, das "Hello World!" (ohne die Anf&uuml;hrungsstriche) ausgibt. Das Programm soll eine Methode enthalten, welche den String "Hello World!" zur&uuml;ckgibt. Die Klasse soll "HelloWorld" und die Methode soll "greet" hei&szlig;en.
</p>]]></description>
```

Here is a corresponding task: [helloWorld.zip](https://github.com/openHPI/codeharbor/files/13391511/helloWorld.zip)

kkoehn:
fixed kramdown issues:
- escaping quotes while importing tasks (and converting descriptions to markdown)
- adding linebreaks to long descriptions

fixes #1095

### DIFF
--- a/db/migrate/20240312201146_migrate_task_descriptions_fix_kramdown_quotes.rb
+++ b/db/migrate/20240312201146_migrate_task_descriptions_fix_kramdown_quotes.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class MigrateTaskDescriptionsFixKramdownQuotes < ActiveRecord::Migration[7.1]
+  def up
+    Task.find_each do |task|
+      task.description = fix_kramdown_descriptions(task.description)
+      task.save!(touch: false)
+    end
+  end
+
+  def fix_kramdown_descriptions(string)
+    # removes double escapes of symbols and unnecessary newlines
+    Kramdown::Document.new(string, line_width: -1).to_kramdown.gsub(/\\([\\*_`\[\]\{"'|])/, '\1').strip
+  end
+end
+
+class Task < ApplicationRecord
+end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -46,7 +46,12 @@ module ApplicationHelper
   end
 
   def render_markdown(markdown)
-    ActionController::Base.helpers.sanitize Kramdown::Document.new(markdown, input: 'GFM', hard_wrap: false).to_html.strip
+    ActionController::Base.helpers.sanitize Kramdown::Document.new(
+      markdown,
+      input: 'GFM',
+      hard_wrap: false,
+      smart_quotes: 'apos,apos,quot,quot'
+    ).to_html.strip
   end
 
   def row(options = {}, &block)
--- a/app/services/proforma_service/convert_proforma_task_to_task.rb
+++ b/app/services/proforma_service/convert_proforma_task_to_task.rb
@@ -43,7 +43,7 @@ module ProformaService
     end
 
     def description
-      Kramdown::Document.new(@proforma_task.description || '', html_to_native: true).to_kramdown.strip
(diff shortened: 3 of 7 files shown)


MrSerth (commented):
Looks good, thanks! I haven't execute the code (but just read it), since I tested most code changes already separately.

One question remains, though: Do we have some "migration" strategy for all existing CodeHarbor tasks (~750)? Can we just process all relevant tables and update the texts accordingly? Ideally, this would be as simple as re-storing the object (I wouldn't go with any custom regex or so).

MrSerth:
... and, would you mind to prepare another PR for CodeOcean, please? There, we don't need the migration, since that's the source of truth for the descriptions.

db/migrate/20240312201146_migrate_task_descriptions_fix_kramdown_quotes.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class MigrateTaskDescriptionsFixKramdownQuotes < ActiveRecord::Migration[7.1]
+  def change
+    Task.find_each {|task| task.update(description: fix_kramdown_descriptions(task.description)) }
```
MrSerth, line 5: Just raising the question here again: Do we want to be notified about validation errors by using `update!` rather than `update`?

MrSerth (commented):
Thanks, works fine for the intended cases (quotes and newlines). For the escaped characters, I noticed that we have further characters that are currently escaped:

```md
Python\_Fort Aufgabe 1.11.3
```

```md
The area of a rectangle (**Rectangle**) is height \* width.
```

```md
For the two lists \["a", "b", "c"\] and \["a", "x", "b"\], the ...
```

These are probably the other entries from the `ESCAPED_CHAR_RE`... 
How shall we proceed here?

kkoehn:
> These are probably the other entries from the `ESCAPED_CHAR_RE`... How shall we proceed here?

Since we don't really want any symbols escape we could of course remove them from the regex resulting in `(\$\$)|^[ ]{0,3}(:)`. I'm not sure what the remaining two cases are trying to cover. (`$$` and `   :`).
Alternatively we could overwrite `convert_text` to just return `el.value` like with the `raw_text` option.

MrSerth:
> > These are probably the other entries from the `ESCAPED_CHAR_RE`... How shall we proceed here?
> 
> Since we don't really want any symbols escape we could of course remove them from the regex resulting in `(\$\$)|^[ ]{0,3}(:)`. I'm not sure what the remaining two cases are trying to cover. (`$$` and ` :`).

Both of these (`$$` and ` :`) are custom extensions to Markdown, the first one (`$$`) also being available in GitHub-flavored Markdown:

- https://kramdown.gettalong.org/syntax.html#math-blocks
- https://kramdown.gettalong.org/syntax.html#definition-lists

> Alternatively we could overwrite `convert_text` to just return `el.value` like with the `raw_text` option.

Probably, that's the best option. So, yes, I would go with this approach. 

<hr>

Your code looks fine so far; the only thing I have left is the migration (which, currently, only fixes quotes, but not the other cases). Is there some good solution available or do I need to check the database manually?

MrSerth (approved):
Alright, looks good -- thanks! I've tested the migration locally with various examples and think that it's sufficient now. 👍 

Just tackle the linter comment (which we probably need to disable for the respective line) and then feel free to merge ✅.

PR #1317, opened 2024-03-25
Add specs for User#avatar_format

MrSerth:
While working on #1316, I got a few more tests suggested that I didn't want to remove (since none were present before).

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -5,9 +5,51 @@ require 'nokogiri'
 
 RSpec.describe User do
   describe '#valid?' do
+    let(:user) { create(:user) }
+
     it { is_expected.to validate_presence_of(:email) }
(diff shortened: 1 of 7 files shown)


spec/models/user_spec.rb
```diff
@@ -208,4 +208,48 @@
       end
     end
   end
+
+  describe '#avatar_format' do
```
kkoehn, line 212: I wouldn't add these tests as tests for a private method, but in the `#valid?` block.

kkoehn (changes requested):
Great addition for the user-validations!
I added one small request concerning the locations of the specs.

spec/models/user_spec.rb
MrSerth: Done. ✅

Dome-GER (approved):

kkoehn (approved):

PR #1352, opened 2024-04-14
Align i18n-tasks with CodeOcean

MrSerth:
This PR aligns i18n-tasks with CodeOcean by fixing an error in the CoffeeScript matcher, and adding the new `SlimRowLocaleMatcher`.


Dome-GER (approved):
The customer is probably nice and handy, but I'm not sure if it's worth the extra code and handling. Anyway, looks good.

PR #1354, opened 2024-04-14
Add factory and seed spec

MrSerth:
In a further attempt to align CodeHarbor with CodeOcean, I added the factory and seed specs. While doing so, I fixed the open issues so that the tests pass.

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 class Report < ApplicationRecord
-  belongs_to :exercise
+  belongs_to :task
   belongs_to :user
 end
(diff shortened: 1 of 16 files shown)


Dome-GER (approved):
Thanks, @MrSerth. The code changes look reasonable to me.

I also understand that aligning factories makes sense. What I'm missing here is some context in the PR description for the "why". Why was it `exercise` before and is a `task` now?

What I _do_ see from the code is that `group` and `user` are no longer considered optional. The same applies to `task`s.

MrSerth:
> I also understand that aligning factories makes sense. What I'm missing here is some context in the PR description for the "why". Why was it `exercise` before and is a `task` now?

Oh, sorry, I missed that. When CodeHarbor was created initially, everything was called a `exercise`. This was based on copying over some domain knowledge from CodeOcean, where the same name is used. However, since we designed CodeHarbor to be compatible with ProFormA, some restrictions introduced in the CodeOcean data model invalidated full compatibility. Therefore, about two to three years ago, it was decided to redesign this core model, align it with ProFormA (a superset of CodeOcean exercise), and rename the model from `exercise` to `task`. 

Unfortunately, there are still some leftovers in the code using the old name, and this is another effort to remove another occurrence.

PR #1369, opened 2024-04-24
Content Security: enable unsafe_inline for script_src_elem

kkoehn:
We have a few views that directly return javascript (ie `import_start`), that require a laxer `script_src_elem` policy to work.


MrSerth:
Okay, I see. I'll proceed merging, but would like us to refactor those parts. Enabling `unsafe_inline` is, as the name suggests, unsafe. Thereby, we are partially disabling the protection introduced by the Content Security Policy.

Issue #1249, opened 2024-02-19
Reference files multiple times


kkoehn:
In ProFormA files can be referenced to multiple times.

We added support for this feature in https://github.com/openHPI/proformaxml/pull/368, but never implemented it in codeharbor.

We would have to change the relation between `TaskFile` and `fileable` to a many-to-many relation and rework some components in the UI (mainly the edit-view) to enable the user to select existing files.

MrSerth:
Do you think that implementing this issue is absolutely necessary, or would we currently "just" duplicate a file that is referenced multiple times? In the latter case, I would see that as an enhancement rather than a bug (since importing and editing a corresponding exercise is still possible, even though a re-export is semantically different).

kkoehn:
With the current implementation an import would fail, since the importer removes the file from the "pool" the first time it is referenced, the second time will break it.
Creating copies would indeed be an easier solution, but we would deviate from the ProFormA spec, which we tried to avoid until now if I remember correctly. If you think we can afford the deviation, I suggest using the "easy" fix as an interim solution and add an issue (with a high priority) to the backlog to refactor and fix this "the right way".

MrSerth:
> Creating copies would indeed be an easier solution, but we would deviate from the ProFormA spec, which we tried to avoid until now if I remember correctly.

Indeed, and I am somewhat sad about that. However, allowing a file to be referenced multiple times is not just a schema change, but also a UI change we need to do. At the current stage, I am somewhat reluctant to do that, given the other topics being worked on, our priorities and the time remaining.

> If you think we can afford the deviation, I suggest using the "easy" fix as an interim solution and add an issue (with a high priority) to the backlog to refactor and fix this "the right way".

Yes, sounds good. I would agree to this solution, since it allows us to proceed with a relatively small compromise for an "edge" case that didn't appear previously. Still, we should definitely keep track of this problem with a dedicated issue and solve it "the right way" (as you suggested).

PR #1370, opened 2024-04-24
Proforma: add support for files that referenced multiple times

kkoehn:
File in ProformaXML can be referenced multiple times. The old implementation prevented importing tasks that made use of that feature.
This solution will create independent `TaskFile`s for every reference, since we current don't support multi-referenced files in codeharbor.
As discussed in #1249

### DIFF
--- a/spec/services/proforma_service/convert_proforma_task_to_task_spec.rb
+++ b/spec/services/proforma_service/convert_proforma_task_to_task_spec.rb
@@ -213,11 +213,25 @@ RSpec.describe ProformaService::ConvertProformaTaskToTask do
         end
       end
 
-      context 'when file is a model-solution-placeholder (needed by proforma until issue #5 is resolved)' do
-        let(:file) { ProformaXML::TaskFile.new(id: 'ms-placeholder-file') }
+      context 'when file is used by multiple tests' do
+        let(:file) do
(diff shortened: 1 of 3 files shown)


spec/services/proforma_service/convert_proforma_task_to_task_spec.rb
```diff
@@ -213,11 +213,31 @@
...
+            used_by_grader: false,
+            visible:,
+            usage_by_lms:,
+            binary:,
+            internal_description: 'internal_description',
+            mimetype:
+          )
+        end
+
+        let(:tests) { [test, test2] }
```
Dome-GER, line 231: Why not inline the `test` and `test2` when they are not referenced explicitly anyway?

Dome-GER (approved):
Some code tweak suggestion.

MrSerth:
I aimed to validate this change, but unfortunately failed to do so. Given this [task 2.0.zip](https://github.com/openHPI/codeharbor/files/15446817/task.2.0.zip), I get the following error:

Bildschirmfoto 2024-05-26 um 10 56 46

I would assume this is due to the XML_ID not being modified with this PR, even when duplicating a file.

@kkoehn can you please check again?

PR #1425, opened 2024-05-27
Fix minor collection issues

kkoehn:
As requested in #1184:
- fix for invitation count
- restrict view of statistics

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -20,7 +20,7 @@ class CollectionsController < ApplicationController
   end
 
   def show
-    @num_of_invites = Message.where(param_type: 'collection', param_id: @collection.id).count
+    @num_of_invites = Message.where(param_type: 'collection', param_id: @collection.id).where.not(recipient_status: 'd').count
   end
 
   def new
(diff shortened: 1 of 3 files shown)


config/locales/en/views/collections.yml
```diff
@@ -3,7 +3,9 @@ en:
   collections:
     collections:
       empty_state:
-        heading: You do not have collections yet
+        heading_favorites: You have not marked a collection as favorite yet
+        heading_mine: You do not have collections yet
+        heading_public: Currently there is no public collection
```
MrSerth, line 8:
```suggestion
        heading_public: Currently, there is no public collection
```

or 

```suggestion
        heading_public: There is currently no public collection
```

config/locales/en/views/collections.yml
```diff
@@ -3,7 +3,9 @@ en:
   collections:
     collections:
       empty_state:
-        heading: You do not have collections yet
+        heading_favorites: You have not marked a collection as favorite yet
```
MrSerth, line 6:
```suggestion
        heading_favorites: You have not yet marked a collection as a favorite
```

MrSerth (approved):
Awesome, thanks for following-up so quickly! I tested the changes locally and think they are fine. Below are just some (minor) translation suggestions, but I would say we're good to go.

Dome-GER (approved):

PR #1426, opened 2024-05-27
Fix task import issues

kkoehn:
As requested in #1370 
- fix missing title in errormessage
- fix duplicate xml_ids


MrSerth (approved):
Nice -- that works great and resolves both issues, too (tested locally)!

Dome-GER (approved):

PR #1326, opened 2024-03-31
Add sign-in flow with NBP Wallet

Mathis-Z:
This is still a work in progress but I would appreciate some feedback.
For SAML I use the mock-IdP instead of the NBP-IdP to get around the certificate issue (when you click on 'Sign-In with NBP' in CodeHarbor you get redirected to mock-SAML). I used the connector configuration of the nbp-school-mock for local testing (just executed docker-compose up in the testconnector directory of the project).

I tried to make the process stateless (on the server at least) by storing the necessary data in the users session and using polling rather than webhooks. The process now is as follows:
1. user signs in via NBP (mock-SAML for now)
2. if a user identity with the received provider-uid/provider combination exists we get signed in
3. otherwise we get redirected to a page where we can scan the wallet QR Code (or enter our email etc. manually); when we request the QR Code a new enmeshed RelationshipTemplate gets created which is valid x minutes
4. a JS snippet on the frontend polls the server which in turn polls the connector to detect when the user has scanned the QR Code
5. when the data entered in the App is valid and the user can be created, the enmeshed Relationship gets created and the user gets signed in

TODOs
- [x] ~~create form for manually entering data (as alternative to QR Code)~~
- [x] check edge cases and error handling
- [x] send verification email
- [x] write tests
- [x] (store users enmeshed address in a second UserIdentity? This might be useful sometime?)
- [x] display welcome message when registration is successful

One problem we have right now is that the app apparently does not recognize when a RelationshipChange gets rejected. It just stays pending. Even when the RelationshipTemplate expires it will still be displayed as pending until the user switches to a different profile and back.

Edit: Apparently, when you register with your wallet, delete the user and then try to create a new CodeHarbor user with the same IdP user, the UserIdentity has not been deleted and you cannot create your account. I just noticed this and did not verify it though.

PR #1436, opened 2024-05-31
Add status group to User model

MrSerth:
In the future, access to some information on CodeHarbor might be restricted due to the status group. For example, only teachers could get access to model solutions.

Further, we need this extension for #1326 and the NBP project.

### DIFF
--- a/db/migrate/20240531160738_add_status_group_to_user.rb
+++ b/db/migrate/20240531160738_add_status_group_to_user.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddStatusGroupToUser < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :status_group, :integer, null: false, limit: 1, default: 0, comment: 'Used as enum in Rails'
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -15,7 +15,7 @@ class User < ApplicationRecord
     :validatable
 
   validates :email, presence: true, uniqueness: {case_sensitive: false}
-  validates :first_name, :last_name, presence: true
+  validates :first_name, :last_name, :status_group, presence: true
   validates :password_set, inclusion: [true, false]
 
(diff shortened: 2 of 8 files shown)


db/migrate/20240531160738_add_status_group_to_user.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddStatusGroupToUser < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :status_group, :integer, null: false, limit: 2, default: 0, comment: 'Used as enum in Rails'
```
Mathis-Z, line 5: This could also be "limit: 1", right?

Mathis-Z (commented):
I am not sure if we should allow users to specify their status group. We would have to remove this once "Mein Bildungsraum" adds support for a verified status group. Maybe we should just leave it at 'unkown' to avoid getting 'dirty' data.

db/migrate/20240531160738_add_status_group_to_user.rb
MrSerth: Yes, indeed. I changed it here (and in the `schema.rb`).

MrSerth:
> I am not sure if we should allow users to specify their status group. We would have to remove this once "Mein Bildungsraum" adds support for a verified status group. Maybe we should just leave it at 'unkown' to avoid getting 'dirty' data.

Nothing easier than that 😁. I agree that it might be useful to avoid potentially incorrect data entered by users manually. Then, getting verified as a teacher is our first feature only available through the integration with "Mein Bildungsraum" 👍.

Mathis-Z:
Okay, I think we can merge the PR :+1:

MrSerth:
@Mathis-Z I saw that you restored the branch -- with a specific reason in mind or my mistake?

Mathis-Z:
I needed it for something but I already forgot for what. I will delete it again.

PR #1439, opened 2024-06-04
task import: handle errors during import better

### LINKED ISSUE #1091: Improve error handling for invalid ProFormA files
Currently, we are just showing a very generic Internal Server Error whenever a ProFormA file could not be processed correctly:

ProFormA error

However, it would be great for users to learn more about the error, so that it can be fixed by themselves (without throwing an exception in the backend). This ticket is designed to improve the error handling in the backend (no exception, do not respond with 503, ...) and frontend (proper error message, do not display `null`, ...).

Here is an invalid example: [task 2.0.1.zip](https://github.com/openHPI/codeharbor/files/13390466/task.2.0.1.zip)

kkoehn:
fixes #1091

removed unnecessary alerts and added more rescues

### DIFF
--- a/config/locales/de/controllers/tasks.yml
+++ b/config/locales/de/controllers/tasks.yml
@@ -10,7 +10,7 @@ de:
       error: 'Der Export der Aufgabe (%{title}) ist fehlgeschlagen. <br> Fehler: %{error}'
       success: Aufgabe (%{title}) erfolgreich exportiert.
     import:
-      internal_error: Bei der Importierung der Aufgabe ist ein interner Fehler aufgetreten.
+      internal_error: Beim Import dieser Aufgabe ist auf CodeHarbor ein interner Fehler aufgetreten.
     import_confirm:
       error: 'Der Import der Aufgabe (%{title}) ist fehlgeschlagen. <br> Fehler: %{error}'
       success: Aufgabe (%{title}) erfolgreich importiert.
--- a/config/locales/de/controllers/tasks.yml
+++ b/config/locales/de/controllers/tasks.yml
@@ -9,15 +9,17 @@ de:
     export_external_confirm:
       error: 'Der Export der Aufgabe (%{title}) ist fehlgeschlagen. <br> Fehler: %{error}'
       success: Aufgabe (%{title}) erfolgreich exportiert.
+    import:
+      internal_error: Bei der Importierung der Aufgabe ist ein interner Fehler aufgetreten.
     import_confirm:
       error: 'Der Import der Aufgabe (%{title}) ist fehlgeschlagen. <br> Fehler: %{error}'
       success: Aufgabe (%{title}) erfolgreich importiert.
     import_external:
-      internal_error: Bei der Importierung der Aufgabe ist ein interner Fehler aufgetreten.
       invalid: Ungültige Aufgabe
       success: Erfolg
     import_start:
       choose_file_error: Sie müssen eine Datei auswählen.
+      error: 'Der Import konnte nicht gestartet werden. <br> Fehler: %{error}'
     proforma_service:
       convert_zip_to_proforma_tasks:
         nested_too_deep: Die ZIP-Datei ist zu tief verschachtelt.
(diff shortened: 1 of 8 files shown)


config/locales/de/controllers/tasks.yml
```diff
@@ -9,15 +9,17 @@ de:
     export_external_confirm:
       error: 'Der Export der Aufgabe (%{title}) ist fehlgeschlagen. <br> Fehler: %{error}'
       success: Aufgabe (%{title}) erfolgreich exportiert.
+    import:
+      internal_error: Bei der Importierung der Aufgabe ist ein interner Fehler aufgetreten.
```
MrSerth, line 13: Just trying to improve the German wording a little; I know that you just moved it (due to the changed key):

```suggestion
      internal_error: Beim Import dieser Aufgabe ist auf CodeHarbor ein interner Fehler aufgetreten.
```

config/locales/en/proforma_errors.yml
```diff
@@ -0,0 +1,4 @@
+---
+en:
+  proforma_errors:
+    version not supported: The version of this ProformaXML document is not supported
```
MrSerth, line 4:
```suggestion
    version not supported: The version of this ProformaXML document is not supported.
```

MrSerth (changes requested):
Nice, thank you! I really like the new error handling and think this will improve the fault tolerance. I have a few minor suggestions, but also one larger one regarding the current error handling: At least in my tests, I was able to generate another exception in the existing error handling (rescuing `ProformaXML::ProformaError`). I annotated two examples and hope that this will allow you to make the handling even more robust.

config/locales/de/proforma_errors.yml
```diff
@@ -1,4 +1,7 @@
 ---
 de:
   proforma_errors:
-    version not supported: Die Version dieses ProformaXML Dokuments wird nicht unterstützt
+    import: Es ist ein Fehler beim Importieren der ProformaXML Zip-Datei aufgetreten.
+    import_default: "%{error}"
+    no task_xml found: Es konnte keine task.xml in der Zip-Datei gefunden werden.
+    version not supported: Die Version dieses ProformaXML Dokuments wird nicht unterstützt.
```
MrSerth, line 7: We recently aimed to unify the spelling to `ZIP-Datei`.

config/locales/en/proforma_errors.yml
```diff
@@ -1,4 +1,7 @@
 ---
 en:
   proforma_errors:
-    version not supported: The version of this ProformaXML document is not supported
+    import: An error occurred while importing your ProformaXML zip-file.
+    import_default: "%{error}"
+    no task_xml found: No task.xml could be found in the supplied zip-file.
+    version not supported: The version of this ProformaXML document is not supported.
```
MrSerth, line 7: In English, we use `ZIP file`.

MrSerth (approved):
Thanks! All errors I've tested were correctly handled and it looks nice. Thanks!

config/locales/de/proforma_errors.yml
```diff
@@ -0,0 +1,7 @@
+---
+de:
+  proforma_errors:
+    import: Es ist ein Fehler beim Importieren der ProformaXML ZIP-Datei aufgetreten.
+    import_default: "%{error}"
```
Dome-GER, line 5: What is the benefit of this locale (in contrast to passing in the `msg` directly to the `default`)?

spec/controllers/tasks_controller_spec.rb
```diff
@@ -736,6 +736,28 @@
         expect(JSON.parse(response.body, symbolize_names: true)).to eql({status: 'failure', message: 'You need to choose a file.'})
       end
     end
+
+    context 'when service throws version not supported error' do
+      before do
+        allow(ProformaService::CacheImportFile).to receive(:call).and_raise(ProformaXML::ProformaError.new(['version not supported']))
+      end
+
+      it 'renders correct json' do
```
Dome-GER, line 745:
```suggestion
      it 'renders correct JSON' do
```

Nitpick. JSON is an acronym. But you probably use `json` everywhere.

Dome-GER (approved):
The `prettify_import_errors` looks a bit brittle. But I also don't have a much better idea right now.

PR #1452, opened 2024-06-10
Adding NBP push connector

Mathis-Z:
This is still missing tests but I thought you could already have a look. It is also missing a config file for solid queue but the defaults are pretty reasonable so maybe we don't even need one.


app/jobs/application_job.rb
MrSerth, line 1: In a follow-up PR, I would be in favor of adding [mission_control-jobs](https://github.com/rails/mission_control-jobs) (with a link for admins) to get some better insights in the processing.

app/jobs/nbp_push_job.rb
MrSerth, line 1: Currently, errors are bubbling up and causing a job to fail. That's okay, per se, but we should think of some error handling for the jobs:

- Logging / Error Reporting to Sentry (keep Sentry limits in mind)
- Potentially a limited number of retries for some errors. ActiveJob has [some built-in retry mechanism](https://edgeguides.rubyonrails.org/active_job_basics.html#retrying-or-discarding-failed-jobs) that we might find useful.

MrSerth (commented):
Awesome, this looks really great and robust already! 💪 I've tested it locally and also took a moment to understand the domain. Many of the design decisions you took are well thought and I am fine with them 👍.

lib/nbp/push_connector.rb
```diff
@@ -0,0 +1,110 @@
...
+    def auth
+      {
+        grant_type: 'client_credentials',
+        client_id: settings.client_id,
+        client_secret: settings.client_secret,
+      }
+    end
+
+    def api_conn
+      @api_conn ||= Faraday.new(url: settings.api_host, headers:)
```
MrSerth, line 74: Since the `api_conn` is memorized, it won't renew the `token`. Hence, five minutes after the first request, any further requests will fail with the current implementation.

config/puma.rb
```diff
@@ -82,3 +82,6 @@
 # - Phased Restarts are only supported in cluster mode with multiple workers (i.e., not in development).
 # - The Puma binary won't be upgraded on phased restarts, but since we have the unattended-upgrades, this is not a major issue.
 # - See https://github.com/casperisfine/puma/blob/master/docs/restart.md.
+
+# Run Solid Queue's supervisor
+plugin :solid_queue
```
MrSerth, line 87: There is one issue with the puma plugin requiring further investigations: https://github.com/rails/solid_queue/issues/241

lib/nbp/push_connector.rb
```diff
@@ -0,0 +1,110 @@
...
+    def push_lom!(xml)
+      response = api_conn.put("/push-connector/api/lom-v2/#{settings.source.slug}") do |req|
+        req.body = {metadata: xml}.to_json
+      end
+      raise_connector_error('Could not push task LOM', response) unless success_status?(response.status)
+    end
+
+    def delete_task!(task_uuid)
+      response = api_conn.delete("/push-connector/api/course/#{settings.source.slug}/#{task_uuid}")
+      raise_connector_error('Could delete task', response) unless success_status?(response.status)
```
MrSerth, line 24: We need to re-think `404` errors when deleting a task.

Let's image I created a task, which was saved successfully but couldn't be pushed due to another error. When I later delete this task again, it isn't found upstream, thus failing this request, too. However, the reported error 404 indicates that the task is (no longer) published online, so somewhat we can just proceed without any further error handling.

app/models/task.rb
```diff
@@ -102,6 +105,14 @@ def self.ransackable_associations(_auth_object = nil)
     %w[labels]
   end
 
+  def push_to_nbp
+    NbpPushJob.perform_later self if Settings.nbp&.push_connector&.enable
+  end
+
+  def remove_from_nbp
+    NbpDeleteJob.perform_later uuid if Settings.nbp&.push_connector&.enable
+  end
```
MrSerth, line 114: We also need to re-consider the visibility level for pushing tasks. Currently, all tasks are pushed regardless of their visibility, aren't they? However, private tasks should remain private.

app/models/task.rb
```diff
@@ -5,11 +5,14 @@
 class Task < ApplicationRecord
   acts_as_taggable_on :state
 
+  before_validation :lowercase_language
+  after_destroy :remove_from_nbp
+  after_save :push_to_nbp
```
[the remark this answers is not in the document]
Mathis-Z: Right, I did not realize `after_destroy_commit` and `after_save_commit` existed.

config/environments/production.rb
```diff
@@ -103,4 +99,8 @@
   # ]
   # Skip DNS rebinding protection for the default health check endpoint.
   # config.host_authorization = { exclude: ->(request) { request.path == "/up" } }
+
+  # Use a real queuing backend for Active Job (and separate queues per environment).
+  config.active_job.queue_adapter = :solid_queue
+  config.active_job.queue_name_prefix = 'codeharbor_production'
```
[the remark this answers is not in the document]
Mathis-Z: I changed it for production.rb but where should I move these lines in test.rb and development.rb?

app/jobs/application_job.rb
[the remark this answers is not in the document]
Mathis-Z: I could not find an option for that...

config/environments/production.rb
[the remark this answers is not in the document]
MrSerth: I'd say put them before `config.action_mailer.perform_caching = false`, which is present in all three environments.

config/puma.rb
MrSerth: Further, there is another warning issues with the introduction of the solid_queue puma plugin (which is new): https://github.com/rails/solid_queue/issues/243

app/jobs/nbp_push_job.rb
Mathis-Z: I added the retry mechanism for Faraday errors.
The other errors (ConnectorError, SettingsError) should already be logged because they are not rescued. I think this is desirable because we also want the job to be marked as failed, right? I don't know if these errors are reported to Sentry though. How would I test that?

app/models/task.rb
Mathis-Z: Done :+1:

app/jobs/nbp_push_job.rb
MrSerth: You can test the Sentry integration by setting the environment variable `SENTRY_DSN` for the Rails server. I just generated a new URL for you and shared it through Slack.

app/jobs/application_job.rb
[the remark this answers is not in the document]
MrSerth: Let's go with the following patch, which will resolve this issue. It will also configure other defaults useful for our production deployment.

<p>

```diff
Subject: [PATCH] Configure defaults for Solid Queue
---
Index: config/application.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/config/application.rb b/config/application.rb
--- a/config/application.rb	(revision e866edd8c98c211d3a3e5a7ca3414a01cd02aae5)
+++ b/config/application.rb	(date 1718742165272)
@@ -50,5 +50,8 @@
 
     # Fix invalid Content-Type header for incoming requests made by edu-sharing.
     config.middleware.insert_before 0, Middleware::EduSharingContentType
+
+    # Configure some defaults for the Solid Queue Supervisor
+    require_relative 'solid_queue_defaults'
   end
 end
Index: app/jobs/application_job.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/jobs/application_job.rb b/app/jobs/application_job.rb
--- a/app/jobs/application_job.rb	(revision e866edd8c98c211d3a3e5a7ca3414a01cd02aae5)
+++ b/app/jobs/application_job.rb	(date 1718734904090)
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
 
 class ApplicationJob < ActiveJob::Base
+  include ActiveRecordLogging
 end
Index: app/jobs/concerns/active_record_logging.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/jobs/concerns/active_record_logging.rb b/app/jobs/concerns/active_record_logging.rb
new file mode 100644
--- /dev/null	(date 1718738811307)
+++ b/app/jobs/concerns/active_record_logging.rb	(date 1718738811307)
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# This module is used to log ActiveRecord queries performed in jobs.
+module ActiveRecordLogging
+  extend ActiveSupport::Concern
+
+  included do
+    around_perform do |_job, block|
+      # With our current Solid Queue setup, there is a difference between both logger:
+      # - *ActiveRecord::Base.logger*: This logger is used for SQL queries and, normally, writes to the log file only.
+      # - *Rails.logger*: The regular logger, which writes to the log file and the console.
+      # For the duration of the job, we want to write the SQL queries to the Rails logger, so they show up in the console.
+      # See config/solid_queue_logging.rb for more information.
+      previous_logger = ActiveRecord::Base.logger
+      ActiveRecord::Base.logger = Rails.logger
+      block.call
+      ActiveRecord::Base.logger = previous_logger
+    end
+  end
+end
Index: config/solid_queue_defaults.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/config/solid_queue_defaults.rb b/config/solid_queue_defaults.rb
new file mode 100644
--- /dev/null	(date 1718747229545)
+++ b/config/solid_queue_defaults.rb	(date 1718747229545)
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# This file must be loaded before other initializers due to the logging configuration.
+Rails.application.configure do
+  # On shutdown, jobs Solid Queue will wait the specified timeout before forcefully shutting down.
+  # Any job not finished by then will be picked up again after a restart.
+  config.solid_queue.shutdown_timeout = 10.seconds
+  # Remove *successful* jobs from the database after 30 days
+  config.solid_queue.clear_finished_jobs_after = 30.days
+  config.solid_queue.supervisor_pidfile = Rails.root.join("tmp/pids/solid_queue_supervisor.pid")
+
+  # For Solid Queue, we want to hide regular SQL queries from the console, but still log them to a separate file.
+  # For the normal webserver, this dedicated setup is neither needed nor desired.
+  next unless Rake.application.top_level_tasks.to_s.include?('solid_queue:')
+
+  # Specify that all logs should be written to the specified log file
+  file_name = "#{Rails.env}.solid_queue.log"
+  config.paths.add 'log', with: "log/#{file_name}"
+
+  # Send all logs regarding SQL queries to the log file.
+  # This will include all queries performed by Solid Queue including periodic job checks.
+  log_file = ActiveSupport::Logger.new(Rails.root.join('log', file_name))
+  config.active_record.logger = ActiveSupport::BroadcastLogger.new(log_file)
+
+  config.after_initialize do
+    # Create a new logger that will write to the console
+    console = ActiveSupport::Logger.new($stdout)
+    console.level = Rails.logger.level
+    # Enable this line to have the same log format as Rails.logger
+    # It will include the job name, the job ID for each line
+    # console.formatter = Rails.logger.formatter
+
+    ActiveSupport.on_load :solid_queue_record do
+      # Once SolidQueue is loaded, we can broadcast its logs to the console, too.
+      # Due to the initialization order, this will effectively start logging once SolidQueue is about to start.
+      Rails.logger.broadcast_to console
+    end
+  end
+end
```

</p>
</details>

config/puma.rb
MrSerth: Based on these issues (and the logging experience), I would recommend not to use the puma plugin. Given the patch supplied above, one can simply start the supervisor through `rails solid_queue:start` or (ideally) through the IDE. 

<details>
<summary>Below are three screenshots showing my configuration</summary>

**Run configuration for Rake task:**
Bildschirmfoto 2024-06-18 um 23 50 37

**Configure a custom log file:**
Bildschirmfoto 2024-06-18 um 23 51 53

**Optionally, use compound to start multiple services together:**
Bildschirmfoto 2024-06-18 um 23 52 07

</details>

With this configuration applied, Solid Queue is debuggable through the IDE (and so are jobs), and logs are split into two tabs:

Bildschirmfoto 2024-06-18 um 23 55 20

The first one contains the most important information, while the second one reading the file log additionally contains UUIDs for the job and more verbose SQL statements.

Given this change, we also need to ensure that the project's README is up-to-date and inform fellow devs about the change via email (ideally with a proposal for their IDE).

db/migrate/20240609104039_create_solid_queue_tables.solid_queue.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+# This migration comes from solid_queue (originally 20231211200639)
+class CreateSolidQueueTables < ActiveRecord::Migration[7.0]
+  def change
+    create_table :solid_queue_jobs do |t|
```
MrSerth, line 6: Solid Queue comes with a dedicated command to clear finished jobs: `SolidQueue::Job.clear_finished_in_batches`. It is currently not run automatically. Therefore, I would suggest to schedule this through `config/schedule.rb` (i.e., daily).

app/jobs/application_job.rb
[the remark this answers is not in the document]
Mathis-Z: I applied and pushed your patch. Thanks :+1:

app/jobs/nbp_push_job.rb
Mathis-Z: The errors are correctly reported to Sentry. When the maximum number of retry attempts have been made the FaradayErrors are also reported.
Do you see the need to limit the error reporting in the current implementation? (You mentioned Sentry limits) I think there should be no case in which these limits get exhausted.

config/puma.rb
Mathis-Z: Okay, I removed the puma plugin. We still have to update the README though.

db/migrate/20240609104039_create_solid_queue_tables.solid_queue.rb
Mathis-Z: What if we set `preserve_finished_jobs=false` instead? I don't see a reason why we should keep these jobs around.

MrSerth: We don't necessarily need the job details, that's right. However, removing finished jobs will, potentially, also reduce the monitoring surface we have and the usefulness of the [mission_control-jobs](https://github.com/rails/mission_control-jobs) plugin, I could assume.

app/jobs/nbp_push_job.rb
MrSerth: Perfect, thanks! In this case, I think almost no further action is required, and especially the limit is okay. 

If I understand the current situation correctly...
- ... a Faraday error is ignored four times and retried, so that the job only fails after 5 attempts?
- ... any other error directly results in a failed job?

In this case, we don't need to worry more about the Sentry limits, since we have some Job limits. It would be different if we would allow unlimited retries each second or some scheduled jobs, which could quickly fill Sentry.

There is just one minor thing I am asking for: Can you please make the attempts explicit? This won't make any code difference, but it includes built-in documentation 🤓 

```rb
retry_on Faraday::Error, wait: :polynomially_longer, attempts: 5
```

db/migrate/20240609104039_create_solid_queue_tables.solid_queue.rb
Mathis-Z: That's true. I added `SolidQueue::Job.clear_finished_in_batches` (and tested it locally)

MrSerth: Thanks :+1: Have you tested the crontab integration? It is just a DSL to edit the crontab file, and at least on my system this currently fails. You can preview the crontab file being written with `whenever` (more options with `whenever --help`, docs [here](https://github.com/javan/whenever)).

The resulting command _should_ work given the environment variables printed as well.

Mathis-Z: Right, I just assumed the code would be executed as Ruby code. It should be fixed now. The cronfile is being generated and when I execute the command in the cronfile, the old jobs are being deleted.

MrSerth: Awesome, thanks! This looks good (without testing it myself), and I would assume it's fine :+1:

Mathis-Z:
@MrSerth and I discussed retrying tasks for responses with status code 500 and replacing the push job with a sync job that checks when the task was last synced with the "NBP Datenraum" and only acts if necessary.

MrSerth:
Today, I went ahead and pushed all existing exercises to "Mein Bildungsraum". While doing so, I discovered two minor issues, which I fixed with #1490 and #1491.

Besides these two, there are a few further things I noticed:

- A return value of 400 is common, when the request body contains an invalid description. At "Mein Bildungsraum", they are treating the input as HTML (which is fine) and thus try to sanitize and parse it. For sanitizing our input, they use [this HTML Sanitizer](https://github.com/mganss/HtmlSanitizer). In this regard, I wondered whether a code of 400 (and maybe others) should be considered "permanent" without retrying. Regardless of how I often I push, the description won't change.
- Besides my push failing with 400, I didn't got a Sentry issue (even after an hour or so).
- One part causing rejections are unpermitted HTML tags or attributes. While some were actually mistakes caused by the missing Markdown render, others are more "permanent". For example, consider the following markdown (gratefully copied from [here](https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/)):
  ```md
  Here is a simple footnote[^1]. With some additional text after it.
  
  [^1]: My reference.
  ```
  This will render with a `class` attribute (or two actually). However, any `class` attribute is not permitted, so that currently any footnote in the text will break the metadata push. I haven't found an option in `Kramdown` (the Markdown renderer) to disable this behavior (and looking at the source code confirms this). Another example would be a table, that we rendered using the kramdown extension `{: .table}` so far (see another, similar example [here](https://kramdown.gettalong.org/quickref.html#tables)).
  
  I am not sure about a good solution to this problem yet. One approach could be to drop any `class` attributes before sending the content to "Mein Bildungsraum" (and potentially extend this list if we learn about more permanent failures), but it doesn't feel right. Mingling with the Kramdown library doesn't feel right, too, so that I would avoid that. Just failing is not ideal either. Any other ideas are welcome.

MrSerth:
> We could copy the list of tags/attributes that are allowed by default in the html sanitizer they use and remove any tags/attributes not matching that. Maybe there is a gem for that too. This will break some descriptions but at least it will not break the export to NBP (assuming they are using the default allow-list and this allow-list does not change in the future).

I also thought about that. However, the assumption you formulated is a rather tough one: Since the sanitizer is an external system that could change configuration-wise at any point in time without prior announcement, it _could_ work, but _could_ also break anytime. Nevertheless, we could go with this option and adjust it with each error, but this doesn't sound too sustainable.

> Another possible solution could be to check if any tag/attribute is not in the default allow-list and if that is the case we strip the original description from all markdown symbols and just export plain text. There might be a gem for this as well. This will keep the description readable in all cases and make sure that the NBP export works.

Removing any markup would work, but would make the exercise description more difficult to read.

<hr>

I consulted @Dome-GER with this problem and while we think it should be the endpoints responsibility to drop undesired tags or attributes, we concluded that copying the default allow-list and restricting it on further errors is probably the most suitable way to move forward. Therefore: Could you try to proceed with this approach, please? [This gem](https://github.com/rails/rails-html-sanitizer), which is already added to the Gemfile with Rails (and currently used with `ActionController::Base.helpers.sanitize` and `ActionView::Base.sanitized_allowed_tags`), might come handy.

lib/nbp/push_connector.rb
```diff
@@ -0,0 +1,114 @@
...
+        missing_source_keys = %i[organization name slug] - settings_hash[:source].keys
+        raise SettingsError.new("Nbp::PushConnector source is missing some settings: #{missing_source_keys}") if missing_source_keys.any?
+      else
+        raise SettingsError.new('Nbp::PushConnector is disabled but got accessed')
+      end
+    end
+
+    def success_status?(status_code)
+      (200..299).cover?(status_code)
+    end
```
nenock, line 104: `response.success?` takes care of that for you :slightly_smiling_face: 
https://www.rubydoc.info/gems/faraday/0.7.4/Faraday%2FResponse:success%3F

app/jobs/application_job.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class ApplicationJob < ActiveJob::Base
```
MrSerth, line 3: In a new `ApplicationJob` file, usually, the following two comments appear. We can check whether we want to enable some of this behavior for our use case:

```rb
  # Automatically retry jobs that encountered a deadlock
  # retry_on ActiveRecord::Deadlocked

  # Most jobs are safe to ignore if the underlying records are no longer available
  # discard_on ActiveJob::DeserializationError
```

Mathis-Z: I think deadlocks should not be possible right now and DeserializationErrors should not occur because we now pass the uuid instead of the entire task to the SyncJob. The SyncJob then handles both existing tasks and not existing tasks.

Mathis-Z:
I added the sanitization of the html LOM description and tested it with some common markdown features. While doing so I noticed some weird behavior in the markdown rendering on CodeHarbor: When editing the task I can see tables and checklists being rendered correctly but when I save the task and view it, the checklists and tables are not being rendered correctly. Is the rendering done by different libraries?

app/jobs/application_job.rb
MrSerth: Okay, that's true. Still (you know me 🤓), I would suggest to copy these comments to our file (and would potentially enable them, since they don't harm and are als applicable to other jobs [future jobs and integrated jobs, such as mail sendings, which will also be handled by SolidQueue now, or ActiveStorage purge jobs]).

MrSerth:
> I added the sanitization of the html LOM description and tested it with some common markdown features. 

Thanks for doing so; I only skimmed through your code so far, but think this looks fine 👍.

> While doing so I noticed some weird behavior in the markdown rendering on CodeHarbor: When editing the task I can see tables and checklists being rendered correctly but when I save the task and view it, the checklists and tables are not being rendered correctly. Is the rendering done by different libraries?

Unfortunately, yes. We have a long tradition of this behavior at openHPI, where the markdown preview is rendered client-sided and the actual markdown rendered server-sided. Despite using the same underlying markdown, there are enough differences. For the server-side rendering, I found this issue `https://github.com/gettalong/kramdown/issues/172` (no link by design to avoid cross linking from their issue), so it is indeed not implemented. Tables are, somewhat, working, as shown [here](https://codeharbor.openhpi.de/tasks/316).

Long story short: Yes, there are differences, and, yes, I am aware of them. We would need to render the preview server-sided as well to align the results (as done by GitHub, for example), but that's nothing we'll tackle too soon (I am afraid).

app/jobs/nbp_sync_all_job.rb
MrSerth, line 1: Especially given the further improvements today, I would suggest adding a comment in this file:

- Why is the order relevant (first querying the "Datenraum"), then doing the `Task.find_each`
- Why aren't we filtering for visibility or so below.

app/services/lom_service/export_lom.rb
```diff
@@ -38,7 +38,9 @@ def oml_general(xml)
         end
         xml.language @task.iso639_lang
         xml.description do
-          xml.string ApplicationController.helpers.render_markdown(@task.description), language: @task.iso639_lang
+          html_fragment = Loofah.fragment(ApplicationController.helpers.render_markdown(@task.description))
+          html_fragment.scrub!(NbpScrubber.new)
```
MrSerth, line 42: Here (and in the `NbpScrubber`) you are (implicitly) using two dependencies: `loofah` and `rails-html-sanitizer`. 

Just thinking aloud for a moment: Those two requirements are not explicitly listed (yet) in our `Gemfile` (only `Gemfile.lock`). The code currently works, since both are included implicitly through `actionpack` and `actionview`, which are dependencies for `rails`. If Rails switches to another library (or, for example drops the usage of `loofah`, which is not part of github.com/rails), our code will break. Hence, I was wondering whether we should list one (`rails-html-sanitizer`) or both (`rails-html-sanitizer`, `loofah`) dependencies directly in our `Gemfile.lock`).

What do you think?

app/services/lom_service/nbp_scrubber.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module LomService
+  class NbpScrubber < Rails::HTML::PermitScrubber
+    ALLOW_LIST = YAML.safe_load_file(Rails.root.join('app/services/lom_service/nbp_scrubber_allow_list.yml'))
+    def initialize
```
MrSerth, line 6:
```suggestion

    def initialize
```

I am still in favor of a new line here, if I may suggest so 🤓

lib/nbp/push_connector.rb
```diff
@@ -0,0 +1,115 @@
...
+    class SettingsError < Error; end
+    class ServerError < Error; end
+
+    def initialize
+      super
+      create_source! unless source_exists?
+    end
+
+    def self.enabled?
+      Settings.nbp&.push_connector&.enable || false
```
MrSerth, line 17: Shall we memoize this value?

```suggestion
      @enabled ||= Settings.nbp&.push_connector&.enable || false
```

lib/nbp/push_connector.rb
```diff
@@ -0,0 +1,115 @@
...
+      raise_connector_error('Failed to create source', response) unless response.success?
+    end
+
+    def token
+      if @token.present? && @token_expiration > 10.seconds.from_now
+        @token
+      else
+        update_token
+      end
+    end
```
MrSerth, line 60: Do we need to consider concurrency here? By default, multiple jobs are executed concurrently. Is each job worker (three by default, if I am right) having an own instance of the `PushConnector` or are they sharing the same? If all workers share the same instance, multiple workers might identify that the token is about to expire and request a new one (resulting in a duplicated token refresh).

MrSerth (commented):
Awesome, thanks! Just some minor code-level comments 🙂

app/services/lom_service/export_lom.rb
Mathis-Z: Yes, I added both.

app/jobs/application_job.rb
Mathis-Z: Done

lib/nbp/push_connector.rb
Mathis-Z: Yes, the workers all share the same PushConnector since in the default config the workers are spawned as 3 threads in a single process thus sharing global variables. I confirmed this by spawning a 100 jobs and having them print their push connector.
I made the PushConnector thread-safe now with `ThreadLocalVar`. This means each thread maintains an own token but this should not be a problem. What do you think?

lib/nbp/push_connector.rb
```diff
@@ -0,0 +1,141 @@
...
+
+    class Error < StandardError; end
+    class SettingsError < Error; end
+    class ServerError < Error; end
+
+    def initialize
+      super
+      @token = Concurrent::ThreadLocalVar.new
+      @token_expiration = Concurrent::ThreadLocalVar.new
+      @connector_settings = Concurrent::ThreadLocalVar.new
```
MrSerth, line 17: Are we expecting any write changes to the connector settings? Otherwise, I would recommend to initialize them once here and don't worry about a `Concurrent::ThreadLocalVar`.

lib/nbp/push_connector.rb
MrSerth: Can't we or do we have issues with concurrency? 

```
@mutex = Mutex.new # in the initializer
@enabled ||= @mutex.synchronize { Settings.nbp&.push_connector&.enable || false }
```

lib/nbp/push_connector.rb
MrSerth, line 1: Can we make some methods private in this file? I think, this would be useful to provide a cleaner, public API.

lib/nbp/push_connector.rb
MrSerth: Thanks, this looks good! While I would have used a mutex, a thread-local variable is a very nice design 👍. 

Besides having one (or two) further questions on concurrency, I am resolving this thread already (there are new dedicated comments).

app/jobs/nbp_sync_all_job.rb
```diff
@@ -0,0 +1,21 @@
...
+class NbpSyncAllJob < ApplicationJob
+  def perform
+    uuids = Set[]
+
+    # add all uploaded uuids so we delete the ones that are still uploaded but not not in local DB
+    offset = 0
+    while (uploaded_uuids = Nbp::PushConnector.instance.get_uploaded_task_uuids(offset)).size.positive?
+      uuids.merge(uploaded_uuids)
+      offset += 100
+    end
```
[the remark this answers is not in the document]
Mathis-Z: That's true. I think the new version looks much better :+1:

lib/nbp/push_connector.rb
Mathis-Z: No, there should be no changes. I removed the thread local var here.

lib/nbp/push_connector.rb
Mathis-Z: Yes, we can :muscle:

lib/nbp/push_connector.rb
Mathis-Z: I did not want to deal with mutexes just for memorizing a variable but I think we don't even need mutexes here. Since all threads have the same Settings object I think we can use a normal variable because read accesses to the settings are not a problem and the writes to `@enabled` should always be identical so there is no problem with threads overwriting each other.

Mathis-Z: This assumes, of course, that a variable getting the same value written to by two threads at the same time will always end up with that value and not break in some weird way.

MrSerth: Alright, thanks. I think this should work correctly 👍

app/jobs/nbp_sync_all_job.rb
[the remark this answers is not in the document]
MrSerth: Thanks!

app/jobs/nbp_sync_all_job.rb
MrSerth, line 1: I didn't know what we decided: Shall we schedule this job to run from time-to-time?

docs/LOCAL_SETUP.md
```diff
@@ -201,6 +203,15 @@ This will launch a dedicated server on port 3045 (default setting) and allow inc
...
 This will launch the CodeHarbor web application server on port 7500 (default setting) and allow incoming connections from your browser.
 
+3. Solid Queue supervisor (optional):
+
+  ```shell
+  bundle exec rake solid_queue:start
+  ```
+  
+Solid Queue is currently only used to synchronize task metadata with the "NBP Datenraum".
+The synchronization is disabled by default. Unless you enable it, you can skip starting the supervisor.
```
MrSerth, line 213: Are we sure some other Rails components are not using the default job backend, e.g., for sending e-mails or purging deleted files (test on the profile page with the avatar)?

lib/nbp/push_connector.rb
```diff
@@ -0,0 +1,152 @@
...
+      elsif response.status == 404
+        false
+      else
+        raise_connector_error('Could not determine if source exists', response)
+      end
+    end
+
+    private
+
+    # the NBP API does not accept limits over 100
```
MrSerth, line 63: Wouldn't this be a dedicated check, so that we `raise` in case a limit > 100 is requested?

spec/fixtures/files/nbp/nodes.json
```diff
@@ -0,0 +1,151 @@
+{
+  "total":5,
```
MrSerth, line 2: Aren't there 3 nodes in the given example?

app/jobs/nbp_sync_all_job.rb
Mathis-Z: Right, I forgot that. I scheduled it for 3 AM every day now. I am not sure how the timezone behavior is but in my testing the time seemed to be in UTC so the job should run at 5 am German time.

MrSerth: Yes, the time zone is usually UTC. Our servers also run with UTC, so that this will be 4am CET / 5am CEST.

docs/LOCAL_SETUP.md
MrSerth: Thanks!

MrSerth:
I just pushed some (final) adjustments in a dedicated PR. Mainly, I am further improving the test coverage, raising it to 100%. During the creation of the additional tests, I also noticed a minor bug in the existing specs (regarding the token renewal). Hence, I feel it is worth to cover as many lines as possible.

Please have another look; I am happy to finally proceed with merging following your check.

lib/nbp/push_connector.rb
```diff
@@ -0,0 +1,153 @@
...
+    def initialize
+      super
+      @token = Concurrent::ThreadLocalVar.new
+      @token_expiration = Concurrent::ThreadLocalVar.new
+
+      create_source! unless source_exists?
+    end
+
+    def self.enabled?
+      @enabled ||= Settings.nbp&.push_connector&.enable || false
```
MrSerth, line 23: Memoizing a boolean value is difficult, and we should probably switch to [another method](https://dev.to/codeandclay/a-memoization-gotcha-14c6).

Mathis-Z:
Thank you for writing the additional tests. I think the PR is ready to be merged. I would either squash all commits or maybe squash all except for "Sanitizing exported html in LOM object description". This would require reordering the commits though.

lib/nbp/push_connector.rb
Mathis-Z: I pushed a fix :+1:

MrSerth (approved):
Awesome, thanks for double-checking. The two commits you suggested sound great. I cannot rebase and reorder them on my mobile right now. Either you do so (allowing me to merge on the go) or I’ll do so later.

PR #1484, opened 2024-06-26
Remove Inline styling for various pages

zuhrasofyan:
Remove inline styling and if possible, subtitute with bootstrap style classes

### DIFF
--- a/app/assets/stylesheets/base.css.scss
+++ b/app/assets/stylesheets/base.css.scss
@@ -86,7 +86,7 @@ a.classic-link {
     width: 96%;
 
     .row-label,
-    .inner-section .row-label {
+    .row-label {
       text-align: left;
       background-color: $gray-bg-color;
       font-weight: 700;
@@ -97,19 +97,14 @@ a.classic-link {
       border-right: 1px #dddddd solid;
     }
     .row-value,
-    .inner-section .row-value {
+    .row-value {
       text-align: left;
       padding: 6px 8px;
-    }
-
-    .inner-section {
-      float: left;
-      display: inline-block;
     }
   }
 
   .row.vertical,
-  .inner-section .row-vertical {
+  .row-vertical {
     .row-label {
       width: 100%;
       display: block;
--- a/app/assets/stylesheets/base.css.scss
+++ b/app/assets/stylesheets/base.css.scss
@@ -79,6 +79,10 @@ a.classic-link {
     border-bottom: none;
   }
 
+  .lead-label {
+    font-size: 1.15em;
+  }
+
   .row {
     margin: 8px 2%;
     border-radius: 2px;
@@ -284,4 +288,9 @@ fieldset.form-group {
 
 .dropdown-title {
   padding-left: 2px;
+}
+
+// Hide the element (useful for hiding element, but does not want to use bootstrap d-none class that persist display: none !important)
+.hide {
+  display: none;
 }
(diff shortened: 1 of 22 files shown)


app/assets/stylesheets/base.css.scss
```diff
@@ -86,7 +86,7 @@ a.classic-link {
     width: 96%;
 
     .row-label,
-    .inner-section .row-label {
+    .row-label {
```
MrSerth, line 89: In this file, some CSS selectors might be duplicated (like here). I see at least three occurrences.

app/views/tasks/_form.html.slim
```diff
@@ -113,7 +113,7 @@
                     = test.remove_nested_fields_link :text, class: 'btn btn-main btn-sm float-end' do
                       i.fa-solid.fa-xmark.text-danger
                       =< t('.button.remove_test')
-                .test-container style="margin-left: 10px; #{display_container}"
+                .test-container.ms-2.mt-4 style="#{display_container}"
```
MrSerth, line 116: Aren't we aiming to get read of all inline styles? Maybe, in a follow-up PR (or another commit here), we should refactor those "dynamic" styles, too. According to my quick code search, there are still 11 of these left (at least one might be difficult regarding labels and their colors, but the remaining 10 should be refactorable, e.g., using CSS classes).

zuhrasofyan: Most of the inline styles left are triggered by javascript and/or dynamically changed. When I tried to change it to CSS classes, it just did not work as intended. 

The best way IMO is to review the changes here and if everything has been resolved, to merge it and make a new PR dedicated to modification of the js to remove remaining inline style, just for the sake of not making this PR bloated and easier to review. I will make another PR draft for the JS modification one. Is that alright?

MrSerth: Sure, let's go with a follow-up PR.

app/views/tasks/show.html.slim
```diff
@@ -536,7 +536,7 @@
     .expandable-panel
       = button_tag type: 'button', class: 'full-window-button show-comment-button', data: {task_id: @task.id} do
         => t('.button.show_comments')
-        span.fa-solid.fa-circle-notch.fa-spin.wait style="display:none;margin-right:5px;"
+        span.fa-solid.fa-circle-notch.fa-spin.wait.me-1 style="display:none;"
```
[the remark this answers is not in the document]
kkoehn: why not use the `hide`-class?
```suggestion
        span.fa-solid.fa-circle-notch.fa-spin.wait.me-1.hide
```

You could remove the `$wait_icon.show()` with `$wait_icon.removeClass('hide')` and `hide()` with `$wait_icon.addClass('hide')`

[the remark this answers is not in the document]
zuhrasofyan: Thank you! done exactly that in the new branch that focused on the js-bound inline styles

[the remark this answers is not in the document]
MrSerth: Just for reference: The new changes are included in [branch `clean_inline_styles_within_script `](https://github.com/openHPI/codeharbor/compare/clean_inline_styles_within_script).

MrSerth:
@kkoehn Would you mind to have a look here again, please? Thanks! 🙏

kkoehn (approved):
looks good to me, thanks for cleaning up the codebase

PR #1490, opened 2024-06-30
Disable footnote backlinks in Markdown

MrSerth:
These backlinks don't work reliably and are thus more harmful than useful.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -50,7 +50,8 @@ module ApplicationHelper
       markdown,
       input: 'GFM',
       hard_wrap: false,
-      smart_quotes: 'apos,apos,quot,quot'
+      smart_quotes: 'apos,apos,quot,quot',
+      footnote_backlink: nil
     ).to_html.strip
   end
 


Mathis-Z (commented):
When testing locally, this does not fix the problem you described in #1452. The converted HTML still contains class attributes. Is this expected?

MrSerth:
> When testing locally, this does not fix the problem you described in #1452. The converted HTML still contains class attributes. Is this expected?

Yes, that's expected. However, if you disable the new option, another link _after_ the footnote is added, which further contains another `class` attribute and didn't work in my tests.

We are talking about two different things here:

**HTML:**
> Bildschirmfoto 2024-07-01 um 10 52 57
This PR removes the ↩ link

**Source Code:**
> Bildschirmfoto 2024-07-01 um 10 53 39
It only tackles the element with the `class="reversefootnote"`

My comment in #1452 is about the element with `class="footnote"`

<hr>

As you can see, the reverse footnote is added with an href anchor. However, Since the counterpart is not added as ID to some text, it won't work (or I don't know how it should). The issue with the footnote itself still exists and is open.

Dome-GER (approved):

MrSerth:
@Mathis-Z What's the current status here? Are you fine with the changes I proposed (and see a change locally) or do you still have concerns?

Mathis-Z:
Yes, I am fine with your changes :+1:

PR #1491, opened 2024-06-30
LOM: Return description in HTML rather than Markdown

MrSerth:
Some IEEE LOM clients, such as "Mein Bildungsraum" expect the description to contain valid HTML. Without parsing the description, we would return the raw Markdown, which is not necessarily correct HTML.

### DIFF
--- a/app/services/lom_service/export_lom.rb
+++ b/app/services/lom_service/export_lom.rb
@@ -38,7 +38,7 @@ module LomService
         end
         xml.language @task.iso639_lang
         xml.description do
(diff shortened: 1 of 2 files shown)


MrSerth:
I also checked the LOM standard and couldn't find any details. A more general web search didn't reveal any common implementation details either. However, when importing a LOM file with HTML markup in the description, our edu-sharing instance will happily reuse most tags (some are disallowed, but just get removed rather than blocking the import):

> Bildschirmfoto 2024-07-01 um 13 29 44

That having said, I would assume returning HTML is fine.

PR #1495, opened 2024-07-01
Fix errors in migrations and schema

MrSerth:
This PR fixes two issues with the migration and schema.

- The first commit ensures we can run `db:migrate:reset` on an empty database by ensuring the model classes are specified correctly.
- The second commit ensures the schema actually matches PostgreSQL capabilities.

_PS: While the migrations can now run successfully, I have not validated or touched them. Probably, some could have been designed better._

### DIFF
--- a/db/migrate/20230110212941_populate_task_file_xml_ids.rb
+++ b/db/migrate/20230110212941_populate_task_file_xml_ids.rb
@@ -1,6 +1,29 @@
 # frozen_string_literal: true
 
 class PopulateTaskFileXmlIds < ActiveRecord::Migration[6.1]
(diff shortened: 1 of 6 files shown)


kkoehn (approved):
Looks good, but I believe you mean `db:migrate:reset`

MrSerth:
> I believe you mean `db:migrate:reset`

Oh, yeah, I do. Sorry, I was not copying the original command I ran several times but rewrote it for the comment -- my bad. I updated the original post; thanks for spotting this mistake.

Dome-GER (approved):

Issue #1492, opened 2024-07-01
Improve unit test generation


MrSerth:
The purpose of this ticket is to collect ideas for improving the unit test generation process.

- [x] per-user tokens for ChatGPT (https://github.com/openHPI/codeharbor/pull/1496)
- [ ] showing the model's error message to users
- [ ] A more visible box for the generate unit test button
- [ ] improving the file name (extension + name)
- [ ] Thinking of further "promotions"
- [ ] Maybe have this view as part of the "edit" view (since this could allow edits before they are persisted). However, this also "hides" the feature more which is not good
- [ ] Check a need for rate limiting
- [x] Improved "billing" overview (i.e., on a per-user base)
- [x] Testing the current implementation as part of a user study
- [x] Improved model output as a json. Could help improve data extraction from the model (#1627)

Probably, we should create sub-tickets for the respective items before tackling them.

MrSerth:
Since we switched to a per-user API key, there is no need for billing or rate limiting any longer (users are spending their own money).

PR #1496, opened 2024-07-02
Allow users to configure a personal OpenAI API key

Melhaya:
Having an API key per user will allow users to utilize the AI capabilities within CodeHarbor (i.e. Unittest generator)

- The generate unittest button is now always visible
- If key is invalid or missing, an alert is shown to the user
- If key is available and valid, a unittest will be created for the task


Melhaya:
I managed to add an API key text area for users to input their key. When I edit the profile and add a key and update, its still blank. I am not really sure what I need to do to make it persist.

@MrSerth @kkoehn : With the current changes I made, am I on the right track to have each user have their own API key? Once I can persist the update in the profile, the next step will be to actually use that for the unit test generator. 

Screenshot 2024-07-02 at 14 35 15

kkoehn:
> I managed to add an API key text area for users to input their key. When I edit the profile and add a key and update, its still blank. I am not really sure what I need to do to make it persist.
> 
> @MrSerth @kkoehn : With the current changes I made, am I on the right track to have each user have their own API key? Once I can persist the update in the profile, the next step will be to actually use that for the unit test generator.
> 
> I also noticed that the `schema.db` had also changed again. do I need to turn it back to how it was? I only added the line `t.string "personal_api_key"` but noticed that the code also worked without it. I can remove that line.
> 
> Update: I was able to persist the API key in the user profile by adjusting the `RegistrationsController` file. Current challenge is to use the API key given by a user in the unit test generator. Not sure which file to adjust for this. need guidance on how to proceed. Thanks

Good work on figuring out the strong parameters!

To use user-specific API-Keys you will have to touch multiple points:
1. The `task_policy` should check for the user-key and not the "application-wide key"
2. The "application-wide key" and all remnants should be removed
3. Add the API-Key as parameter to your `GptGenerateTests`-Service
4. In your `generate_test`-Action in the `tasks_controller` supply the API-Key to the service-call

I hope this gives you a way forward, if you need more help, let me know!

app/models/user.rb
```diff
@@ -17,6 +17,7 @@ class User < ApplicationRecord
   validates :email, presence: true, uniqueness: {case_sensitive: false}
   validates :first_name, :last_name, :status_group, presence: true
   validates :password_set, inclusion: [true, false]
+  validates :personal_api_key, allow_blank: true, length: {maximum: 255}
```
MrSerth, line 20: Since there could be multiple APIs or API keys, I would probably go with a more specific name hinting that this is for OpenAI.

Melhaya: Yup, good point. I will adjust that.

I am pushing a working solution very soon ;)

MrSerth: > Yup, good point. I will adjust that.

Thanks 🙏 

> I am pushing a working solution very soon ;)

Awesome, that's great news and pretty fast 💪

Melhaya:
Everything should now work as intended. There is only one 'issue' that I can't seem to trace.

In the code, I am raising `Gpt::InvalidApiKeyError` if the user's API key is empty or invalid. the alert message is shown correctly when the API key is invalid. but for some reason, I get a `You are not authorized for this action.` alert when the API key is empty. It was working before correctly but not anymore and I am not sure what caused this 'new' alert message as I never used it before. Any hints to fix this is appreciated :)

app/views/users/show.html.slim
```diff
@@ -46,6 +46,12 @@
         | :
       .col.row-value
         = @user.email
+    .row
+      .col-auto.row-label
+        = t('.personal_openai_api_key')
+        | :
+      .col.row-value
+        = @user.personal_openai_api_key
```
kkoehn, line 54: Instead of just showing the API-Key, I would just show the information if a key has been entered. That way it is less likely to unintentionally leak secret information

app/policies/task_policy.rb
```diff
@@ -54,7 +54,7 @@ def manage?
   end
 
   def generate_test?
-    Settings.open_ai.access_token.present? and update?
+    user.present? and update?
```
kkoehn, line 57: I think this should be something like this:
```suggestion
    @user.personal_openai_api_key.present? and update?
```

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -2,11 +2,14 @@
 
 module TaskService
   class GptGenerateTests < ServiceBase
-    def initialize(task:)
+    def initialize(task:, user:)
       super()
       raise Gpt::MissingLanguageError if task.programming_language&.language.blank?
 
       @task = task
+      @user = user
```
kkoehn, line 10: Why did you decide to pass the `User` instead of the API-Key itself? I would prefer the latter. Either way you are assigning `@user` here and don't use it anywhere outside the initialize, so you could just inline it in the line below.

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -2,11 +2,14 @@
...
+      @access_token = @user.personal_openai_api_key.presence
```
kkoehn, line 11: Why are we calling the `api_key` `access_token` now?

db/migrate/20240702224053_add_personal_openai_api_key_to_users.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPersonalOpenaiApiKeyToUsers < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :personal_openai_api_key, :string
```
kkoehn, line 5: I think calling it `personal_openai_api_key` may be to specific, I think `openai_api_key` would be enough, since it already is tied to the user

spec/controllers/tasks_controller_spec.rb
```diff
@@ -1043,17 +1043,16 @@
   end
 
   describe 'POST #generate_test' do
-    let(:task_user) { create(:user) }
+    let(:task_user) { create(:user, personal_openai_api_key: 'valid_api_key') }
     let(:access_level) { :public }
     let(:task) { create(:task, user: task_user, access_level:) }
+    let(:mock_client) { instance_double(OpenAI::Client) }
+    let(:mock_models) { instance_double(OpenAI::Models, list: {'data' => [{'id' => 'text-davinci-002'}]}) }
```
kkoehn, line 1050: You are testing an implementation detail of your service here. That should be tested in the spec for the service itself.
Instead you should be mocking the service here and create different context, where it behaves differently (raise error, etc)

spec/policies/task_policy_spec.rb
```diff
@@ -32,40 +32,12 @@
...
-      end
-
       context 'with gpt access token' do
         before do
-          Settings.open_ai.access_token = 'access_token'
-        end
-
-        after do
-          Settings.open_ai.access_token = nil
+          allow(user).to receive(:personal_openai_api_key).and_return('access_token')
```
kkoehn, line 40: You could add a variable attribute to the user in line 30 and only set it when needed.

app/views/users/registrations/edit.html.slim
```diff
@@ -51,6 +51,14 @@
               autocomplete: 'new-password',
               class: 'form-control'
 
+    .form-group.field-element
+      = f.label :personal_openai_api_key, class: 'form-label'
```
kkoehn, line 55: The label is missing proper capitalization:

image

kkoehn (changes requested):
Thanks for the quick work with this PR. It looks good in general and seems to work fine, but I found some points that could be improved and had some questions.

app/views/tasks/show.html.slim
```diff
@@ -573,10 +573,9 @@
             i.fa-regular.fa-clone
             =< t('common.button.duplicate')
 
-      - if policy(@task).generate_test?
```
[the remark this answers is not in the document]
Melhaya: With the policy check, the button only appears if you are a user and with an API key. but since I am handling now the missing or invalid key, I want the feature to be always visible. does that make sense?

app/views/users/show.html.slim
Melhaya: how to do that?

app/policies/task_policy.rb
Melhaya: when I did it like that, the errors were not showing properly. It should be possible to click the button regardless of whether you have a key or not it think

app/services/task_service/gpt_generate_tests.rb
Melhaya: becasue the user might not have an API key available. so it made sense to enter the user instead of the key. is that wrong?

app/services/task_service/gpt_generate_tests.rb
Melhaya: I will try to clean up this area better

spec/controllers/tasks_controller_spec.rb
Melhaya: I am not sure I understand. Here I am testing that the validation of the key is correct. doesn't that belong here?

Melhaya: Aha I see, I looked at the full code and understand what you mean. will remove them as they are not used here as you mentioned

spec/policies/task_policy_spec.rb
Melhaya: Not sure what I need to do here. could you give an example, please?

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
kkoehn: In that case I would hide the button, if the user is not able to `update` the task and disable it with an informative tooltip if they have no key.

app/views/users/show.html.slim
kkoehn: see new suggestion

app/views/users/show.html.slim
```diff
@@ -46,6 +46,12 @@
         | :
       .col.row-value
         = @user.email
+    .row
+      .col-auto.row-label
+        = t('.openai_api_key')
+        | :
+      .col.row-value
+        = @user.openai_api_key
```
kkoehn, line 54:
```suggestion
      .col.row-value
        - if @user.openai_api_key.present?
          = 'entered'
        - else
          = 'not entered'
```

Please use translations and try to think of better word than (not) entered.

app/policies/task_policy.rb
kkoehn: As described above, I would only let the user click the button, if they have the key. Otherwise disable and tooltip.

app/services/task_service/gpt_generate_tests.rb
kkoehn: I would still just pass the API-Key and check for `present?`. The service doesn't care about any other information from the `User` but the Key.

spec/policies/task_policy_spec.rb
kkoehn: You can compare it to the `task` created in line 8-11.
You have to add an attribute to the `create(:user)` in line 30 (`create(:user, open_ai_key:)`) and then initialize it with nil (`let(:open_ai_key) {nil}`).
In your context you can then just overwrite the let with the desired value

db/schema.rb
```diff
@@ -326,7 +326,8 @@
     t.datetime "locked_at"
     t.string "preferred_locale"
     t.boolean "password_set", default: true, null: false
-    t.integer "status_group", limit: 1, default: 0, null: false, comment: "Used as enum in Rails"
+    t.integer "status_group", limit: 2, default: 0, null: false, comment: "Used as enum in Rails"
```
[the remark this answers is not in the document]
MrSerth: We merged #1495, so that a rebase will ensure the schema in this branch is up-to-date.

app/views/users/show.html.slim
Melhaya: Does `Vorgesehen` and `Nicht Vorgesehen` work?

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
Melhaya: I am a bit confused as to why we need to do this?
Since I am handling the fact that a user might have an empty key, why do we need to disable it? There are alerts for missing or invalid keys already in place.

Also by adding the extra condition in the policy, my alert is no longer visible because I then get 'you are not authorized for this' alert

I feel like my approach is handling all cases better. but maybe I am misunderstanding something?

P.S I have tackled all your other concerns and will push the changes now. :)

app/views/users/show.html.slim
kkoehn: "Vorgesehen" does not fit here.
Maybe "entered" and "eingegeben" would be fine. I can't really think of better alternatives

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
kkoehn: I think my approach would be more intuitive, since the user can immediately see what he is able to do. Showing him the "generate test" button when is he not even able to edit the task, doesn't make sense to me. And in the same way suggesting, that you can use the feature (by showing the button) without having the ability (since the API-Key is missing) does not feel right either. 
It might make sense to show that the feature exists, that's why I suggested disabling (greying out) the button and adding a tooltip.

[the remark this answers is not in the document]
Melhaya: I feel like this is more about how each of us would like to see the feature look like at this point :D 
I think both solutions are viable and which one is more intuitive would have to be decided by the user (future experiment)

I am okay with disabling the button with some kind of message that a key is needed to use it. Could you help show how this can be done? i have no idea how to make a tooltip even with the help of chatgpt it doesn't make sense :D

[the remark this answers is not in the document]
Melhaya: Additionally I was planning to add a block for the generate unit test instead of the button to make it more visible with a clear message that this is a beta-feature. sebastian suggested something like that in the first PR.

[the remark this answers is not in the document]
kkoehn: for the disabling you could look at https://github.com/openHPI/codeharbor/blob/d41c1ffd09f3900e438b60d2f0f5ce2e4b9b0a22/app/views/groups/show.html.slim#L88 and the tooltip: https://github.com/openHPI/codeharbor/blob/3030671711248276c7c46d8df8730d23c230252b/app/views/tasks/_tasks.html.slim#L71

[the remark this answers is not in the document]
kkoehn: > Additionally I was planning to add a block for the generate unit test instead of the button to make it more visible with a clear message that this is a beta-feature. sebastian suggested something like that in the first PR.

I'm not sure what you mean by "using a block"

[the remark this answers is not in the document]
Melhaya: a block like the `add to collection` one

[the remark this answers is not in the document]
kkoehn: Ah I see you discussed it here right? https://github.com/openHPI/codeharbor/pull/1261#discussion_r1656853037
I would add it to the list in #1492 and tackle it separately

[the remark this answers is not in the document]
Melhaya: Yes exactly. It will be done separately from this PR

app/views/users/show.html.slim
Melhaya: Is there a translation for `provided` that fits?

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
Melhaya: @kkoehn is this how the code should look like:

```
    = link_to generate_test_task_path(@task), method: :post, class: "btn btn-important #{'disabled' unless current_user.openai_api_key}", data: {bs_toggle: 'tooltip', bs_placement: 'top', bs_title: t('.button.api_key_required')} do
      i.fa-solid.fa-wand-magic-sparkles
      = t('.button.generate_test')
```

The button is still clickable though like that, what is missing?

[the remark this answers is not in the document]
Melhaya: when I do like this:

```
      - if policy(@task).generate_test?
        = link_to generate_test_task_path(@task), method: :post, class: 'btn btn-important' do
          i.fa-solid.fa-wand-magic-sparkles
          = t('.button.generate_test')
      - else
        = link_to '#', method: :post, class: 'btn btn-important disabled', data: {bs_toggle: 'tooltip', bs_placement: 'top', bs_title: t('.button.api_key_required')} do
          i.fa-solid.fa-wand-magic-sparkles
          = t('.button.generate_test')
```

The button is disabled when the key is not provided, but no tooltip appears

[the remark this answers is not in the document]
Melhaya: The issue seems to arise because disabled elements often don’t trigger hover or focus events, which are required to show tooltips.

In my opinion, just disabling the button and not showing why is not ideal at all and I wouldn't go with this approach if that's the case

app/views/users/show.html.slim
Melhaya: Maybe `Inserted` and `Not inserted`? what would be a good german translation for that?

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
Melhaya: I figured it out by not linking the button to anything without disabling it: is that a viable solution:

```
      - if policy(@task).generate_test?
        = link_to generate_test_task_path(@task), method: :post, class: 'btn btn-important' do
          i.fa-solid.fa-wand-magic-sparkles
          = t('.button.generate_test')
      - else
        = link_to '#', class: 'btn btn-important', data: {bs_toggle: 'tooltip', bs_placement: 'top', bs_title: t('.button.api_key_required')} do
          i.fa-solid.fa-wand-magic-sparkles
          = t('.button.generate_test')
```

Pushing all changes now :)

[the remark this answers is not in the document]
MrSerth: I don't want to jump into this discussion here, but the tooltips @kkoehn is referring to are working for a disabled button. [Check this task](https://codeharbor.openhpi.de/tasks/202?locale=en) in a private browsing window and you'll see:

Bildschirmfoto 2024-07-08 um 23 49 29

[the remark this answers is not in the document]
Melhaya: hmm that's weird then, it's not working for me when I test it locally. could you share the code for this please? :)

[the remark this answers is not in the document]
Melhaya: I did this and it didn't work:

```
      - if policy(@task).generate_test?
        = link_to generate_test_task_path(@task), method: :post, class: 'btn btn-important' do
          i.fa-solid.fa-wand-magic-sparkles
          = t('.button.generate_test')
      - else
        = link_to '#', class: 'btn btn-important disabled', data: {bs_toggle: 'tooltip', bs_placement: 'top', bs_title: t('.button.api_key_required')} do
          i.fa-solid.fa-wand-magic-sparkles
          = t('.button.generate_test')
```

Only when I remove the `disabled` part, it works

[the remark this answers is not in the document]
MrSerth: Are you referring to this one? 

https://github.com/openHPI/codeharbor/blob/3030671711248276c7c46d8df8730d23c230252b/app/views/tasks/show.html.slim#L566-L574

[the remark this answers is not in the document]
Melhaya: Yes. sorry I didn't see that in the file :) 
I was able to make it work in a similar way. thanks for your help

app/views/users/show.html.slim
kkoehn: `Inserted` as well as `Provided` makes it sound like something that the system provides and not the user. 
I'd go for `Entered` / `Not entered` or `(hidden)` / `Not entered`

Melhaya: Understood. Done and pushed. 

Everything should be tackled now ✌🏼

spec/policies/task_policy_spec.rb
```diff
@@ -27,26 +28,22 @@
...
 
     context 'when user is admin' do
-      let(:user) { create(:admin) }
+      let(:user) { create(:admin, openai_api_key:) }
 
       context 'without gpt access token' do
+        let(:openai_api_key) { nil }
+
         it { is_expected.to forbid_only_actions %i[generate_test] }
       end
```
kkoehn, line 43: You could just do this, because the `openai_api_key` is already set to `nil` in line 12

```suggestion
        it { is_expected.to forbid_only_actions %i[generate_test] }
```

Melhaya: I will apply the same in other similar contexts

Melhaya:
@kkoehn  I don't understand why the tests are currently failing. what changed?

Could you also help me resolve the conflict, please? I have updated my branch but the conflict still exists.

Melhaya:
@kkoehn Thanks for fixing the issue 🎉

Everything you commented on should now be tackled and resolved. Is there anything else you would like to change before merging this? :)

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
kkoehn: Your solution was almost what I wanted. I made a commit to which implements it all the way - please check, if you find it acceptable

app/policies/task_policy.rb
kkoehn: In the commit mentioned above, I also change the policy, it checks only for the `update?` right now. The API-Key presence check is independent

kkoehn (commented):
I added some more notes and committed an example, how the button could be implemented

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -2,11 +2,13 @@
 
 module TaskService
   class GptGenerateTests < ServiceBase
-    def initialize(task:)
+    def initialize(task:, openai_api_key:)
       super()
       raise Gpt::MissingLanguageError if task.programming_language&.language.blank?
 
       @task = task
+      @openai_api_key = openai_api_key.presence
```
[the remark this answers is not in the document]
Melhaya: It was useful here in the scenario when there was no openAI key given by the user. but since we changed that, it should be okay to remove it I think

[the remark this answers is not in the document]
Melhaya: I removed it and everything still works as intended.

app/assets/stylesheets/base.css.scss
```diff
@@ -275,6 +275,10 @@ fieldset.form-group {
   margin-bottom: 15px;
 }
 
+.btn.disabled {
+  pointer-events: auto;
+}
+
```
MrSerth, line 281: This change comes unexpected. I haven't executed the code myself, but was wondering why this is needed now? We are not adding the first disabled button, are we?

app/models/user.rb
```diff
@@ -17,6 +17,7 @@ class User < ApplicationRecord
   validates :email, presence: true, uniqueness: {case_sensitive: false}
   validates :first_name, :last_name, :status_group, presence: true
   validates :password_set, inclusion: [true, false]
+  validates :openai_api_key, allow_blank: true, length: {maximum: 255}
```
MrSerth, line 20: Where does this limit of 255 come from? Is it imposed by OpenAI or some assumption?

app/services/task_service/gpt_generate_tests.rb
```diff
@@ -64,5 +66,18 @@ def training_prompts
...
+    def validate_api_key!
+      if @openai_api_key.blank?
+        raise Gpt::InvalidApiKeyError
+      else
+        begin
+          response = client.models.list
+          raise Gpt::InvalidApiKeyError unless response['data']
+        rescue Faraday::UnauthorizedError, OpenAI::Error
+          raise Gpt::InvalidApiKeyError
+        end
```
MrSerth, line 79: Wouldn't it make sense to have this validation when saving an API key? Sure, it could have been revoked in the mean time, but then we might also be able to identify that with a response received from OpenAI?

Just thinking aloud, since this check means we are making many API requests despite being not directly interested in the result.

app/views/tasks/show.html.slim
```diff
@@ -574,9 +574,13 @@
             =< t('common.button.duplicate')
 
       - if policy(@task).generate_test?
```
MrSerth, line 576: I know you discussed about this line already... I am still wondering why you are making `generate_test?` a synonym of `update?` effectively. Wouldn't it make sense to have both, the OpenAI API key check and the check for `update?` combined? Then, we would need two cases here (pseudo-code):

```
- if generate_test?
  # show enabled button
- elsif edit?
  # show disabled button
- else
  # show no button // no op
```

app/views/users/registrations/edit.html.slim
```diff
@@ -51,6 +51,14 @@
               autocomplete: 'new-password',
               class: 'form-control'
 
+    .form-group.field-element
```
MrSerth, line 54: Shall we add a small hint (see example in line 45) about the key's usage and where to get it (https://platform.openai.com/api-keys)?

spec/controllers/tasks_controller_spec.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require 'webmock/rspec'
```
MrSerth, line 4: This import shouldn't be required here (haven't run the code), since it is already loaded through `rails_helper`.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -9,7 +10,6 @@
   let(:collection) { create(:collection, users: [user], tasks: []) }
   let(:valid_attributes) { {user:, access_level:} }
   let(:access_level) { :private }
-
```
MrSerth, line 12: This change seems unrelated, is it?

spec/services/task_service/gpt_generate_tests_spec.rb
```diff
@@ -2,36 +2,72 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require 'webmock/rspec'
```
MrSerth, line 5: Same as above: Why is this required?

MrSerth (commented):
Just some minor feedback...

app/assets/stylesheets/base.css.scss
Melhaya: @kkoehn added it.

But I think the reason is that with this solution, the default `disabled` prevents the tooltip from working and by overwriting it like that, it allows the tooltip to work. But this is just a guess :)

app/models/user.rb
Melhaya: I think the max length is 51. Was not sure if I should make it strict to that or not as things might change. 255 seemed reasonable

app/services/task_service/gpt_generate_tests.rb
Melhaya: The response from openAI to an invalid key is this `Faraday::UnauthorizedError`. There is no other response.

It made more sense to me to do this check here and everything related to the generate_tests feature in one place. makes it easier to make changes in the future.

app/views/tasks/show.html.slim
Melhaya: there is no option to not show the button at all. the button will be visible regardless.

I added the check for the openAI key but @kkoehn removed it in the last commit. but the functionality is still working as intended. not sure if its then really needed to have that check

app/views/users/registrations/edit.html.slim
Melhaya: Thats a good idea. will add that now. :)

spec/controllers/tasks_controller_spec.rb
Melhaya: not sure what you are referring to here?

spec/services/task_service/gpt_generate_tests_spec.rb
Melhaya: removed. it was not needed

spec/controllers/tasks_controller_spec.rb
Melhaya: I probably just removed an empty space I think. that's all. Is that what you are talking about?

app/views/tasks/show.html.slim
Melhaya: Karol mentioned that the API key check is done independently from here. https://github.com/openHPI/codeharbor/pull/1496#discussion_r1671268295

app/services/task_service/gpt_generate_tests.rb
Melhaya: @MrSerth I have changed the design as discussed, which works as intended. Now many of the specs are failing. I am very stuck..  Need some help figuring out what is the issue with the specs and how to cover the changes in the user_specs file. not sure if I should push with the tests failing (?) 
@kkoehn would you maybe have some time to look at them with me today?

Melhaya: this is one of the errors for reference. They are all similar

```
WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET https://api.openai.com/v1/models with headers {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Authorization'=>'Bearer valid_api_key', 'Content-Type'=>'application/json', 'User-Agent'=>'Ruby'}

You can stub this request with the following snippet:

stub_request(:get, "https://api.openai.com/v1/models").
  with(
    headers: {
	  'Accept'=>'*/*',
	  'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
	  'Authorization'=>'Bearer valid_api_key',
	  'Content-Type'=>'application/json',
	  'User-Agent'=>'Ruby'
    }).
  to_return(status: 200, body: "", headers: {})

============================================================

  0) TasksController POST #generate_test when GptGenerateTests is successful redirects to the task show page
     Failure/Error: response = client.models.list

     WebMock::NetConnectNotAllowedError:
       Real HTTP connections are disabled. Unregistered request: GET https://api.openai.com/v1/models with headers {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Authorization'=>'Bearer valid_api_key', 'Content-Type'=>'application/json', 'User-Agent'=>'Ruby'}

       You can stub this request with the following snippet:

       stub_request(:get, "https://api.openai.com/v1/models").
         with(
           headers: {
       	  'Accept'=>'*/*',
       	  'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
       	  'Authorization'=>'Bearer valid_api_key',
       	  'Content-Type'=>'application/json',
       	  'User-Agent'=>'Ruby'
           }).
         to_return(status: 200, body: "", headers: {})

       ============================================================
```

app/assets/stylesheets/base.css.scss
kkoehn: Yes it enables the tooltip for disabled buttons without having to wrap the button in a separate div

MrSerth: Ah, I see - thanks!

While I am fine with the change itself, I am still concerned about consistency: Shall we go with the previous approach or align the other buttons, too?

config/locales/de/external/devise.yml
```diff
@@ -95,6 +95,7 @@ de:
     registrations:
       destroyed: Ihr Account wurde gelöscht. Wir hoffen, Sie bald wiederzusehen.
       edit:
+        api_key_usage: 'Geben Sie hier Ihren OpenAI API-Schlüssel ein, um verschiedene KI-Funktionen innerhalb der Plattform zu nutzen. Einen API-Schlüssel können Sie von der '
```
MrSerth, line 98: If you include the Website here (and thereby complete the sentence), we could also fix the German sentence: There is a verb missing at the end of the sentence. Correct would be something like: `Einen API-Schlüssel können Sie auf der <a ...>Webseite von OpenAI</a> erstellen.`

app/models/user.rb
```diff
@@ -17,6 +17,8 @@ class User < ApplicationRecord
   validates :email, presence: true, uniqueness: {case_sensitive: false}
   validates :first_name, :last_name, :status_group, presence: true
   validates :password_set, inclusion: [true, false]
+  validates :openai_api_key, presence: true, if: -> { openai_api_key.present? }
```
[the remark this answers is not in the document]
Melhaya: You mentioned that both validates and validate are required and this is what I did.

app/views/tasks/show.html.slim
```diff
@@ -573,10 +573,21 @@
             i.fa-regular.fa-clone
             =< t('common.button.duplicate')
 
-      - if policy(@task).generate_test?
-        = link_to generate_test_task_path(@task), method: :post, class: 'btn btn-important' do
-          i.fa-solid.fa-wand-magic-sparkles
-          =< t('.button.generate_test')
+      - if policy(@task).update?
+        - if policy(@task).generate_test?
```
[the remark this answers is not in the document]
Melhaya: @kkoehn please discuss together because it is exhausting to go back and forth 😅

This way the pseudo code that you provided is implemented and it handles all the cases where a user can or cannot create a unit test :)

app/models/user.rb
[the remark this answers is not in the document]
MrSerth: Oh, sorry, then I didn't express clearly: After dropping the length validation, this line can be removed.

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
Melhaya: to reply to this comment: https://github.com/openHPI/codeharbor/pull/1496#discussion_r1672982221

yes it is required to have them both. because while a user might input a valid API key in their profile, it could be that this API key becomes invalid at some point in the future. and so having the tests done in both places insures that this is handled

config/locales/de/external/devise.yml
MrSerth: I am referring to [this comment](https://github.com/openHPI/codeharbor/pull/1496#discussion_r1672994562) regarding the inclusion of the website and otherwise just suggested another German locale and highlighted the link text (between the `<a>` tags).

app/views/users/registrations/edit.html.slim
```diff
@@ -51,6 +51,17 @@
...
+      = f.label :openai_api_key, t('users.show.openai_api_key'), class: 'form-label'
+      = f.text_field :openai_api_key,
+              required: false,
+              placeholder: t('users.show.openai_api_key'),
+              autocomplete: 'off',
+              class: 'form-control'
+      small.form-text.text-body-secondary
+        = t('devise.registrations.edit.api_key_usage')
+        = link_to 'OpenAI API website', 'https://platform.openai.com/api-keys', target: '_blank', rel: 'noopener noreferrer'
```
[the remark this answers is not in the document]
Melhaya: where in config/locales? I thought I used the correct destination as it is in the `registrations.edit` and is used previously for other translations.  Could you be more specific so I can edit it correctly? :)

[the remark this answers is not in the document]
MrSerth: The destination is correct. I am just saying: Please move the string 'OpenAI API website' from the view (where it currently lives) to  the same place where the previous text is stored (currently `devise.registrations.edit.api_key_usage`).

Apropos talking about the previous text: The location `devise.registrations.edit.api_key_usage` doesn't seem perfect, since everything in the `devise` namespace is externally managed (through the Devise gem for authentication). What about `users.registrations.edit.api_key_usage`? In the view, this might be as simple as `.api_key_usage`. (And, finally, with the `openai_` prefix, resulting in: `.openai_api_key_usage`?)

config/locales/de/external/devise.yml
Melhaya: the <a> tag didn't work with the link:

`"Geben Sie hier Ihren OpenAI API-Schlüssel ein, um verschiedene KI-Funktionen innerhalb der Plattform zu nutzen. Einen API-Schlüssel können Sie auf der <a href='https://platform.openai.com/api-keys'>Website von OpenAI</a> erstellen."`

Am i doing it correctly?

Melhaya: I was able to fix it :)

app/services/task_service/gpt_generate_tests.rb
Melhaya: yes it is required to have them both. because while a user might input a valid API key in their profile, it could be that this API key becomes invalid at some point in the future. and so having the tests done in both places insures that this is handled

MrSerth: Sure, an API key could be revoked. I was just wondering whether the call to `client.chat` (in the `TaskService::GptGenerateTests`) wouldn't also throw this same error if something isn't correct with the key.

Melhaya: But with the current setup in place, the code would never go to client.chat if there is something wrong with the key since this will be stopped either when the user inputs the key in their profile or when they attempt to generate a test and the key is again validated before going to client.chat.

config/locales/de/external/devise.yml
Melhaya: I pushed the new changes

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
Melhaya: I think we can remove `update` check from the `generate_test` policy in that case if i understood your comment correctly. Since it is checked separately. would that work?

app/assets/stylesheets/base.css.scss
Melhaya: We reverted back to the previous approach :)

app/views/users/registrations/edit.html.slim
```diff
@@ -51,6 +51,16 @@
...
 
+    .form-group.field-element
+      = f.label :openai_api_key, t('users.show.openai_api_key'), class: 'form-label'
+      = f.text_field :openai_api_key,
+              required: false,
+              placeholder: t('users.show.openai_api_key'),
+              autocomplete: 'off',
+              class: 'form-control'
+      small.form-text.text-body-secondary
+        = t('.openai_api_key_usage', openai_website: link_to(t('.openai_api_website'), 'https://platform.openai.com/api-keys', target: '_blank', rel: 'noopener noreferrer')).html_safe
```
MrSerth, line 62: If you rename the key to something ending with `_html`, you don't need to have the `.html_safe` at the end (see the `info_html` example I linked previously). Further, can we integrate everything except the URL in the respective translation?

```suggestion
        = t('.openai_api_key_usage_html', openai_api_link: 'https://platform.openai.com/api-keys')
```

```yml
openai_api_key_usage_html: Enter your OpenAI API key here to use various AI features within the platform. You can create an API key on the <a href=%{openai_api_link} target='_blank' rel='noopener noreferrer'>OpenAI website</a>.
```

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
MrSerth: Rather than removing the `update` check from `generate_test?`, I would reorder the permission check here.

Isn't it similar to:

```rb
- if policy(@task).generate_test? # A user can update the task and has a working OpenAI API key stored in their profile.
- elsif policy(@task).update? # No OpenAI API key configured is configured for the user yet, but updating is allowed.
```

This would just reduce the nesting by one level.

app/views/users/registrations/edit.html.slim
Melhaya: Thanks for the suggested change. I have updated the code and pushing now with these changes

MrSerth: Thanks! While you fixed the missing quotes in German (that I forgot), you didn't do so in English: `href=%{openai_api_link}`.

Melhaya: aha I thought i did in both. thanks :)

app/views/tasks/show.html.slim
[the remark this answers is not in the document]
Melhaya: But a user can still not be able to update and have an openAI API key? wouldn't that not cover this case?

I am getting confused about shuffling all of this multiple times. I don't see the reason behind both of your decisions 😅 could you please elaborate more why this is better than what is already done for my understanding? :)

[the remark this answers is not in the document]
Melhaya: One other question: There is nesting in other parts of the show file as well. so why is it better there and not in this example?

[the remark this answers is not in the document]
MrSerth: > One other question: There is nesting in other parts of the show file as well. so why is it better there and not in this example?

Yes, there is, and that's not ideal either. The more nesting we have, the harder it is to read. It is just [a question of style](https://rubystyle.guide/#no-nested-conditionals):

> Avoid use of nested conditionals for flow of control.

In the current case, we cannot use any early return, but we can still reduce the nesting. [Here](https://stackoverflow.com/a/33817929) is a quick example and while it is not specific to Ruby, it provides the idea I am trying to apply here. Please correct me if I am wrong!

Then, regarding the other places: We try to live the [boy scouting rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html), aiming to improve code snippets along the way. Hence: If you found other places where nesting is not ideal, please feel free to improve those as well (ideally in at least a dedicated commit or -- even better -- in a dedicated PR). I do so, too, and you can expect some boy scouting improvements by me tomorrow (in dedicated PRs).

<hr>

Otherwise, back to the topic: Which two decisions are you referring to? The `update?` check within `generate_test?` and the suggestion about the less-nested conditional?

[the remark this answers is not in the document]
Melhaya: @MrSerth Thank you for your explanations. I understand your point. And I quickly tested your logic here and it works as intended. I will implement it and push it in a couple of moments 😊
 
> ```ruby
> - if policy(@task).generate_test? # A user can update the task and has a working OpenAI API key stored in their profile.
> - elsif policy(@task).update? # No OpenAI API key configured is configured for the user yet, but updating is allowed.
> ```

app/services/task_service/gpt_generate_tests.rb
Melhaya: would doing a `begin` and `rescue` sequence around the response of `client.chat` help alleviate the worry of the code raising this issue? 
Even though I don't see it ever being activated, there are already 2 checks as I mentioned before. but we can always add extra checks if it doesn't hurt to have and style-wise it's acceptable.

Melhaya:
```
      begin
        response = client.chat(
          parameters: {
            model: Settings.open_ai.model,
            messages:,
            temperature: 0.7, # Lower values insure reproducibility
          }
        )
      rescue Faraday::UnauthorizedError, OpenAI::Error
        raise Gpt::InvalidApiKeyError
      end
```

Melhaya: These are the different [error codes](https://platform.openai.com/docs/guides/error-codes/api-errors) for reference from the openAI website.

We might attempt to catch every possible error in a future PR for clearer direction for the user to follow. But the most common and relevant one in this PR is the invalid API key error.

kkoehn (approved):
All of my review points have been addressed. Good work!

app/services/task_service/gpt_generate_tests.rb
MrSerth: @kkoehn What is your point of view here? I think, we are discussing whether checking the API for each request is required or whether we can just rely on `client.chat` raising an `Faraday::UnauthorizedError` (and, consequentially, use @Melhaya proposal while removing the dedicated check `validate_api_key!` when initializing the service)?

app/models/user.rb
```diff
@@ -146,6 +147,19 @@ def to_s
...
+    return unless openai_api_key_changed?
+
+    client = OpenAI::Client.new(access_token: openai_api_key)
+
+    begin
+      response = client.models.list
+      errors.add(:base, :invalid_api_key) unless response['data']
+    rescue Faraday::UnauthorizedError, OpenAI::Error
+      errors.add(:base, :invalid_api_key)
+    end
```
MrSerth, line 160: To reduce code duplication, I would prefer to have this code exclusively put in the `TaskService::GptGenerateTests`.

Can't we do something like 

```rb
TaskService::GptGenerateTests.validate_api_key! openai_api_key
```

here?

Melhaya: What do you mean? where are you proposing to do this, in the `user.rb` ? and how can this be done? :)

Melhaya: I tried to make it work but I couldn't. Also after some examination, I noticed they are not entirely duplicated. 
In the `user.rb` file I also check if the key has been changed or not to avoid checking it everytime a user updates their profile without changing their key. Also the way the error is raised in the `user.rb' file is different than the one used at the `gpt_generate_tests.rb` file. Simply calling the `validate_api_key` in the `gpt_generate_tests.rb` from the `user.rb` file will cause an error that is not handled in that case.

While the code does a similar function, it is not necessarily the same. I think its better to have them separated but my ruby knowledge is also limited with its capabilities. So maybe you are aware of a way to do it without causing the issues I am facing :)

MrSerth: Thanks for trying, I appreciate that. The first part of the method (checking whether the key was changed), not marked in my comment, needs to stay in the `User` model.

I just pushed my proposal as two commits. Please have a look and test them.

Melhaya: @MrSerth Thank you for you making the changes yourself. I appreciate you taking the time to do that. The generate test specs looks cleaner.

I have tested the API key change and I now receive an error when changing the key to an invalid one. `the server responded with status 503` caused by this line:
```
    def self.validate!(client)
      response = client.models.list
  ```
  
  Not sure what happened?

Comment (Melhaya) [reply app/models/user.rb]: After some investigation, it seems that there is currently an outage in ChatGPT and API assistance services. So the error is not from our side. (But of course, we now need to handle this case as well so the user sees a better error message). 🥹👀

Comment (Melhaya) [reply app/models/user.rb]: @MrSerth I just pushed a commit with a solution for this issue :) 
Please check it out and if it's satisfactory, then feel free to resolve this comment.

Comment (Melhaya) [reply app/models/user.rb]: p.s. I managed to test it quickly before openAI fixed the error and it worked.

Comment (MrSerth) [reply app/models/user.rb]: Thanks! I refactored the error handling a little further, to make it easier extendable and maintainable. Please have another look and feel free to resolve the thread 😉

Comment (Melhaya) [reply app/models/user.rb]: Thanks! this way it will be easier to add future error handling.  Thanks for your feedback and help :)

File: app/services/task_service/gpt_generate_tests.rb (hunk 17 above)
Comment (Melhaya) [reply app/services/task_service/gpt_generate_tests.rb]: So what are we doing with this last comment? 👀

Comment (Melhaya) [reply app/services/task_service/gpt_generate_tests.rb]: @MrSerth, @kkoehn What is needed at this point to get this resolved? :) I would like to move on to further adjustments; having this merged would make it easier and faster.

Comment (kkoehn) [reply app/services/task_service/gpt_generate_tests.rb]: The `gpt_response` method does already have a `rescue` which should catch that case with (`rescue Faraday::Error => e`) therefore I don't see the need to add the extra validation. If you think the error message generated from that rescue isn't specific enough, you could add another rescue for the `UnauthorizedError`.

Comment (Melhaya) [reply app/services/task_service/gpt_generate_tests.rb]: Hey @kkoehn thanks for your reply. What are you recommending we change or is this something for the future?

Comment (Melhaya) [reply app/services/task_service/gpt_generate_tests.rb]: Isn't what you are suggesting, is what is already done in the code here? :

```
      raw_response[/```(.*?)```/m, 1].lines[1..]&.join&.strip
    rescue Faraday::Error => e
      raise Gpt::Error::InternalServerError.new("Could not communicate with OpenAI due to #{e.inspect}")
    rescue Net::OpenTimeout, Net::ReadTimeout, Errno::ECONNRESET, SocketError, EOFError
      raise Gpt::Error.new(e)
    end
```

Comment (Melhaya) [reply app/services/task_service/gpt_generate_tests.rb]: @kkoehn, I have added the UnauthorizedError to the error handling process in the latest commit. please check if this is what you meant. :)

@MrSerth If its all good, you can resolve this comment 👍

Comment (Melhaya) [reply app/services/task_service/gpt_generate_tests.rb]: I am not sure now why I am getting this migration error though. any hint on how to fix it is appreciated

File: config/locales/de/errors.yml [hunk 29]
```diff
@@ -0,0 +1,9 @@
+---
+de:
+  errors:
+    gpt:
+      error: Beim Generieren des Unit-Tests ist ein Fehler aufgetreten. Bitte versuchen Sie es später erneut.
+      internal_server_error: Der OpenAI-Server ist derzeit nicht verfügbar. Bitte versuchen Sie es später erneut, wenn das Problem behoben wurde.
+      invalid_api_key: Der API-Schlüssel in Ihrem Profil ist ungültig. Bitte fügen Sie den entsprechenden API-Schlüssel in Ihrem Profil hinzu, um die KI-Funktionen zu nutzen.
+      invalid_task_description: Die angegebene Aufgabenbeschreibung stellt keine gültige Programmieraufgabe dar und kann daher nicht zum Generieren eines Unit-Tests genutzt werden. Bitte stellen Sie sicher, dass die Aufgabenbeschreibung eine klar formulierte Problemstellung enthält, die durch ein Programm gelöst werden kann.
+      missing_language: Für diese Aufgabe ist keine Programmiersprache angegeben. Bitte geben Sie die Sprache an, bevor Sie fortfahren.
```
kkoehn, line 9: I think the new error has to be added here right?

app/services/task_service/gpt_generate_tests.rb
MrSerth: Which migration error are you referring to? (btw, I would have been great to have this in a dedicated thread).

Melhaya: nevermind. i fixed it :)

config/locales/de/errors.yml
Melhaya: @MrSerth I have added the translations as well as the test that verifies their presence. Please confirm that it is as desired. :)

app/services/task_service/gpt_generate_tests.rb
MrSerth: I refactored the code a little, dropping the custom `UnauthorizedError`. If I get this right (and please correct me if I am wrong), an `UnauthorizedError` translates to an invalid key in our domain. Hence, I used our domain-specific error `InvalidApiKey` for this purpose.

While doing so, I also removed some code duplication regarding the error handling and, in a dedicated commit, also skipped querying the model list during each "generate test" run (this was one of the purposes of this discussion IIRC). Please have another look 🙂

config/locales/de/errors.yml
MrSerth: Thanks! I decided to move the spec to a dedicated file (it only belongs to the `Error` class) and use the `not_to raise_exception`, since this is more generic (and independent from the actual translation).

Melhaya: Aha I see. Thanks.

app/services/task_service/gpt_generate_tests.rb
Melhaya: Thanks for spending the time to refactor the code. I appreciate your input and improvements.

Yes, you are correct about the `UnauthorizedError`

Everything looks good to me. I tested it out and it works. I am just a bit confused about what the `wrap_api_error!` does in the code and unable to follow the syntax. and why the function is there twice once with the `self` and the other without? and how this is better as well? I have a few questions and perhaps you can have the time to walk me through what is written? it would make it easier for me when I am making further changes in the future to understand what is currently the process :)

kkoehn: @MrSerth I think the service does to many things now and even though some parts are very elegant, it's very confusing. 

What do you think about splitting the service into two and maybe even extracting a base-class for the error wrapping?
I sketched an approach here: #1557

MrSerth: @kkoehn Sounds good, thank you! Feel free to continue this approach, I am happy to see these improvements. I found the architecture difficult too, especially regarding the key validation, client, and error handling. During my refactorings a month ago, I also considered a new class to wrap the `OpenAI::Client` completely and catch the different errors there (just passing a single one). This would be a good alternative to your approach, but multiple services are fine too. Would you mind to finish the refactoring?

MrSerth: Thanks for the refactoring, @kkoehn. I just took a brief moment to check with @Melhaya on your changes and the PR in general. Following this discussion, we agreed that it's finally time to ship this improvement. 🎉  

Thanks everyone!

PR #1511, opened 2024-07-09
Update Ruby version

kkoehn:
update ruby version


MrSerth (approved):
Thanks for taking care!

MrSerth (approved):

PR #1527, opened 2024-07-12
Clean inline styles within scripts

zuhrasofyan:
This part is to clean up all remaining dynamic script-controlled inline styles

### DIFF
--- a/app/views/tasks/_editor.html.slim
+++ b/app/views/tasks/_editor.html.slim
@@ -1,22 +1,18 @@
 - file_attached = file.object.attachment.attached?
 - show_attachment = file_attached || file.object.use_attached_file
-- if show_attachment
-  - editor_display = 'display: none'
-  - attachment_display = 'display: block'
-- else
-  - editor_display = 'display: block'
-  - attachment_display = 'display: none'
-- fileupload_display = file_attached ? 'display: none' : 'display: block'
+- editor_class = show_attachment ? 'hide' : ''
+- attachment_class = show_attachment ? '' : 'hide'
+- fileupload_class = file_attached ? 'hide' : ''
 
 = f.label :content, TaskFile.human_attribute_name('content'), class: 'form-label mb-3'
 a.btn.btn-main.btn-sm.toggle-input href='#'
   = t('.button.toggle_file')
 .toggle-divs
-  .edit style=editor_display
+  .edit class=editor_class
     = file.text_area :content, disabled: show_attachment, class: 'd-none form-control'
     .editor
       = file.text_area :content
-  .attachment style=attachment_display
+  .attachment class=attachment_class
     = file.hidden_field :use_attached_file, value: show_attachment, class: 'use-attached-file'
     - if file_attached
       .btn-group.attachment_present
@@ -29,5 +25,5 @@ a.btn.btn-main.btn-sm.toggle-input href='#'
         - if file.object.text_data?
           .extract-text.btn.btn-main data-file-id=file.object.id
             = t('.button.extract_text')
-    .alternative style=fileupload_display
+    .alternative class=fileupload_class
       = file.file_field :attachment, class: 'alternative-input form-control'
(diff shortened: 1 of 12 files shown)


app/views/tasks/_form.html.slim
```diff
@@ -100,20 +100,20 @@
...
-                  - display_container = 'display: none'
+                  - display_container_class = 'd-none'
                   .table-header.closed.toggle-next
                     => test.object.title
                     span.fa-solid.fa-caret-down
                     = test.remove_nested_fields_link :text, class: 'btn btn-main btn-sm float-end' do
                       i.fa-solid.fa-xmark.text-danger
                       =< t('.button.remove_test')
-                .test-container.ms-2.mt-4 style="#{display_container}"
+                .test-container.ms-2.mt-4 class="#{display_container_class}"
```
kkoehn, line 116: shouldn't this be enough?

```suggestion
                .test-container.ms-2.mt-4 class=display_container_class
```

kkoehn (changes requested):
On task `show` and `edit` I have the following problem:

image

Can you check that again please?

app/views/tasks/_editor.html.slim
```diff
@@ -1,22 +1,18 @@
...
-- if show_attachment
-  - editor_display = 'display: none'
-  - attachment_display = 'display: block'
-- else
-  - editor_display = 'display: block'
-  - attachment_display = 'display: none'
-- fileupload_display = file_attached ? 'display: none' : 'display: block'
+- editor_class = show_attachment ? 'hide' : ''
+- attachment_class = show_attachment ? '' : 'hide'
+- fileupload_class = file_attached ? 'hide' : ''
```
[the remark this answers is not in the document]
zuhrasofyan: unfortunately, if we move it to inline, the logic will not be applied until we first click the button, so to toggle it, we have to click the button twice before it operates correctly.

app/views/tasks/_form.html.slim
zuhrasofyan: Thanks. I have corrected it.

app/views/tasks/_editor.html.slim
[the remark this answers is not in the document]
zuhrasofyan: the problem in page task `show` and `edit`  has been fixed:

app/views/tasks/_form.html.slim
zuhrasofyan: Ups, the suggested change broke the slim/test.

```
app/views/tasks/_form.html.slim:103 [W] RuboCop: Lint/UselessAssignment: Useless assignment to variable - `display_container_class`.
app/views/tasks/_form.html.slim:109 [W] RuboCop: Lint/UselessAssignment: Useless assignment to variable - `display_container_class`.
```

Should we leave it as it is, or there is some workaround?

kkoehn: we have the same problem in `app/views/tasks/_nested_file_form.html.slim` where we disabled the cop for these cases. Maybe `"#{display_container_class}"` would be better after all or you could also disable the cop. Up to you, but do it consistently :)

app/views/tasks/_editor.html.slim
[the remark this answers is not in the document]
kkoehn: > unfortunately, if we move it to inline, the logic will not be applied until we first click the button, so to toggle it, we have to click the button twice before it operates correctly.

for me it works fine, I added it like this:
```slim
    .alternative class="#{file_attached ? 'hide' : ''}"
```

app/views/tasks/_form.html.slim
zuhrasofyan: Since we already suppress similar lint warnings, I also use the same approach here.

app/views/tasks/_editor.html.slim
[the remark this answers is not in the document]
zuhrasofyan: Done

kkoehn (approved):
Looks good, thanks!

PR #1550, opened 2024-07-29
1094 select proforma version for export direct download api

kkoehn:
depends on https://github.com/openHPI/proformaxml/pull/409

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -79,8 +79,9 @@ GEM
       minitest (>= 5.1)
       mutex_m
       tzinfo (~> 2.0)
-    acts-as-taggable-on (10.0.0)
-      activerecord (>= 6.1, < 7.2)
+    acts-as-taggable-on (11.0.0)
+      activerecord (>= 7.0, < 8.0)
+      zeitwerk (>= 2.4, < 3.0)
     addressable (2.8.7)
       public_suffix (>= 2.0.2, < 7.0)
     amq-protocol (2.3.2)
@@ -124,7 +125,7 @@ GEM
       coffee-script-source
       execjs
     coffee-script-source (1.12.2)
-    concurrent-ruby (1.3.3)
+    concurrent-ruby (1.3.4)
     config (5.5.1)
       deep_merge (~> 1.2, >= 1.2.1)
     connection_pool (2.4.1)
@@ -171,7 +172,7 @@ GEM
     faraday-net_http (3.3.0)
       net-http
     ffi (1.17.0)
-    fugit (1.11.0)
+    fugit (1.11.1)
       et-orbi (~> 1, >= 1.2.11)
       raabro (~> 1.4)
     glob (0.4.1)
@@ -261,7 +262,7 @@ GEM
     mini_magick (4.13.2)
     mini_mime (1.1.5)
     mini_portile2 (2.8.7)
-    minitest (5.24.1)
+    minitest (5.25.1)
     mnemosyne-ruby (2.0.0)
       activesupport (>= 6.1)
       bunny
@@ -305,7 +306,7 @@ GEM
       ast (~> 2.4.1)
       racc
     pg (1.5.7)
-    proformaxml (1.5.0)
+    proformaxml (1.5.1)
       activemodel (>= 5.2.3, < 8.0.0)
       activesupport (>= 5.2.3, < 8.0.0)
       dachsfisch (~> 1.0.0)
@@ -392,7 +393,7 @@ GEM
       zeitwerk (~> 2.6)
     rainbow (3.1.1)
     rake (13.2.1)
-    ransack (4.2.0)
+    ransack (4.2.1)
       activerecord (>= 6.1.5)
       activesupport (>= 6.1.5)
       i18n
@@ -408,7 +409,7 @@ GEM
     responders (3.1.1)
       actionpack (>= 5.2)
       railties (>= 5.2)
-    rexml (3.3.4)
+    rexml (3.3.6)
       strscan
     rouge (4.3.0)
     rqrcode (2.2.0)
@@ -427,7 +428,7 @@ GEM
     rspec-mocks (3.13.1)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.13.0)
-    rspec-rails (6.1.3)
+    rspec-rails (6.1.4)
       actionpack (>= 6.1)
       activesupport (>= 6.1)
       railties (>= 6.1)
@@ -492,10 +493,10 @@ GEM
       rubyzip (>= 1.2.2, < 3.0)
       websocket (~> 1.0)
     semantic_range (3.0.0)
-    sentry-rails (5.18.2)
+    sentry-rails (5.19.0)
       railties (>= 5.0)
-      sentry-ruby (~> 5.18.2)
-    sentry-ruby (5.18.2)
+      sentry-ruby (~> 5.19.0)
+    sentry-ruby (5.19.0)
       bigdecimal
       concurrent-ruby (~> 1.0, >= 1.0.2)
     set (1.1.0)
@@ -505,7 +506,7 @@ GEM
       rack-proxy (>= 0.6.1)
       railties (>= 5.2)
       semantic_range (>= 2.3.0)
-    shoulda-matchers (6.2.0)
+    shoulda-matchers (6.4.0)
       activesupport (>= 5.2.0)
     simple_form (5.3.1)
       actionpack (>= 5.2)
@@ -527,7 +528,7 @@ GEM
     slim_lint (0.28.0)
       rubocop (>= 1.0, < 2.0)
       slim (>= 3.0, < 6.0)
-    solid_queue (0.4.1)
+    solid_queue (0.6.0)
       activejob (>= 7.1)
       activerecord (>= 7.1)
       concurrent-ruby (>= 1.3.1)
@@ -631,7 +632,7 @@ DEPENDENCIES
   omniauth-rails_csrf_protection (~> 1.0)
   omniauth-saml (~> 2.0)
   pg
-  proformaxml (~> 1.5.0)
+  proformaxml (~> 1.5.1)
   pry-byebug
   pry-rails
   puma
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -79,9 +79,8 @@ GEM
       minitest (>= 5.1)
       mutex_m
       tzinfo (~> 2.0)
-    acts-as-taggable-on (11.0.0)
-      activerecord (>= 7.0, < 8.0)
-      zeitwerk (>= 2.4, < 3.0)
+    acts-as-taggable-on (10.0.0)
+      activerecord (>= 6.1, < 7.2)
     addressable (2.8.7)
       public_suffix (>= 2.0.2, < 7.0)
     amq-protocol (2.3.2)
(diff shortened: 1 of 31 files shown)


app/controllers/collections_controller.rb
```diff
@@ -84,7 +84,7 @@ def push_collection
   end
 
   def download_all
-    binary_zip_data = ProformaService::ExportTasks.call(tasks: @collection.tasks)
+    binary_zip_data = ProformaService::ExportTasks.call(tasks: @collection.tasks, options: {version: params[:version]})
```
MrSerth, line 87: When manipulating the URL (in my browser) and specifying an unsupported version (such as `1.0`) an unhandled exception is thrown (`ProformaXML::PostGenerateValidationError "version not supported"`). Maybe we could handle this error, so that we don't respond with an internal server error in production?

app/controllers/tasks_controller.rb
```diff
@@ -85,7 +85,7 @@ def add_to_collection
   end
 
   def download
-    zip_file = ProformaService::ExportTask.call(task: @task)
+    zip_file = ProformaService::ExportTask.call(task: @task, options: {version: params[:version]})
```
MrSerth, line 88: When manipulating the URL (in my browser) and specifying an unsupported version (such as `1.0`) an unhandled exception is thrown (`ProformaXML::PostGenerateValidationError "version not supported"`). Maybe we could handle this error, so that we don't respond with an internal server error in production?

app/views/collections/show.html.slim
```diff
@@ -89,17 +89,25 @@
...
+          .dropdown.btn-group
             = button_tag class: 'btn btn-important dropdown-toggle', 'data-bs-toggle': 'dropdown' do
-              => t('common.button.export')
+              => t('common.button.download_zip')
               span.caret
             .scrollable.dropdown-menu role='menu'
+              .dropdown-header = "#{t('common.button.available_versions')}: "
               - if @collection.tasks.empty?
                 a.dropdown-item.disabled-cursor title=t('.download_no_exercises')
                   = t('common.button.download_zip')
```
MrSerth, line 100: This is not yet working:

- it looks off
- the dropdown entry is clickable
- the tooltip is not a correct Bootstrap tooltip

Bildschirmfoto 2024-08-09 um 17 13 30

app/views/account_links/_form.html.slim
```diff
@@ -15,6 +15,9 @@
       .input-group
         = f.text_field :api_key, data: {toggle: 'tooltip', placement: 'bottom'}, title: t('.info.token'), class: 'form-control'
         = button_tag t('.button.generate'), type: 'button', class: 'generate-api-key-token btn btn-light'
+    .field-element.form-group
+      = f.label :proforma_version, AccountLink.human_attribute_name('proforma_version'), class: 'form-label'
```
MrSerth, line 19: The translation for this label is missing (in both locales).

app/services/proforma_service/cache_import_file.rb
```diff
@@ -10,15 +10,15 @@ def initialize(user:, zip_file:)
 
     def execute
       data = {}
-      ActiveRecord::Base.transaction do
-        import_file = ImportFileCache.create!(user: @user, zip_file: @zip_file)
-        ProformaService::ConvertZipToProformaTasks.call(zip_file: @zip_file).each do |proforma_task|
-          task = Task.find_by(uuid: proforma_task[:uuid])
+      # ActiveRecord::Base.transaction do
```
MrSerth, line 13: This might still be some work-in-progress, I assume...

MrSerth (commented):
Great work already! If you don't mind, here is some feedback already (besides the "draft" status of the PR). I tested the functionality with CodeOcean already and it worked flawlessly with either version 😁

kkoehn:
@MrSerth thanks for the feedback, I fixed your suggestions. I'm still waiting to merge the proforma PR, so we can also fix those edge-cases.

config/locales/de/models.yml
```diff
@@ -6,6 +6,7 @@ de:
         api_key: API-Schlüssel
         check_uuid_url: URL zum Überprüfen der UUID
         name: Name
+        proforma_version: Proforma Version
```
MrSerth, line 9: Could you please double-check the spelling of "ProFormA"? I would like to have a single, common spelling used throughout the app: "ProFormA".

For compound words, please use "ProFormA-XML" (actually, only in German, but since the group uses the same term with `-` in English too, let's reuse it in both languages). But, if possible, I would drop the "-XML" suffix, since this can create really misleading wordings: "Es ist ein Fehler beim Importieren der ProformaXML ZIP-Datei aufgetreten."

app/controllers/tasks_controller.rb
```diff
@@ -85,8 +85,10 @@ def add_to_collection
...
     send_data(zip_file.string, type: 'application/zip', filename: "task_#{@task.id}.zip", disposition: 'attachment')
+  rescue ProformaXML::PostGenerateValidationError => e
+    redirect_to :root, danger: JSON.parse(e.message).map {|msg| t("proforma_errors.#{msg}", default: msg) }.join('<br>')
```
MrSerth, line 91: What about redirecting to `@task`? I feel, this would increase usability.

app/controllers/collections_controller.rb
```diff
@@ -84,9 +84,10 @@ def push_collection
   end
 
   def download_all
-    binary_zip_data = ProformaService::ExportTasks.call(tasks: @collection.tasks)
-
+    binary_zip_data = ProformaService::ExportTasks.call(tasks: @collection.tasks, options: {version: params[:version]})
     send_data(binary_zip_data.string, type: 'application/zip', filename: "#{@collection.title}.zip", disposition: 'attachment')
+  rescue ProformaXML::PostGenerateValidationError => e
+    redirect_to :root, danger: JSON.parse(e.message).map {|msg| t("proforma_errors.#{msg}", default: msg) }.join('<br>')
```
MrSerth, line 90: What about redirecting to `@collection`? I feel, this would increase usability.

app/views/collections/show.html.slim
```diff
@@ -89,17 +89,26 @@
               .input-group
                 = text_field_tag 'user', nil, form: 'share', class: 'form-control'
                 = submit_tag t('.button.share'), form: 'share', class: 'btn btn-light'
-          .dropup.btn-group
+          .dropdown.btn-group
+            - if @collection.tasks.empty?
+              div data-bs-toggle='tooltip' title=t('.download_no_exercises') data-bs-delay=150
+                = button_tag class: 'btn btn-important dropdown-toggle disabled', 'data-bs-toggle': 'dropdown' do
+                  => t('common.button.download_zip')
```
MrSerth, line 96: The disabled button doesn't wrap correctly in three lines if needed (but the enabled button does):

**Disabled:**
Bildschirmfoto 2024-08-13 um 19 10 27

**Enabled:**
Bildschirmfoto 2024-08-13 um 19 11 12

config/locales/de/views/account_links.yml
```diff
@@ -1,6 +1,9 @@
 ---
 de:
   account_links:
+    common:
+      proforma_version:
+        select_default: aktuellste Version (momentan %{latest_version})
```
MrSerth, line 6: Please ensure to align the English and German locale. The German would translate to "latest version", whereas the one you committed is only "latest" (without "version").

app/models/account_link.rb
```diff
@@ -5,6 +5,8 @@ class AccountLink < ApplicationRecord
   validates :push_url, presence: true
   validates :api_key, presence: true
   validates :name, presence: true
+  validates :proforma_version, inclusion: {in: [*ProformaXML::SCHEMA_VERSIONS], allow_nil: true}
+  normalizes :proforma_version, with: ->(proforma_version) { proforma_version.presence }
```
MrSerth, line 9: Thanks for sharing the `normalizes` with me, I didn't know about it.

app/views/account_links/_form.html.slim
```diff
@@ -15,6 +15,9 @@
...
+      = f.select :proforma_version, options_for_select(ProformaXML::SCHEMA_VERSIONS, @account_link.proforma_version), {include_blank: t('account_links.common.proforma_version.select_default', latest_version: ProformaXML::SCHEMA_VERSION_LATEST)}, {class: 'form-select'}
```
MrSerth, line 20: Do we have any consistent use of select2 vs native selects? I was shortly wondering whether we should use select2 here, too. However, the only other place we have a single select is for a task's license, where we don't enable select2 either. 

The only place where it is used consistently is for the advanced search filter, where also single selects have it enabled.

MrSerth (commented):
Thanks for integrating my feedback already. Here's some follow-up feedback I found, but otherwise I am fine with the changes.

For sure, I'll remind @Dome-GER about the missing review for https://github.com/openHPI/proformaxml/pull/409 to unblock the current PR.

app/controllers/tasks_controller.rb
kkoehn: I wasn't sure what the optimal way would be. Currently (on master) when you click the link you change your current location to start the download, but the task page stays loaded. Therefore I added the `target: '_blank'` to the `link_to`'s. This works well in the success case - a new tabs opens, the download starts and the tab closes again.
In the error case the tab doesn't close and it felt wrong to just redirect back to the task. The obvious answer would be to use ajax to start the download and show the error, but I think that's and over-engineered approach for such an edge case. The user should only get that error if they start messing around with the parameters (like you :) )

app/views/account_links/_form.html.slim
kkoehn: I try to use normal selects where possible and select2 if I want a search box or I want to use another feature of select2

MrSerth: Okay, makes sense. For now, I would say that's fine :+1:.

app/controllers/tasks_controller.rb
MrSerth: Ah, that makes sense -- I didn't notice that behavior in the first place (since I was opening the link in a new tab anyway). Thank that's fine for me.

app/controllers/collections_controller.rb
MrSerth: As discussed for tasks [see here](https://github.com/openHPI/codeharbor/pull/1550#discussion_r1715644539), we agreed to keep the redirect to `:root` since it occurs in a new tab.

MrSerth (approved):

MrSerth:
Since https://github.com/openHPI/proformaxml/pull/409 was merged, we can now draft a new ProFormA version, update this PR and undraft it.

kkoehn:
@MrSerth Do you think we need to address the case of more ProformaXML version being added? Currently it is assumed that every saved task has the latest version. If a new version gets added that could lead to problems. A possible solution could be to add a `proforma_version` column on task. What do you think?

MrSerth (commented):
> @MrSerth Do you think we need to address the case of more ProformaXML version being added? 

Not necessarily. The ProFormA working group is currently not working on a new version, so that I don't expect major changes in the near future (there could be a bugfix release, of course).

> Currently it is assumed that every saved task has the latest version. If a new version gets added that could lead to problems.

I see. I would actually keep that assumption and would ensure that all existing tasks are migrated to a newer version when we add the respective version support in the future.

> A possible solution could be to add a `proforma_version` column on task. What do you think?

Sure, could work. However, as mentioned above, I would not do so right now.

Gemfile.lock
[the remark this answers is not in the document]
kkoehn: fixed

MrSerth (approved):

PR #1553, opened 2024-07-30
Use Bulk Enqueueing for ActiveJob

MrSerth:
With this new approach, only one operation with the ActiveJob backend is required to schedule all jobs.

See the ActiveJob Guide for more details on [Bulk Enqueuing](https://edgeguides.rubyonrails.org/active_job_basics.html#bulk-enqueuing).


Mathis-Z (approved):
Yes, that is better.

Dome-GER (approved):

Dome-GER:
Much better to safe some roundtrips.

PR #1555, opened 2024-07-31
Resolve warnings about unpermitted parameters during task#create and task#update

MrSerth:
As noticed during #1103, the following warnings were present when saving a task:

```
Unpermitted parameter: :uuid.
Unpermitted parameters: :_method, :authenticity_token, :task, :button, :id. 
```

This PR fixes said behavior and includes two minor adjustments in the form builder for tasks, too.

### DIFF
--- a/app/views/tasks/_form.html.slim
+++ b/app/views/tasks/_form.html.slim
@@ -40,7 +40,7 @@
             = f.text_area :internal_description, class: 'form-control'
           .field-element
             = f.label :uuid, Task.human_attribute_name('uuid'), class: 'form-label'
-            = f.text_field :uuid, class: 'form-control'
+            = f.text_field :uuid, class: 'form-control not-allowed', disabled: true, title: t('.unchangeable_uuid'), 'data-bs-toggle': 'tooltip'
           .field-element
             = f.label :parent_uuid, Task.human_attribute_name('parent_uuid'), class: 'form-label'
             = f.text_field :parent_uuid, class: 'form-control'
--- a/app/views/tasks/_form.html.slim
+++ b/app/views/tasks/_form.html.slim
@@ -7,7 +7,7 @@
 
 .row
   .col-md-12.my-4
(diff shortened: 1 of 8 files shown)


app/views/tasks/_form.html.slim
```diff
@@ -40,10 +40,10 @@
             = f.text_area :internal_description, class: 'form-control'
           .field-element
             = f.label :uuid, Task.human_attribute_name('uuid'), class: 'form-label'
-            = f.text_field :uuid, class: 'form-control'
+            = f.text_field :uuid, type: :uuid, class: 'form-control not-allowed', disabled: true, title: t('.unchangeable_uuid'), 'data-bs-toggle': 'tooltip', pattern: '^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
           .field-element
             = f.label :parent_uuid, Task.human_attribute_name('parent_uuid'), class: 'form-label'
-            = f.text_field :parent_uuid, class: 'form-control'
+            = f.text_field :parent_uuid, type: :uuid, class: 'form-control', pattern: '^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
```
Dome-GER, line 46: You could extract the pattern to a useful constant / class method.

Dome-GER (approved):

app/views/tasks/_form.html.slim
MrSerth: I went with a method in the `ApplicationHelper` for now. There, we already have minor helpers for views, so that it aligns with the remaining architecture.

MrSerth: As discussed in person, we agreed that the newly-introduced method is not ideal (since it is just returning a static string), but found it to be the best-fitting solution in the current architecture.

florian-str (approved):

PR #1557, opened 2024-07-31
refactor and split GptService

kkoehn:
suggestion to refactor the `GptGenerateTests` Service to something small and handable


MrSerth (approved):
Thanks for your changes, they look great!

Melhaya (approved):
Thanks for spending the time to do so @kkoehn. This new refactoring is very good. It makes it easier to expand on the service now that it has its own module. It will be very useful moving forward.

PR #1574, opened 2024-08-06
Add more fine granular rating system with sub-categories 

### LINKED ISSUE #1483: More fine granular rating system with sub-categories
The rating system should allow users to rate a task in different categories like: originality, quality of tests, quality of model solutions, etc.
These sub-ratings should be combined into a total rating.

MrSerth:
_to be written_

Closes #1483


app/assets/javascripts/tasks_ratings.coffee
```diff
@@ -2,48 +2,46 @@ ready = ->
...
-
-        overallrating = response.overall_rating
-        $('.starrating span.fa-solid.fa-star[data-rating=1]').attr("color", "red")
-        for num in [1, 2, 3, 4, 5]
-          do (num) ->
-            if overallrating >= num
-              $('.overall-rating[data-rating=' + num + ']').removeClass("fa-star").removeClass("fa-star-half-stroke").removeClass("fa-regular").addClass("fa-star").addClass("fa-solid")
-            else if (overallrating + 0.5) >= num
-              $('.overall-rating[data-rating=' + num + ']').removeClass("fa-star").removeClass("fa-solid").addClass("fa-star-half-stroke")
+        console.log(response.average_rating)
```
MrSerth, line 25: When finishing, we should remove these few log statements.

db/migrate/20240709190021_add_categories_to_rating.rb
```diff
@@ -0,0 +1,30 @@
...
+    add_column :ratings, :test_quality, :integer, null: false, default: 1
+    add_column :ratings, :model_solution_quality, :integer, null: false, default: 1
+
+    ActiveRecord::Base.connection.execute(
+      %(UPDATE ratings SET
+                originality = rating,
+                description_quality = rating,
+                test_quality = rating,
+                model_solution_quality = rating)
+    )
```
MrSerth, line 16: Feel free to use the following:

```suggestion
execute <<~SQL.squish
  UPDATE ratings
  SET originality = rating,
      description_quality = rating,
      test_quality = rating,
      model_solution_quality = rating
SQL
```

db/schema.rb
```diff
@@ -181,11 +181,15 @@
   end
 
   create_table "ratings", id: :serial, force: :cascade do |t|
-    t.integer "rating"
```
MrSerth, line 184: In contrast to the categories listed below, it looks like `NULL` is allowed. Is that the case actually?

app/models/task.rb
```diff
@@ -140,15 +140,26 @@ def initialize_derivate(user = nil)
...
+
+    @average_rating = ActiveRecord::Base.connection.exec_query(
+      "SELECT #{category_averages} FROM ratings WHERE ratings.task_id = $1",
+      'Averaging ratings', [id]
+    ).first.symbolize_keys.transform_values {|rating| (rating || 0).round(1) }
+  end
+
+  def average_rating_stars
+    @average_rating_stars ||= average_rating.transform_values {|rating| (rating * 2).round / 2.0 }
+  end
```
MrSerth, line 155: Manually concatenating an SQL string is working here and, most likely, secure. Still, I am usually not a big fan of that, since one must be very careful when changing the query later on. One potential alternative that is definitely secure (but unfortunately longer):

```rb
def average_rating # rubocop:disable Metrics/AbcSize
  return @average_rating if @average_rating

  ratings_arel = Rating.arel_table

  category_averages = Rating::CATEGORIES.map do |category|
    average_rating = ratings_arel[category].average
    coalesced_rating = Arel::Nodes::NamedFunction.new('COALESCE', [average_rating, 0])
    rounded_rating = Arel::Nodes::NamedFunction.new('ROUND', [coalesced_rating, 1])
    rounded_rating.as(category.to_s)
  end

  condition = ratings_arel[:task_id].eq(id)
  query = ratings_arel.project(category_averages).where(condition)
  result = ActiveRecord::Base.connection.exec_query(query.to_sql, 'Averaging ratings').first

  @average_rating = result.symbolize_keys
end
```

MrSerth (commented):
Great changes, thank you!

db/schema.rb
Mathis-Z: That's right, I added a not-null constraint to be consistent with the other columns.

Mathis-Z:
Do you know if `lib/templates/slim/scaffold/_form.html.slim.erb` is still needed? The commit that added the file is very old and it has not been changed since then. Also the safe delete function of RubyMine does not find any uses?

MrSerth:
> Do you know if `lib/templates/slim/scaffold/_form.html.slim.erb` is still needed? The commit that added the file is very old and it has not been changed since then. Also the safe delete function of RubyMine does not find any uses?
> 
> Edit: I decided to remove the file.

The file is still used, at least somehow 😁 

Rails features so-called generators, which allow someone to add another resource quickly. These generators can either be used for specific parts (like a migration, model, controller, views, tests, ...) or altogether. For example, the following introduces a new model `Repository` and all related files. It already specifies that a repository should have three attributes and one association:

```bash
rails generate scaffold repository name:string description:text remote:string group:references
```

<p>

```log
invoke  active_record
create    db/migrate/20240812125339_create_repositories.rb
create    app/models/repository.rb
invoke    test_unit
create      test/models/repository_test.rb
create      test/fixtures/repositories.yml
invoke  resource_route
 route    resources :repositories
invoke  scaffold_controller
create    app/controllers/repositories_controller.rb
invoke    slim
 exist      app/views/repositories
create      app/views/repositories/index.html.slim
create      app/views/repositories/edit.html.slim
create      app/views/repositories/show.html.slim
create      app/views/repositories/new.html.slim
create      app/views/repositories/_form.html.slim
create      app/views/repositories/_repository.html.slim
invoke    resource_route
invoke    test_unit
create      test/controllers/repositories_controller_test.rb
create      test/system/repositories_test.rb
invoke    helper
create      app/helpers/repositories_helper.rb
invoke      test_unit
invoke    jbuilder
create      app/views/repositories/index.json.jbuilder
create      app/views/repositories/show.json.jbuilder
create      app/views/repositories/_repository.json.jbuilder
```

</p>
</details> 

As part of this generation (specifically for `app/views/repositories/_form.html.slim`) the `lib/templates/slim/scaffold/_form.html.slim.erb` is used. Hence, the resulting file will look like this:

```slim
= simple_form_for(@repository) do |f|
  = f.error_notification

  .form-inputs
    = f.input :name
    = f.input :description
    = f.input :remote
    = f.association :group

  .form-actions
    = f.button :submit
```

As you can see, this really much instantiates the template you deleted. The resulting page doesn't look too nice yet, but is fully functional (especially the reference!):

Bildschirmfoto 2024-08-12 um 15 05 51

All of that was magically created by Rails 🪄

<hr>

That having said, I have to admit that we aren't using generators often and they also lack the latest changes. For example, no pundit policy is generated or considered in the controller, the styling is off, and errors are also caught differently than expected.

<details><summary> Invalid record</summary>
<p>
 
Bildschirmfoto 2024-08-12 um 15 09 32

</p>
</details> 

Without the file, the view for adding or editing entries looks different (mostly better, I would say)

Bildschirmfoto 2024-08-12 um 15 17 30

The only downside is that the "nice" dropdown for the group is lost and replaced with a text field (which expects a group ID). Long story short: I think, I am fine with removing the file, but would wish for a backlog ticket to fix the Rails generators (views, policies, ...) and verify them.

app/controllers/tasks_controller.rb
```diff
@@ -35,7 +35,7 @@ def show
     @tests = @task.tests
     @model_solutions = @task.model_solutions
 
-    @user_rating = @task.ratings&.find_by(user: current_user)&.rating
+    @user_rating = @task.ratings&.find_by(user: current_user) || Rating.new
```
MrSerth, line 38: By the way: Do we need to handle incomplete tasks, where some aspects are missing? For example, when a task doesn't feature any test or model solution, users would still need to rate these. 

I think, there are two possibilities to conclude: Either, one argues that this is bad, since users would otherwise rate with 1 star only, which is still kept after adding a test / model solution (making it more difficult to examine). Or, one argues that it is fair to rate everything, since a missing element justifies a low rating for this category.

What do you think?

app/models/rating.rb
```diff
@@ -1,8 +1,19 @@
...
 class Rating < ApplicationRecord
-  validates :rating, numericality: {only_integer: true, less_than_or_equal_to: 5, greater_than: 0}
+  SUBCATEGORIES = %i[originality description_quality test_quality model_solution_quality].freeze
+  CATEGORIES = SUBCATEGORIES + %i[overall_rating]
+
+  validates :overall_rating, numericality: {only_integer: true, less_than_or_equal_to: 5, greater_than: 0}
+  validates :originality, numericality: {only_integer: true, less_than_or_equal_to: 5, greater_than: 0}
+  validates :description_quality, numericality: {only_integer: true, less_than_or_equal_to: 5, greater_than: 0}
+  validates :test_quality, numericality: {only_integer: true, less_than_or_equal_to: 5, greater_than: 0}
+  validates :model_solution_quality, numericality: {only_integer: true, less_than_or_equal_to: 5, greater_than: 0}
```
MrSerth, line 11: I am not sure whether I like the meta programming more, but an alternative would be:

```suggestion
CATEGORIES.each do |category|
  validates category, numericality: {only_integer: true, less_than_or_equal_to: 5, greater_than: 0}
end
```

app/views/tasks/show.html.slim
```diff
@@ -4,40 +4,40 @@
...
+      .d-flex.align-items-center
+        span.task-star-rating.rating-subcategories-trigger.averaged-task-ratings
+          = render('task_star_rating', rating: @task.overall_rating, category: :overall_rating)
+          .rating-subcategories-overlay
+            = render('rating_subcategories', task: @task)
+        span.mx-2
+          - if current_user && policy(Rating).new_for_task?(@task)
+            .btn.btn-light.btn-sm#ratingModalOpenButton data-bs-toggle="modal" data-bs-target="#ratingModal"
+              = t('.button.rate')
+            .modal.fade#ratingModal aria-hidden="true" data-task-id=@task.id
```
MrSerth, line 17: Adding `tabindex="-1"` will ensure one can close the modal by pressing <kbd>ESC</kbd >. As a publicly-funded project, we strive for great accessibility 😁.

MrSerth, line 17: Maybe, you could consider to extract the modal to a dedicated partial, allowing us to reuse it later? For example, similar to the [structure found in CodeOcean](https://github.com/openHPI/codeocean/blob/master/app/views/shared/_modal.html.slim) (with the necessary adaptions you made already for CodeHarbor).

app/views/tasks/show.html.slim
```diff
@@ -4,40 +4,40 @@
...
+                  .modal-header
+                    h6.modal-title
+                      = Rating.model_name.human
+                    button.btn-close data-bs-dismiss="modal" aria-label="Close"
+                  .modal-body
+                    - Rating::SUBCATEGORIES.each do |category|
+                      span.rating-category-container
+                        span
+                          = t("activerecord.attributes.rating.#{category}")
+                        span.task-rating style="font-size: 1.2em;"
```
MrSerth, line 29: I (and Zuhra) would appreciate if you could remove any inline-styles. We almost got rid of them completely, and don't want to reintroduce them...

_Background: Inline-styles such as this one prevents advanced CSP rules. Instead, it requires `style-src unsafe-inline`. Read more on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#unsafe_inline_styles). 

(Another occurrence is a few lines below).

config/shakapacker.yml
```diff
@@ -69,7 +69,9 @@ development:
     # live_reload: true
     client:
       # Should we show a full-screen overlay in the browser when there are compiler errors or warnings?
-      overlay: true
+      overlay:
+        warnings: false
+        errors: true
```
MrSerth, line 74: I would prefer if you could revert this change (and keep it local). 🙏

app/models/task.rb
```diff
@@ -56,14 +56,14 @@ class Task < ApplicationRecord
                        }.fetch(visibility, public_access)
                      }
   scope :created_before_days, ->(days) { where(created_at: days.to_i.days.ago.beginning_of_day..) if days.to_i.positive? }
-  scope :average_rating, lambda {
-    select('tasks.*, COALESCE(avg_rating, 0) AS average_rating')
+  scope :overall_rating, lambda {
```
MrSerth, line 59: This scope is currently untested (but used in the advanced filter). Could you please add a (single) test?

app/models/rating.rb
```diff
@@ -1,8 +1,19 @@
...
 
   belongs_to :task
   belongs_to :user
+
+  def to_h
```
MrSerth, line 16: Where are you using this method? While I like the method and simplicity, I am not finding any usage.

Depending on how it is used, I also wondered about potential alternatives, such as using `.attributes` rather than this `.to_h`.

MrSerth (commented):
Here's some more code-level feedback. While some aspects might be fairly easy to change, others could be more tricky. Please consider the time required and potentially create a follow-up ticket instead.

MrSerth:
Ah, one further thing I wanted to share: About two weeks ago, online ratings were subject of a trial at the German Federal Court of Justice (BGH). The question was whether an online platform needs to show the distribution of ratings for the different star levels. In the ruling, the court decided that this information doesn't need to be revealed, since the average user should be aware of the potential discrepancy of individual ratings. 

Sources: [Press Release](https://www.bundesgerichtshof.de/SharedDocs/Pressemitteilungen/DE/2024/2024154.html?nn=10690868), [Ruling](https://juris.bundesgerichtshof.de/cgi-bin/rechtsprechung/document.py?Gericht=bgh&Art=en&nr=138503&anz=19&pos=1)

app/views/tasks/_tasks.html.slim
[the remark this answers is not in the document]
Mathis-Z: Right, I was wondering if that was intended because there was a CSS class aligning it to the bottom that I did not add. I agree that it looks better centered.

app/assets/javascripts/tasks_ratings.coffee
```diff
@@ -2,48 +2,42 @@ ready = ->
...
-    rating = this.getAttribute("data-rating")
-    lower = $('.rating span').filter (->
-      $(this).attr("data-rating") < rating
-    )
-    $(lower).removeClass("fa-regular").addClass("fa-solid")
-    upper = $('.rating span').filter (->
-      $(this).attr("data-rating") > rating
-    )
-    $(upper).removeClass("fa-solid").addClass("fa-regular")
+  $(".task-star-rating[data-is-rating-input='true'] .rating-star").hover(->
```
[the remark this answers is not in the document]
Mathis-Z: I actually liked the hover mechanism because it is faster but I can see why this could also be annoying. I switched to onclick :+1:

app/controllers/tasks_controller.rb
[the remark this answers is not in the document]
Mathis-Z: I just disabled the button for incomplete ratings. This should fix the problem, right?

app/models/rating.rb
Mathis-Z: I think that is better (also because it reduces the amount of code that needs to be changed if these categories ever change).

config/shakapacker.yml
Mathis-Z: Sorry, I missed that.

app/models/task.rb
Mathis-Z: Since the scope does not seem to be used anywhere else I merged it with min_stars. I still added a test though.

app/models/rating.rb
Mathis-Z: Yes, that is not needed anymore. I removed it :+1:

app/assets/javascripts/tasks_ratings.coffee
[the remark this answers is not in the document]
MrSerth: Thanks for updating; I think clicking is better to confirm a choice. Still, I liked the hover *effect* to "preview" what my click would do. Can we re-add that in a way that the left three stars are dark gray if I hover over the third and get all white again if I move out of the rating area without clicking?

app/controllers/tasks_controller.rb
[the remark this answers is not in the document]
MrSerth: Yes, looks good. I've tested with various edge cases and think this is fine :+1:.

MrSerth: Yeah, that's true. Let's encourage users to add all components here, too, and keep the current implementation 😉.

app/views/tasks/show.html.slim
MrSerth: Thanks for extracting the modal as requested, but I meant something different (even if that's not too clear when reading my own comment) 🙈.

Let's keep the extracted partial you've created, it reduces the overall length of the `tasks/show.html.slim` file.

What I meant was to have a *generic* partial for any modal, that could be instantiated multiple times. For example, imagine there is a file called `app/views/shared/_modal.html.slim` which accepts multiple parameters. Then, you could simply do something like:

```slim
= render('shared/modal', id: 'ratingModal', title: Rating.model_name.human, template: 'tasks/_rating_modal')
```

The partial `app/views/shared/_modal.html.slim` would contain everything that is generally needed for all modals, such as setting the `zindex`, Bootstrap classes, etc. And in `app/views/tasks/_rating_modal.html.slim`, you would simply add the *content* of the modal only and potentially content for the footer (if any). This way, maintaining changes to Bootstrap and the views is easier to implement. The example I linked from CodeOcean is this generic file, with [this one](https://github.com/openHPI/codeocean/blob/092e8426671324db7489378c1f0d17b344ee58b5/app/views/programming_groups/_info_pair_programming.html.slim) being an example for a specific template (modal content).

Do you think my proposal makes sense?

app/assets/javascripts/tasks_ratings.coffee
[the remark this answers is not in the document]
Mathis-Z: I added a hover effect. Not sure if that is exactly what you had in mind but I think it looks good.

app/views/tasks/show.html.slim
Mathis-Z: Yes, that makes sense but I implemented it a bit different to CodeOcean to avoid needing a new partial for every different modal body. What do you think?

app/assets/javascripts/tasks_ratings.coffee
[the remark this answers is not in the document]
MrSerth: Yes, that's what I intended and indeed, it looks great!

There is just a minor visual issue I noticed: Hovering over the last star somewhat changes the spacing of all (other) items. Since I cannot explain it with words, here's a screencast:

https://github.com/user-attachments/assets/527b3408-86ab-44db-9e47-53ef06830298

app/views/tasks/_rating_modal.html.slim
```diff
@@ -1,27 +1,17 @@
...
-            = render('task_star_rating', rating: @user_rating.overall_rating, is_input: true, category: :overall_rating)
-      .modal-footer.pt-0.border-0
-        - if @user_rating.valid?
-          button.btn.btn-light#ratingModalSaveButton data-bs-dismiss="modal"
-            = t('common.button.save')
-        - else
-          button.btn.btn-light#ratingModalSaveButton[data-bs-dismiss="modal" disabled]
-            = t('common.button.save')
+= render 'shared/modal', title: Rating.model_name.human, root_attributes: {id: 'ratingModal', 'data-task-id': @task.id}, size: 'modal-sm', footer: false do
+  .modal-body
```
MrSerth, line 2: I would prefer to move this `div` over to `app/views/shared/_modal.html.slim`

app/views/tasks/_rating_modal.html.slim
```diff
@@ -1,27 +1,17 @@
...
-          span.task-rating
-            = render('task_star_rating', rating: @user_rating.overall_rating, is_input: true, category: :overall_rating)
-      .modal-footer.pt-0.border-0
-        - if @user_rating.valid?
-          button.btn.btn-light#ratingModalSaveButton data-bs-dismiss="modal"
-            = t('common.button.save')
-        - else
-          button.btn.btn-light#ratingModalSaveButton[data-bs-dismiss="modal" disabled]
-            = t('common.button.save')
+= render 'shared/modal', title: Rating.model_name.human, root_attributes: {id: 'ratingModal', 'data-task-id': @task.id}, size: 'modal-sm', footer: false do
```
MrSerth, line 1: The `footer` variable is not used, is it?

app/views/tasks/_rating_modal.html.slim
```diff
@@ -1,27 +1,17 @@
...
+          = render('task_star_rating', rating: @user_rating.send(category), is_input: true, category:)
+    hr
+    span.rating-category-container
+      span
+        = t('activerecord.attributes.rating.overall_rating')
+      span.task-rating
+        = render('task_star_rating', rating: @user_rating.overall_rating, is_input: true, category: :overall_rating)
+
+    button.btn.btn-light.float-end.mt-3#ratingModalSaveButton data-bs-dismiss="modal" disabled=@user_rating.invalid?
+      = t('common.button.save')
```
MrSerth, line 17: Let's move this part to the footer, I'd say (we don't need a close button anyway).

app/views/shared/_modal.html.slim
```diff
@@ -0,0 +1,20 @@
...
+      - if header
+        .modal-header
+          h6.modal-title = title
+          button.btn-close data-bs-dismiss="modal" aria-label="Close"
+
+      = yield
+
+      - if footer_button_text.blank?
+        .modal-footer
+          button.btn.btn-light data-bs-dismiss="modal" = footer_button_text
```
MrSerth, line 20: I would tend to keep this more generic than just displaying a button to dismiss the modal (i.e., with the option to move arbitrary tags to the footer).

app/views/shared/_modal.html.slim
```diff
@@ -0,0 +1,20 @@
+- root_attributes ||= {}
+- root_attributes[:id] ||= id
```
MrSerth, line 2: If neither `root_attributes` nor `id` is passed, this would fail, wouldn't it?

app/views/tasks/show.html.slim
MrSerth: Yes, that's fine for me. I've made a few code suggestions there. Also, it currently feels a bit weird to have an empty button in the footer -- let's simply move the "Save" button there.

Bildschirmfoto 2024-08-17 um 18 38 14

app/views/tasks/_rating_modal.html.slim
Mathis-Z: I know but if we do that we cannot use the modal template with a custom modal-header or modal-footer because everything gets wrapped in the modal-body. Ideally we could pass 3 blocks to the modal partial - one for header, body and footer - but ruby allows only one block. I could not get it to work like that so I went with this instead to allow custom headers/footers.

app/views/shared/_modal.html.slim
Mathis-Z: As I wrote above I am not sure how to do this (how to use multiple blocks?).

app/views/shared/_modal.html.slim
Mathis-Z: Yes. I will just remove `root_attributes[:id] ||= id` because it is not really needed anyway.

app/views/tasks/show.html.slim
Mathis-Z: Sorry, I somehow missed that...

app/assets/javascripts/tasks_ratings.coffee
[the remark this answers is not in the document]
Mathis-Z: This seems to be caused by a size difference between the filled stars and the empty stars. I changed some CSS and it works (on my machine) now.

app/views/tasks/_rating_modal.html.slim
Mathis-Z: Okay, so I used `content_for` now to achieve something like that. I think this is much better. There is a problem with content_for though: If you define the same `content_for` twice the default behavior is to append the contents which is not what we want if we define multiple modals on one page. There is a flush option but still: If one modal defines a custom footer through content_for and another modal wants to use the default footer this might conflict (if we use `content_for?` to check if we want to include a custom footer) so I had to add two additional flags `custom_header` and `custom_footer` to allow the second modal to ignore the defined custom footer of the first modal.

app/views/tasks/_tasks.html.slim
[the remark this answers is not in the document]
MrSerth: This issue is back, unfortunately. I've tested with three different lines. In the following screenshot, lines of the same color have the same length:

Bildschirmfoto 2024-08-26 um 18 42 32

The green one is centered, but neither yellow nor red is.

app/assets/javascripts/tasks_ratings.coffee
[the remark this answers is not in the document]
MrSerth: Yes, I can confirm it works on my machine, too 🎉. Thank you!

app/views/tasks/_tasks.html.slim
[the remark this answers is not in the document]
MrSerth: Maybe @zuhrasofyan can help out here?

app/views/tasks/show.html.slim
MrSerth: Don't worry, thanks for taking care. I can confirm the button is gone 😉

app/models/task.rb
MrSerth: If I see this right, the merged scope `min_stars` is tested in `spec/models/task_spec.rb`. Hence, this conversation can be resolved, I assume.

app/views/tasks/_tasks.html.slim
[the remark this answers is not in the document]
Mathis-Z: I already pushed a fix.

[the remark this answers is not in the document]
MrSerth: Perfect! We may just increase the spacing between the stars and the number of ratings, and than this is perfectly fine 👌 

Bildschirmfoto 2024-08-27 um 12 40 34

MrSerth:
We just merged this branch and included all changes for the rating with fbbf0e9f.

app/views/tasks/_tasks.html.slim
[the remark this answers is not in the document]
MrSerth: Done.

app/views/tasks/_rating_modal.html.slim
MrSerth: We included the `clear_content_for` method and cleaned the modal template a little before merging.

Issue #1566, opened 2024-08-05
Potential (minor) security issue when exporting using account_links


kkoehn:
https://github.com/openHPI/codeharbor/blob/458b7f88c1aef6c09f745c5b705254c9c823ef23/app/controllers/tasks_controller.rb#L193

This looks like you can potential export the task to any `account_link`, since you just have to supply the `id`

Todo:
- [x] check
- [x] fix

MrSerth:
Indeed, that is a problem - I verified it manually. To fix the issue, we probably need to call `authorize @account_link` in all relevant places. A first check revealed the following:

- [ ] `CollectsionsController#push_collection` https://github.com/openHPI/codeharbor/blob/458b7f88c1aef6c09f745c5b705254c9c823ef23/app/controllers/collections_controller.rb#L73
- [ ] `TasksController#export_external_start` https://github.com/openHPI/codeharbor/blob/458b7f88c1aef6c09f745c5b705254c9c823ef23/app/controllers/tasks_controller.rb#L166
- [ ] `TasksController#export_external_check` https://github.com/openHPI/codeharbor/blob/458b7f88c1aef6c09f745c5b705254c9c823ef23/app/controllers/tasks_controller.rb#L175
- [ ] `TasksController#export_external_confirm` https://github.com/openHPI/codeharbor/blob/458b7f88c1aef6c09f745c5b705254c9c823ef23/app/controllers/tasks_controller.rb#L193

To prevent any consequence of this bug for now, I took immediate action and disabled the export to any account link for now.

PR #1577, opened 2024-08-09
Fix account_link security issue

kkoehn:
Fixes a security issue enabling access to any account_link given the id. I only fixed the occurrences in the `task_controller` since the `collections_controller` would need a spec as well and is not in use at the moment. The collection-export should be fixed and tested in another PR.

fixes: #1566

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -4,6 +4,7 @@ require 'zip'
 
 class TasksController < ApplicationController # rubocop:disable Metrics/ClassLength
   before_action :load_and_authorize_task, except: %i[index new create import_start import_confirm import_uuid_check import_external]
+  before_action :load_and_authorize_account_link, only: %i[export_external_start export_external_check export_external_confirm]
   before_action :only_authorize_action, only: %i[import_start import_confirm import_uuid_check import_external]
 
   before_action :handle_search_params, only: :index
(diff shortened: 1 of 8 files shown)


app/controllers/tasks_controller.rb
```diff
@@ -231,6 +226,11 @@ def load_and_authorize_task
     authorize @task
   end
 
+  def load_and_authorize_account_link
+    @account_link = AccountLink.find(params[:account_link])
+    authorize @account_link, :update?
```
MrSerth, line 231: Why are you checking for `:update?` permissions here? It seems to be semantically incorrect and also prevents the use of shared account links.

Note: I think we cannot share an account link currently with any further users, but we still have the data model and existing account links present and can use them.

Would it make sense to add another verb for the usage, so that we do not overload `:show?`? Or shall we simply keep using it?

spec/controllers/tasks_controller_spec.rb
```diff
@@ -921,6 +921,15 @@
       get_request
       expect(response).to render_template('export_external_start')
     end
+
+    context 'when account_link is not owned by user' do
```
MrSerth, line 925: I would prefer to go with a human description here and in the following:

```suggestion
    context 'when the account link is neither owned by nor shared with the requesting user' do
```

spec/controllers/tasks_controller_spec.rb
MrSerth, line 1: I feel here are some tests missing for shared account links (which should still work).

MrSerth (changes requested):
Thanks for fixing the issue, please find my review attached.

Regarding the `CollectionsController`: I agree, the functionality is currently broken. Still, I would strongly suggest to fix the security vulnerability there too, based on the design you chose for the `TasksController` (before_action for `push_collection`, setting the `@account_link` variable, passing this to the `push_exercises` method, mocking the broken and incomplete `push_exercises` method in the specs).

My main reason to fix it there too (even if still not working) would be that the broken code is completely removed and a secure pattern is established. Otherwise, one might (at one day) implement the missing method and thereby "re-introduce" the vulnerability.

MrSerth:
By the way: What I really like about the refactoring you did is the refactoring within the ProFormA service. I like passing the `AccountLink` object rather than just the ID :+1:.

Dome-GER (approved):

Issue #1576, opened 2024-08-09
GptService::GenerateTests creates invalid test files


MrSerth:
The `GptService::GenerateTests` currently creates test files as follows:

https://github.com/openHPI/codeharbor/blob/992aeab7e085e286b48cf4d5710e467617af2fd3/app/services/gpt_service/generate_tests.rb#L16

However, the option `visible: false` is invalid, since `visible` is a string (no boolean) with fixed values (according to the ProFormA standard). When saved in the database, the boolean `false` is saved as `f`. 

When downloading such a task as ProFormA ZIP file, the following error is raised:

```ruby
ProformaXML::PostGenerateValidationError at /tasks/59/download
[#<Nokogiri::XML::SyntaxError: 281:0: ERROR: Element '{urn:proforma:v2.1}file', attribute 'visible': [facet 'enumeration'] The value 'f' is not an element of the set {'yes', 'no', 'delayed'}.>]
```

We should:

1. fix the error in the unit test generation
2. add a validation to the filed
3. test the behavior
4. (maybe) write a migration to resolve invalid values in our production database

Melhaya:
@MrSerth Thanks for identifying the issue.

I have added a validation in the `TaskFile` and adjusted the creation of one in the `GptService`. 

Currently a few tests are failing though. I am not sure how to tackle this. Of course due to the added validation in the `TaskFile`, these errors are raised.  I am just not sure how to fix them. Could you please advise? thanks.

MrSerth:
Fixed by #1588.

PR #1588, opened 2024-08-16
Fix test file creation in GPTService

Melhaya:
Related to https://github.com/openHPI/codeharbor/issues/1576 raised by @MrSerth 

Actions taken:

- [x] fix the error in the unit test generation
- [x] add a validation to the file
- [x] test the behavior

### DIFF
--- a/app/services/gpt_service/generate_tests.rb
+++ b/app/services/gpt_service/generate_tests.rb
@@ -13,7 +13,7 @@ module GptService
     def execute
       file_name = "test#{@task.programming_language.file_extension}"
 
-      test_file = TaskFile.new(content: gpt_response, name: file_name, used_by_grader: true, visible: 'yes', xml_id: SecureRandom.uuid)
+      test_file = TaskFile.new(content: gpt_response, name: file_name, used_by_grader: true, visible: 'no', xml_id: SecureRandom.uuid)
       test = Test.new(task: @task, title: I18n.t('tests.model.generated_test'), xml_id: SecureRandom.uuid, files: [test_file])
 
       @task.tests << test
--- a/app/services/gpt_service/generate_tests.rb
+++ b/app/services/gpt_service/generate_tests.rb
@@ -13,7 +13,7 @@ module GptService
     def execute
       file_name = "test#{@task.programming_language.file_extension}"
 
-      test_file = TaskFile.new(content: gpt_response, name: file_name, used_by_grader: true, visible: false, xml_id: SecureRandom.uuid)
+      test_file = TaskFile.new(content: gpt_response, name: file_name, used_by_grader: true, visible: 'yes', xml_id: SecureRandom.uuid)
       test = Test.new(task: @task, title: I18n.t('tests.model.generated_test'), xml_id: SecureRandom.uuid, files: [test_file])
 
       @task.tests << test
--- a/db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
+++ b/db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class FixUsedByGraderAndVisibleForTaskFiles < ActiveRecord::Migration[7.1]
+  class TaskFile < ApplicationRecord
+  end
+
+  disable_ddl_transaction!
+
+  def change
+    visibility_options = %w[yes no delayed]
+    default_visibility = {'Task' => 'yes', 'Test' => 'no', 'ModelSolution' => 'delayed'}
+
+    TaskFile.where.not(visible: visibility_options).or(TaskFile.where(used_by_grader: nil)).find_each do |task_file|
+      task_file.visible = default_visibility[task_file.fileable_type] unless visibility_options.include? task_file.fileable_type
+      task_file.used_by_grader = true if task_file.used_by_grader.nil?
+      task_file.save!(touch: false)
+    end
+  end
+end
--- a/spec/services/gpt_service/generate_tests_spec.rb
+++ b/spec/services/gpt_service/generate_tests_spec.rb
@@ -65,7 +65,9 @@ RSpec.describe GptService::GenerateTests do
         test_file = task.reload.tests.last.files.first
         expect(test_file).to have_attributes(
           content: "def test_script():\n  assert true",
-          name: 'test.py'
+          name: 'test.py',
+          visible: 'yes',
+          used_by_grader: true
         )
       end
 
--- a/spec/services/gpt_service/generate_tests_spec.rb
+++ b/spec/services/gpt_service/generate_tests_spec.rb
@@ -66,7 +66,7 @@ RSpec.describe GptService::GenerateTests do
         expect(test_file).to have_attributes(
           content: "def test_script():\n  assert true",
           name: 'test.py',
-          visible: 'yes',
+          visible: 'no',
           used_by_grader: true
         )
       end
--- a/app/models/task_file.rb
+++ b/app/models/task_file.rb
@@ -7,6 +7,8 @@ class TaskFile < ApplicationRecord
   validates :name, presence: true
   validates :attachment, presence: true, if: -> { use_attached_file == 'true' }, on: :force_validations
   validates :xml_id, presence: true
+  validates :visible, inclusion: {in: %w[yes no delayed]}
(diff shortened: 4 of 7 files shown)


MrSerth:
Since @kkoehn is an expert on the ProFormA standard and valid values, I am passing the review request over to him.

kkoehn:
I would not fix it in the `before_validation`, but in the factory for the tests. 

I noticed that `used_by_grader` is also required per ProFormA standard. It would make sense to implement a validation for that, too.

Currently, when creating a test-file through the UI, `visible` has the default-value `yes` (`used_by_grader` is not set by default).
I think we could extend the scope of this PR to also add a validation for `used_by_grader`, as well as the frontend modification to select a default value (should just be adding a `checked: true` in the correct line in `views/tasks/_file_config.html.slim`)

Concerning which default values to choose: There are no rules regarding that. I thought it makes sense, that a file added through the UI would be visible by default. I think it would make sense to extend that to your generated test files. By extension I would choose `true` for `used_by_grader` as default.

Melhaya:
Hey @kkoehn, Thanks for your comment. what do you mean by "I would not fix it in the `before_validation`, but in the factory for the tests." where is the factory for tests?

in GptService:

test_file = TaskFile.new(content: gpt_response, name: file_name, used_by_grader: true, visible: 'yes', xml_id: SecureRandom.uuid)
The default value for used_by_grader and visible are set correctly. What else should be implemented?

kkoehn:
- [ ] use factories instead of `before_validation`-hook
- [ ] add validation for `used_by_grader`
- [ ] add default value for `used_by_grader` in mentioned view-template

The factories are located in `spec/factories/`. There you are look for `task_files.rb`.

kkoehn:
Looks good, now you just need to a spec for the `visible`-validation. We use the gem `shoulda-matchers` to help with tests of validations. It should be pretty straight-forward. ([hint](https://github.com/thoughtbot/shoulda-matchers/blob/8890f3d182eebf2a6c8d5024afd656c2274a90c1/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb#L20))
You don't need to add a spec for the `used_by_grader`-validation ([see here](https://github.com/thoughtbot/shoulda-matchers/blob/8890f3d182eebf2a6c8d5024afd656c2274a90c1/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb#L59))

Melhaya:
Thanks for the hint and information regarding boolean values :)

I have added the spec for the `visible` value in the latest commit

app/services/gpt_service/generate_tests.rb
```diff
@@ -13,7 +13,7 @@ def initialize(task:, openai_api_key:)
     def execute
       file_name = "test#{@task.programming_language.file_extension}"
 
-      test_file = TaskFile.new(content: gpt_response, name: file_name, used_by_grader: true, visible: false, xml_id: SecureRandom.uuid)
+      test_file = TaskFile.new(content: gpt_response, name: file_name, used_by_grader: true, visible: 'yes', xml_id: SecureRandom.uuid)
```
MrSerth, line 16: Don't we want to hide test files as default?

spec/services/gpt_service/generate_tests_spec.rb
```diff
@@ -65,7 +65,9 @@
         test_file = task.reload.tests.last.files.first
         expect(test_file).to have_attributes(
           content: "def test_script():\n  assert true",
-          name: 'test.py'
+          name: 'test.py',
+          visible: 'yes',
```
MrSerth, line 69: As above, I would opt to hide a test by default.

MrSerth (commented):
Ideally, as written in the ticket, one would also add a migration for existing (production) data.

spec/services/gpt_service/generate_tests_spec.rb
Melhaya: Thanks. I have reverted to `no` and will push now.

Melhaya:
@MrSerth Thanks for the quick review :)

For the migration, I am honestly not sure how to do that by myself as it is a confusing area for me, and would need assistance to do it. If you would also prefer to do it to get it done faster, that's fine with me :)

kkoehn:
I added a migration, that should fix the existing TaskFiles

db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class FixUsedByGraderAndVisibleForTaskFiles < ActiveRecord::Migration[7.1]
+  class TaskFile < ApplicationRecord
+  end
+
+  def change
+    visible_options = %w[yes no delayed]
+    TaskFile.where.not(visible: visible_options).or(TaskFile.where(used_by_grader: nil)).find_each do |task_file|
+      task_file.visible = {'Task' => 'yes', 'Test' => 'no', 'ModelSolution' => 'delayed'}[task_file.fileable_type] unless visible_options.include? task_file.file_type
```
MrSerth, line 10: I don't get the condition with `task_file.file_type` here, and I cannot execute similar code locally:

```
# => TaskFile.last.file_type
NoMethodError: undefined method `file_type' for an instance of TaskFile
```

MrSerth, line 10: Should we maybe extract the has to a dedicated variable, e.g., `default_visibility` to reduce the file length (and make it easier to understand)?

db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
```diff
@@ -0,0 +1,15 @@
...
+      task_file.used_by_grader = true if task_file.used_by_grader.nil?
+      task_file.save(touch: false)
```
MrSerth, line 12: An invalid record, if any, will silently return `false`. We won't notice that at all.

db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
MrSerth, line 1: Since you are not changing the schema, you could add `disable_ddl_transaction!` (in a dedicated line before `def change`, followed by an empty line for better readability).

MrSerth (commented):
Thanks Karol for taking the time to add a migration! 💪 A few minor remarks and then I think this is good to go.

db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
kkoehn: Do you mean line length of file length?

db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
kkoehn: What does it do? I looked it up and it seems to only deactivate the automatic transaction for the migration. I don't see the benefit or the connection to not changing the schema.

db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
MrSerth: I was referring to the line length. Hence, extracting it (as you did) is great -- thanks!

db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
MrSerth: A migration might block further access to the table, and thus can be blocking (for other requests). Since we don't stop our Rails server during deployments (but actively use a phased restart for the web server nodes), this would slow down the application during deployments. Hence, it is recommended to perform data backfills or migrations outside of the default transaction, preventing this locking mechanism.

It is one of the techniques I learned from the `strong migrations" guide, see a better example [in their docs](https://github.com/ankane/strong_migrations/blob/master/README.md#good-7).

Honestly, I might not be required in the current case (since no change to the table structure is being performed), but I just got used to it in all migrations mingling data.

MrSerth (approved):
Thanks for fixing the bug and adding the additional validators! I've just tested the PR locally and think it's ready to go 🚀

db/migrate/20240820192904_fix_used_by_grader_and_visible_for_task_files.rb
kkoehn: :+1: thanks for the explanation

PR #1589, opened 2024-08-16
Upgrade to Rails 7.2, switch to system specs, and add PWA and devcontainers support

MrSerth:
With this PR, we are upgrading CodeHarbor to Rails 7.2 and making use of all the great features the new version provides. Therefore, this PR looks rather big at a first glance, but the commits are intended to provide a meaningful grouping of the changes. All commits are required to complete the upgrade and the activation of current best-practices for Rails application.

From a high-level point of view, the following changes are relevant:
- CodeHarbor is running Rails 7.2 (instead of Rails 7.1)
- Remove deprecated `secrets.yml`, require setting environment variable `SECRET_KEY_BASE` for production
- Rspec feature specs are migrated to system specs, which allows the usage of advanced features (automatic screenshots in case of failures, easier management of Capybara, no need to use the DatabaseCleaner gem any longer)
- Applying best-practices for Rails and Rspec, reordering some files
- Adding a webmanifest for CodeHarbor, which allows to install CodeHarbor as a PWA
- Chrome is used as a new default browser for system specs (rather than Firefox)
- A production-ready Dockerfile and a pre-configured devcontainer for local development are provided
- GitHub Codespaces is officially supported

<details><summary>Screenshots of enhanced PWA support</summary>
<p>

**Shortcuts with native app experience:**
Bildschirmfoto 2024-08-16 um 14 17 42

</p>
</details> 

<details><summary>Screenshots of GitHub Codespaces running CodeHarbor</summary>
<p>

**Codespaces Editor:**
Bildschirmfoto 2024-08-16 um 14 03 46
**Remote Access:**
Bildschirmfoto 2024-08-16 um 14 03 51

</p>
</details> 

<details><summary>List of commits</summary>
<p>

Specifically, this PR adds the following commits:

- Upgrade to Rails 7.2 and apply new framework defaults
- Remove outdated reference to node_modules folder
- Add brakeman with GitHub action pipeline
- Add production-ready Docker image
- Add webmanifest and basic PWA support
- Apply current config for Rspec and move extensions to support directory
- Move ungrouped specs to a dedicated folder `other`
- Fix warning about stubbing a method on `nil`
- Ensure select2 is initialized regardless of page load events
- Migrate feature specs to system specs
- Switch to Rspec native transactional fixtures to replace database cleaner
- Add support for devcontainers
- Add support for GitHub Codespaces and port forwarding
- Add recent changes to Local Setup Guides

</p>
</details>


spec/system/authentication_system_spec.rb
```diff
@@ -26,8 +26,8 @@
     context 'with invalid credentials' do
       it 'does not allow to sign in' do
         click_link(I18n.t('common.button.log_in')) # rubocop:disable Capybara/ClickLinkOrButtonStyle
-        fill_in('Email', with: user.email)
-        fill_in('Password', with: password.reverse)
+        fill_in(:user_email, with: user.email)
+        fill_in(:user_password, with: password.reverse)
```
Dome-GER, line 30: I don't know the details here tbh, but usually I prefer using real labels. This does not seem to be possible anymore.

spec/system/tasks_system_spec.rb
```diff
@@ -87,7 +88,9 @@ def displayed_task_rating
         find_by_id('ratingModalOpenButton').click
 
         Rating::CATEGORIES.each do |category|
-          first(".task-star-rating[data-rating-category=#{category}][data-is-rating-input=true] .rating-star[data-rating='#{new_rating[category]}']", minimum: 0, wait: false)&.click
+          next unless new_rating.fetch(category, 0).positive?
+
+          first(".task-star-rating[data-rating-category=#{category}][data-is-rating-input=true] .rating-star[data-rating='#{new_rating[category]}']")&.click
```
Dome-GER, line 93: Safe navigation may not be needed anymore? (...and tbh feels a bit weird in specs)

Dome-GER (approved):

spec/system/authentication_system_spec.rb
MrSerth: We could still use a real label, for sure. However, I simply went for consistency: In other tests, we used the symbol already (which is not a `I18n.t()` label, but the ID of the HTML element.

We could change it, for sure, but then we should do so globally. For now, I would rather have it uniformed with the other occurrences.

spec/system/tasks_system_spec.rb
MrSerth: I was checking this code explicitly during the upgrade. It is something I overlooked when the feature was merged, but I considered a more detail change out of scope for the Rails upgrade (this PR).

The issue is that the subject (which this snippet is a part of) is very general and also tests invalid input. Specifically, it tries to click the "zeroth" out of 5 stars for some selection. As you can imagine, it is not possible to click that (but rather the first, second, ... fifth star). After some initial try and error, I opted to limit the scope of this PR and keep the current implementation.

PR #1592, opened 2024-08-20
Show labels on the task index page

zuhrasofyan:
Show labels on the task index page, and make it vertically centered and flexible to not disturb other elements in the footer


kkoehn (commented):
I'm not sure if this needs fixing, but I noticed, that the labels suppress the existing buttons, when there are too many
image

zuhrasofyan:
> I'm not sure if this needs fixing, but I noticed, that the labels suppress the existing buttons, when there are too many image

Tried to use columns in addition of d-flex to adjust the labels and button

kkoehn:
looks good now! Thank you.
:+1:

kkoehn (approved):

PR #1606, opened 2024-08-25
Refactor deletion of dangling labels

MrSerth:
The previous mechanism required newly-assigned TaskLabels to be saved before aiming to destroy dangling labels. However, there is one edge-case previously used in the tests which caused erroneous behavior:

Imagine, there is a dangling label (with no matching TaskLabel). Then, a task is assigned that dangling label through the `label_names=` mechanism. This will cause the `Label.find_or_initialize_by(name:)` method to return the dangling label and assign that to the task. However, when this assignment is not saved it (and no `TaskLabel` record is created), the following call to destroy dangling labels also removed the one just assigned to a task.

This caused errors in tests, when switching from `labels` to `label_names` in the TaskController specs to remove the use of unpermitted parameters and transactional fixtures.

The new mechanism will still remove dangling labels, but only those associated with a TaskLabel just removed. Through the transactional setting of destroy hooks, this still ensures there aren't any leftovers.

### DIFF
--- a/app/models/label.rb
+++ b/app/models/label.rb
@@ -72,6 +72,10 @@ class Label < ApplicationRecord
     }.compact
   end
 
+  def to_s
+    name
+  end
+
   def self.ransackable_attributes(_auth_object = nil)
     %w[name]
   end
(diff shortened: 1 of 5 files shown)


Mathis-Z (approved):
Looks good, I am just wondering if we need `Label#to_s`? I could not find any usage and it is also not covered by the tests.

Dome-GER (approved):

PR #1612, opened 2024-08-26
refactor import to reuse task_files based on xml_id

kkoehn:


app/services/proforma_service/convert_proforma_task_to_task.rb
```diff
@@ -52,15 +60,24 @@ def files
     end
 
     def file_from_proforma_file(proforma_task_file)
-      task_file = TaskFile.new(file_attributes(proforma_task_file))
+      task_file = (@task.id && TaskFile.all_files_for_task(@task).find_by(xml_id: proforma_task_file.id)) || TaskFile.new
```
MrSerth, line 63: Wouldn't it be more efficient to query all files once and then cache the result?

What do you think about the following implementation?

```rb
  # models/task_file.rb

  validate :unique_xml_id, if: -> { !fileable.nil? && xml_id_changed? }

  def unique_xml_id
    task = fileable.is_a?(Task) ? fileable : fileable.task
    xml_ids = (task.all_files(cached: false) - [self]).map(&:xml_id)
    errors.add(:xml_id, :not_unique) if xml_ids.include? xml_id
  end
```

```rb
  # models/task.rb
  def all_files(cached: true)
    return @all_files if cached && defined?(@all_files)

    @all_files = (files + tests.map(&:files) + model_solutions.map(&:files)).flatten
  end
```

```suggestion
      task_file = @task&.all_files&.select{|file| file.xml_id == proforma_task_file.id}&.first || TaskFile.new
```

Then, the new scope wouldn't be needed any longer and the tests you've written could be moved to `Task#all_files` (where we don't have any tests yet).

MrSerth (commented):
I've tested the changes locally and think everything works as expected :+1:

There is just a comment left regarding the implementation.

app/services/proforma_service/convert_proforma_task_to_task.rb
kkoehn: that seems easier :+1:

MrSerth (approved):
Very nice, thank you! 🦚

MrSerth (changes requested):
Sorry to add another issue to this PR, but I noticed an actual issue when updating changed tasks (through CodeOcean). Consider the following task in the "old" form, updated with the "new" version. (Hint: The bug is also visible through the ZIP upload)

[task_845_new.zip](https://github.com/user-attachments/files/16856736/task_845_new.zip)
[task_845_old.zip](https://github.com/user-attachments/files/16856735/task_845_old.zip)

The new task should contain a file for the tests, but after the update it doesn't. The test is present but without any files. Only a second update with the same "new" version "restores" (actually: recreates) the test file.

Reverting back to the old version yields a similar problem, any might optionally display an error.

kkoehn:
We should also double check that files get deleted, when they are not present anymore

kkoehn:
@MrSerth I can't seem to get a handle on the problem.
The assignment of the files is correct, but during the save some of them end up with `fileable_id` nil and get destroyed because of `dependent: destroy` the `fileable_type` seems to be set correctly

MrSerth:
> @MrSerth I can't seem to get a handle on the problem. The assignment of the files is correct, but during the save some of them end up with `fileable_id` nil and get destroyed because of `dependent: destroy` the `fileable_type` seems to be set correctly.
> 
> Do you have an idea?

Barely 😵‍💫. I played around with Rails for many hours and was trying many different things, until finding a suitable (not necessarily nice) solution. Your previous tests were super helpful, thanks again for this true test-driven approach!

I am not really satisfied with the approach, and I find overwriting `files` not ideal. However, I was really unable to find another, better solution. Fixing the "save" action was relatively easy and straight forward, but broke the modified (not saved) version. Fixing both was super hard, and the only solution that worked and did not overwrite much of the internals is the one I present. From my point of view, I find the suggestion okay-ish and would proceed with that base -- all changes are based on the public API and thus should be stable.

See my PR #1657 for details and the in-depth suggestion.

kkoehn:
@MrSerth I merged your suggestions, thanks again for the help. I think this should be finished now.

MrSerth (approved):
I think so too, and would consider this PR to be ready. Updating a task with moved files (as mentioned earlier in https://github.com/openHPI/codeharbor/pull/1612#pullrequestreview-2278555662) is now working correctly 🥳

PR #1628, opened 2024-09-01
Implement missing features for Task Contributions

MrSerth:
This PR adds the missing functionality as discovered during review / pair programming to PR #1103.

@florian-str I've tested the behavior manually but would like that you add some suitable test cases for the changes I made. Also, please take a moment to carefully read through my changes and check whether I made any mistake or whether something is left unclear.

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -138,9 +138,9 @@ class Task < ApplicationRecord # rubocop:disable Metrics/ClassLength
   end
 
   def apply_contribution(contrib)
-    transfer_attributes(contrib.suggestion, %w[id parent_uuid access_level user_id uuid created_at])
-    transfer_multiple_entities(model_solutions, contrib.suggestion.model_solutions, 'model_solution')
-    transfer_multiple_entities(tests, contrib.suggestion.tests, 'test')
+    transfer_attributes(contrib.suggestion)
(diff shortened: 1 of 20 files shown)


app/models/task.rb
```diff
@@ -77,7 +77,9 @@ class Task < ApplicationRecord # rubocop:disable Metrics/ClassLength
     where.missing(:task_contribution)
   }
   scope :pending_contribution, lambda {|user|
-    joins(:task_contribution).where(user:, task_contribution: {status: :pending})
+    # Since accessing any contribution that belongs to a non-public task doesn't work, we filter for the parent visibility.
+    # TODO: Do we want to keep this behavior?
```
florian-str, line 81: I'm not sure. There are good arguments for both keeping the current implementation and changing it.

Pros of keeping the current implementation: If the original task's visibility is changed to private accidentally, restoring its visibility would also restore access to the contribution.

Cons: The contributions could reappear after a significant time period, making them less useful.

florian-str (approved):
Thank you for your input.

app/models/task.rb
MrSerth: That's true, thanks for your input. As discussed, I created #1633 as a follow-up issue.

PR #1632, opened 2024-09-02
Aggregate all contribution-related policies

MrSerth:
So far, the `TaskPolicy` was called for real tasks (non-contributions) and for suggestions of a TaskContribution. With this PR, we change said behavior to always use the `TaskContribution` if checking for a suggestion or an actual contribution. We still mingle with TaskContributions and Tasks in said policy, but since both models are much closer than the Task policies are for real tasks and suggestions, this is probably better.

I still want to look into reducing the `tasks/show.html.slim` with this new approach, so that this PR is still pending.


MrSerth:
@florian-str I finished my refactoring; please have a look 🙂

app/policies/task_contribution_policy.rb
```diff
@@ -19,37 +24,64 @@ def base
     end
   end
 
-  def discard_changes?
-    return no_one unless task_contribution.status_pending?
+  def index?(base: nil)
```
florian-str, line 27: This function does not yet work as expected (or match our original description of the permissions).

florian-str (approved):
Thank you for your input. It did significantly improve the readability and maintainability of the code. 

While the Policy `index?` method does not yet work, I'll still merge the changes and correct the issues locally as part of #1103.

app/policies/task_contribution_policy.rb
MrSerth: What do you think is missing? I am truly interested, since I already spent quite some time in implementing the `index?` method. Thanks for discovering a potential issue and raising my awareness!

The issues I faced during refactoring was regarding the different use cases for this method. Of course, it is used for the actual "View contributions" button (and corresponding controller action) of TaskContributions. Despite that, it is also used within the bread crumbs navigation and in some tests. All three have a slightly different context and I aimed to make the method as robust as possible.

Here's an annotated version:

```rb
  def index?(base: nil)
    # If we are checking for the class (not a specific record), we cannot know which contribution we're looking at or which task it is created for.
    # This mostly happens during the breadcrumbs and title partial, where we are checking for general permissions.
    return everyone if @record == TaskContribution

    # If no base model is passed as an argument, we try to find the corresponding one.
     if base.blank?
       # Since index? might be called on an array, we use the last element if possible. This matches the general behavior of Pundit.
       contribution = @record.try(:last) || @record
       # For the given contribution, we aim to fetch the base task.
       base = contribution.try(:base)
       # If no base task is found, we deny access
       return no_one unless base
     end

     # Only those users who can edit the base task can view the index page.
     Pundit.policy(@user, base).edit?
   end
```

florian-str: I'm sorry, I was checking this part manually in the browser and wrongly assumed that the permissions were flawed.

It turns out the error was not located in the policy, but rather in the controller (you were checking the permissions but not doing anything with the result).

The only thing where I'm unsure is whether `return everyone if @record == TaskContribution` is really necessary, or if it should rather be replaced by `no_one` for reducing the chance of an accidental granting of permissions for everyone.

MrSerth: Ah, thanks for explaining. That‘s true, I am missing a necessary check in the controller:
https://github.com/openHPI/codeharbor/blob/5291ef93142e567879a457b93c6dca39c9ce2f51/app/controllers/task_contributions_controller.rb#L39

This line should be:

```rb
raise pundit::NotAuthorizedError unless policy(@task_contributions).index? base: @task
```

The issue is that the `authorize` method doesn’t allow passing a custom argument, which we need here.

Regarding your other question:

```rb
return everyone if @record == TaskContribution
```

I was shortly considering `no_one` here, too. However, the breadcrumbs are only intended when visiting a specific contribution. They are not shown right now, but would as

```
Tasks > Base Task > Task Contributions > Suggestion
```

There, the policy checks whether the ”Task Contributions“ is clickable or not. Here is a similar example from CodeOcean:

IMG_3054

For this reason, I decided to go with `everyone` here.

MrSerth: I've just double-checked my suggestion

```rb
raise Pundit::NotAuthorizedError unless policy(@task_contributions).index? base: @task
```

and have run the tests locally. With this suggestion applied, I got a few failures for `spec/controllers/task_contributions_controller_spec.rb:16`. If I see this right, the failures are legit, since the contributing author (not the `original_author`) is accessing the index page. According to our discussion and the policy, this is not allowed, rendering the errors. Ensuring that the `original_author` is signed in lets the tests pass.

PR #1650, opened 2024-09-08
Update Ruby version and Bundle

MrSerth:
This PR updates to the latest Ruby version and bumps all dependencies.

Fixes CVE-2024-45296
Fixes GHSA-9wv6-86v2-598j
Fixes CVE-2024-47220
Fixes GHSA-6f62-3596-g6w7


.ruby-version
```diff
@@ -1 +1 @@
-3.3.4
+ruby-3.3.5
```
Dome-GER, line 1: Why did you decide to go for "ruby-"? `ruby-3.3.5` and `3.3.5` are semantically equivalent, aren't they?

Dome-GER (approved):

.ruby-version
MrSerth: > `ruby-3.3.5` and `3.3.5` are semantically equivalent, aren't they?

Yes, they are. 

> Why did you decide to go for "ruby-"? 

Since Rails has version `ruby-3.3.5` as a default, I opted to change to this syntax, too: https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/ruby-version.tt

PR #1652, opened 2024-09-10
Migrate to Solid Queue 0.8.x and extract tables

MrSerth:
With Solid Queue 0.8.0, support for dedicated  databases were introduced. In fact, Rails recommends moving the database, which we perform with the migration.

Unfortunately, it is currently unclear how exactly schema migrations will be provided. Therefore, we cannot use the same GitHub actions workflow for now and will re-add it once a new migration has been published.


Dome-GER (approved):
The `db:prepare` change could have received a note in the PR description. ;) Thanks for the verbal explanation today (the first case switching to a setup with multiple DBs where DB prepare is the suitable command, loading the schema or seeding the DB and your change with not executing seeds if users exist).

PR #1657, opened 2024-09-11
Allow moving ProFormA files between fileables

MrSerth:
We recently discovered an edge case where a moved file was not updated correctly. At first, it got removed and recreated, and later on only removed. While these changes look pretty difficult, they are the best and only working approach I could find to fix the problem ;). For now, I'd suggest to go with this solution. See #1606 for more details.


kkoehn:
this looks good :+1: 
thanks for the help!

kkoehn (approved):

PR #1664, opened 2024-09-17
Hide comment box part by default

zuhrasofyan:
Fix the bug that the comment box in task not hidden by default


MrSerth (approved):
Nice, thanks! I was able to confirm that the PR fixes the issue I reported 👍

Issue #1066, opened 2023-11-01
Task search breaks with Created parameter 'Today'


Mathis-Z:
When you use the advanced task search parameters to filter tasks created 'Today' the server returns an internal server error.

Mathis-Z:
Fixed by #1667

PR #1667, opened 2024-09-18
Fixing multiple issues with the advanced search options

Mathis-Z:
Closes tickets #1485 and #1066.
Fixes an untracked bug when sorting tasks after their rating.


MrSerth (approved):
Awesome, thanks for fixing these three bugs and covering the functionality with tests 💪 As expected, my local testing also confirmed that these advanced search options are now working as expected.

PR #1675, opened 2024-09-22
Rejecting or revoking a collection invite destroys the invitation

### LINKED ISSUE #1414: Share collection: Re-invite member who left previously
If a user was a member of a collection and left voluntarily, one cannot invite the same member again. This is likely due to some validation with for the `Messages` model.

Bildschirmfoto 2024-05-26 um 11 38 29

### LINKED ISSUE #1668: Cannot send second collection invite
Steps to reproduce:
- send a collection invitation
- delete the invitation message
- send a second collection invitation to the same user for the same collection
=> flash message: "Something went wrong"

Mathis-Z:
Closes #1414. Closes #1668.

I also added a nicer flash message when a user has already been invited.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -91,11 +91,14 @@ class CollectionsController < ApplicationController
   end
 
   def share
-    if @collection.users.exclude?(share_message.recipient) && share_message.save
-      redirect_to collection_path(@collection), notice: t('.success_notice')
-    else
-      redirect_to collection_path(@collection), alert: t('common.errors.something_went_wrong')
-    end
+    flash_message = if @collection.users.include?(share_message.recipient) || Message.exists?(share_message_args)
+                      {alert: t('.duplicate_share')}
+                    elsif share_message.save
+                      {notice: t('.success_notice')}
+                    else
+                      {alert: t('common.errors.something_went_wrong')}
+                    end
+    redirect_to collection_path(@collection), flash_message
   end
 
   def view_shared
@@ -150,12 +153,11 @@ class CollectionsController < ApplicationController
   end
 
   def share_message
-    @share_message ||= Message.new(
-      sender: current_user,
-      recipient: User.find_by(email: params[:user]),
-      param_type: 'collection',
-      param_id: @collection.id
-    )
+    @share_message ||= Message.new(share_message_args)
+  end
+
+  def share_message_args
+    {sender: current_user, recipient: User.find_by(email: params[:user]), param_type: 'collection', param_id: @collection.id}
   end
 
   def push_tasks
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -91,12 +91,10 @@ class CollectionsController < ApplicationController
   end
 
   def share
-    flash_message = if @collection.users.include?(share_message.recipient) || Message.exists?(share_message_args)
-                      {alert: t('.duplicate_share')}
-                    elsif share_message.save
+    flash_message = if @collection.users.exclude?(share_message.recipient) && share_message.save
                       {notice: t('.success_notice')}
                     else
-                      {alert: t('common.errors.something_went_wrong')}
+                      {alert: share_message.errors.full_messages.join(', ')}
                     end
     redirect_to collection_path(@collection), flash_message
   end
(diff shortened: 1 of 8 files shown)


config/locales/de/controllers/collections.yml
```diff
@@ -19,6 +19,7 @@ de:
         already_member: Sie sind bereits ein Mitglied dieser Sammlung.
       success_notice: Sie können jetzt an dieser Sammlung zusammenarbeiten.
     share:
+      duplicate_share: Sie haben die Sammlung bereits mit diesem Nutzer geteilt.
```
MrSerth, line 22: I would suggest the following gender-neutral wording, which is in line with the remaining translations:

```suggestion
      duplicate_share: Sie haben die Sammlung bereits mit dieser/diesem Benutzer:in geteilt.
```

... or something different, if we move this to the `errors` (see other comment).

app/controllers/collections_controller.rb
```diff
@@ -91,11 +91,14 @@ def download_all
...
-      redirect_to collection_path(@collection), alert: t('common.errors.something_went_wrong')
-    end
+    flash_message = if @collection.users.include?(share_message.recipient) || Message.exists?(share_message_args)
+                      {alert: t('.duplicate_share')}
+                    elsif share_message.save
+                      {notice: t('.success_notice')}
+                    else
+                      {alert: t('common.errors.something_went_wrong')}
+                    end
+    redirect_to collection_path(@collection), flash_message
```
MrSerth, line 101: I would prefer to have the message validation in the model, where business logic should be placed. With a proper backend validation, we could simply write:

```suggestion
    flash_message = if @collection.users.exclude?(share_message.recipient) && share_message.save
                      {notice: t('.success_notice')}
                    else
                      {alert: share_message.errors.full_messages.join(', ')}
                    end
    redirect_to collection_path(@collection), flash_message
```

I am currently not sure whether we have this kind of validation already, since the following line looks already suitable:

https://github.com/openHPI/codeharbor/blob/5dac5a122eab1e6d66199d07ff349658f1e7c03e/app/models/message.rb#L5

We can modify this line slightly:

```rb
  validates :recipient_id, uniqueness: {scope: %i[param_id param_type], message: :duplicate_share}, if: -> { param_type == 'collection' }
```

then, in the translations, we just use:

```yml
en:
  activerecord:
    errors:
      models:
        message:
          duplicate_share: has already been added to this collection. # TODO @Mathis-Z: Or some other nice text.
```

I feel this would be more Rails-like to handle this error.

MrSerth (commented):
Thank you! The changes work as expected, I just have some stylistic feedback 🙂

app/controllers/collections_controller.rb
MrSerth: Since some tests are failing due to the missing changes for the `Messages` model (and the error translations), I am reopening this thread. I would assume that the changes proposed by me earlier should resolve the test failures.

Mathis-Z: Sorry, I was not done yet.

Mathis-Z: Now I am done :)

app/models/message.rb
```diff
@@ -26,7 +27,7 @@ def text # rubocop:disable Metrics/AbcSize
       when 'group_declined'
         I18n.t('groups.send_deny_access_message.message', user: sender.name, group: Group.find(param_id).name)
       when 'collection'
-        I18n.t('collections.share_message.text', user: sender.name, collection: Collection.find(param_id).title)
+        I18n.t('collections.share_message.text', user: sender.name, collection: Collection.find_by(id: param_id)&.title)
```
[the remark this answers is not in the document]
MrSerth: I've debugged the broken model test and made another proposal. As soon as we have a proper polymorphic association, the test will pass without any manual modifications.

Do you agree to the temporary solution I pushed? If so, I will just proceed merging, so that we can continue in #1679.

MrSerth (approved):
Awesome, thanks for this rework. I think having proper validations in the Message model will pay off. When testing manually, I was able to confirm that everything works as expected 🥳

app/models/message.rb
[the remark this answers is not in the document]
Mathis-Z: Yes, that looks good :+1:

PR #1679, opened 2024-09-23
Fixing dangling references in messages breaking the users inbox

### LINKED ISSUE #1670: Some messages stored in the database are invalid
In CodeHarbor, messages are not only direct messages between users, but are also created for collection invitations, group access requests, etc.

In the current state, some messages might become invalid. This could potentially happen when a user, collection, or group (or other object referenced) gets deleted. As a consequence, the view to read messages is broken, too. We should validate that messages are handled properly when dependent objects get removed or updated, without breaking the view and validation.

I am not sure about a good strategy. Normally, I would prefer to keep messages for non-deleted users (e.g., when a group was deleted), but without showing the corresponding actions (e.g., visiting the group the user was added to). If this approach is feasible, I would prefer it.

To make debugging easier, I will share some examples with @Mathis-Z.

Mathis-Z:
Closes #1670.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -21,7 +21,7 @@ class CollectionsController < ApplicationController
   end
 
   def show
-    @num_of_invites = Message.where(param_type: 'collection', param_id: @collection.id).where.not(recipient_status: 'd').count
+    @num_of_invites = @collection.messages.where(action: :collection_shared).where.not(recipient_status: :deleted).count
   end
 
   def new
@@ -107,7 +107,7 @@ class CollectionsController < ApplicationController
 
     ActiveRecord::Base.transaction(requires_new: true) do
       @collection.users << current_user
-      message = Message.received_by(current_user).find_by(param_type: 'collection', param_id: @collection.id)
+      message = Message.received_by(current_user).find_by(action: :collection_shared, attachment: @collection)
       message.mark_as_deleted(current_user)
       message.save!
       redirect_to @collection, notice: t('.success_notice')
@@ -154,7 +154,7 @@ class CollectionsController < ApplicationController
   end
 
   def share_message_args
-    {sender: current_user, recipient: User.find_by(email: params[:user]), param_type: 'collection', param_id: @collection.id}
+    {sender: current_user, recipient: User.find_by(email: params[:user]), action: :collection_shared, attachment: @collection}
   end
 
   def push_tasks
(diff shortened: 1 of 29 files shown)


MrSerth (commented):
I have a very general, high level feedback on this PR: When looking into the existing Messages model and how the param is calculated, I was pretty ... surprised.

My main problem with the existing implementation is that it mixes the param object (the element the message is about) with a status/action *in one column param_type*. That's not ideal.

I would prefer if we could have one regular, polymorphic element param (or some better name) that defines what this message is about:

```rb
  # models/message.rb
  belongs_to :param, polymorphic: true, optional: true
```

Then, for each object, we would simply list:

```rb
  # models/collection.rb, ...
  has_many :messages, as: :param, dependent: :destroy, inverse_of: :param
```

As a result, Rails will automatically handle the heavy stuff with correct integrity, dependent destroys and so on. The issue is that for the current implementation, we cannot use these "links", since the format stored in the `param_type` column would be different: In Rails, it is simply the model name, e.g. `Collection`, `Group`, ... (module name with capital letter). In the current implementation, we have `collection` (lowercase letter) and `group_accepted`, `group_requested`, `group_declined` (lowercase letter and mixed with status/action).

If time permits, I would hence suggest to have a migration renaming the (content of the) `param_type` column and extracting the status information to a dedicated column (status, action, ....) with a Rails-managed enum:

```rb
  enum :action, {}, prefix: true # See list of potential values below
```

Similarly, the read status (recipient / sender) would also deserve an enum...

```rb
  # Tested already, works 👍 
  enum :sender_status, {deleted: 'd', read: 'r', unread: 'u'}, default: :unread, prefix: true
  enum :recipient_status, {deleted: 'd', read: 'r', unread: 'u'}, default: :unread, prefix: true
```

Having the change for `param_type`, one could simply do `@collection.messages` to get all messages for a given `@collection` (and also append `.where(sender_status: :unread)` or other filters.

The suggested approach is definitely better in the long-term, and preferred to the introduction of the various `group`, `collection`, ... methods.

<hr>

Here's a list of possible `param_type`s I found in the code:

- group_requested
- group_accepted
- group_declined
- collection
- report 
- exercise # not used, but if: should be `task`
- exercise_accepted # not used, but if: should be `task`
- exercise_declined # not used, but if: should be `task`

The `recipient_status` and `sender_status` can be one of:
- d # deleted
- r # read
- u # unread

MrSerth (approved):
Awesome, thank you for figuring out this migration and introducing a real polymorphic association! I am very happy about this change and it makes the message model much easier to maintain and use! 💪

I've included my feedback as a commit, that we can then squash with your first one. Please have a short look, so that we may than finish this together 🙂

PR #1683, opened 2024-09-25
Explicitly remove file attachment if changed after validation failure

MrSerth:
We recently discovered an issue where an old attachment would be incorrectly linked to an updated file after validation errors occurred. Specifically, the reproduction scenario is as follows:

1. Create a new task with a native file attached that passes all validations.
2. Edit this task and introduce a validation error that is unrelated to the file. For example, remove the task's title or language.
3. Select another native file attachment for the already existing TaskFile.
4. Attempt to save the changes
5. As expected, a validation error regarding the change is shown (title, language). Furthermore, the upload field has been cleared, since the file attachment was changed and the newly selected file was not yet saved on the server.
6. Fix the validation error and (without re-selecting the file attachment) save the task again.
7. Now, the task was saved again, despite the "missing" file attachment. However, it was expected that another validation error regarding the missing file attachment is shown. Unfortunately, the existing TaskFile also gets updated with the name of the previously selected file name (new file name, old content).

During our investigation, we identified that the attachment param of TaskParameters#file_params was simply missing, despite `use_attached_file=true`. Consequentially, assigning the params as attributes to the task (and the task file respectively) did not modify the existing `attachment` relationship, so that the previously saved attachment was still "connected". With the commit at hand, we therefore set the `attachment` param to `nil` explicitly, when it is not yet present (indicating that no file was sent by the browser) and the `parent_blob_id` is not present either (indicating that the file attachment got touched in one of the previous attempts to save changes).

<hr>

I didn't add tests on my own, since @florian-str discovered the issue while working on tests for #1103, so that the changes will be covered there.

<hr>

Thanks @nenock for your valuable input on this issue!

### DIFF
--- a/app/controllers/concerns/task_parameters.rb
+++ b/app/controllers/concerns/task_parameters.rb
@@ -20,7 +20,7 @@ module TaskParameters
   def task_params
     params.require(:task).permit(:title, :description, :internal_description, :parent_uuid, :language, :license_id,
       :programming_language_id, :access_level, files_attributes: file_params, tests_attributes: test_params,
-      model_solutions_attributes: model_solution_params, label_names: [])
+      model_solutions_attributes: model_solution_params, label_names: []).tap {|parameters| fix_nested_files_params(parameters) }
   end
 
   def group_tasks_params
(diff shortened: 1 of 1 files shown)


nenock:
Looks good to me. I really like the refactoring since our pair programming!

One thought though: I'd add to the commit message and to the comment in the file, that the behaviour we're prevent here is a unintended renaming of the old file.

MrSerth:
Thanks for your feedback! I've updated the comment and commit message to reflect that fact, too.

nenock (approved):

Issue #1320, opened 2024-03-25
Check task overwriting through ProFormA


MrSerth:
When pushing an existing task through our ProFormA API, one gets ask whether the existing task should be overwritten (given sufficient permissions) or whether a new copy should be created. That's fine and working as expected.

My questions: 
- [x] Is overwriting the task keeping or resetting further data previously set in CodeHarbor that is not sent through ProFormA? For example, I am referring to
  - [x] labels
  - [x] groups
  - [x] the license selected
  - [x] collections
  - [x] comments
  - [x] ratings
  - [x] the visibility level
  - [x] the parent UUID
- [ ] How is CodeOcean behaving there with regards to the CodeOcean-specific data:
  - [ ] tips
  - [ ] tags
  - [ ] LTI parameters
  - [ ] collections
  - [ ] proxy exercises
  - [ ] learner feedback
  - [ ] learner submissions, test runs
  - [ ] RfCs

I was unsure (haven't tested it myself) and would like to know how the systems behave. If any of these data gets lost, I would strongly suggest to keep this "additional" data, only overwriting those "attributes" sent through the API.

kkoehn:
In CodeHarbor we work with `assign_attributes`
```ruby
      @task.assign_attributes(
        user: @user,
        title: @proforma_task.title,
        description:,
        internal_description: @proforma_task.internal_description,
        programming_language:,
        uuid:,
        parent_uuid: @proforma_task.parent_uuid,
        language: @proforma_task.language,
        meta_data: @proforma_task.meta_data,

        submission_restrictions: @proforma_task.submission_restrictions,
        external_resources: @proforma_task.external_resources,
        grading_hints: @proforma_task.grading_hints,

        tests:,
        model_solutions:,
        files: files.values # this line has to be last, because tests and model_solutions have to remove their respective files first
      )

```
Therefore all mentioned relations stay in place and only attributes get changed. `parent_uuid` will get overwritten, because it's part of proforma. `access_level` is not touched and will also stay the same. For some reason `user` is overwritten - I would suggest removing that part, because I don't see a reason to change the ownership of a task because of an import.

MrSerth:
Thanks for checking! That's a valuable finding and seems to be mostly okay; we probably have to change only a few minor things:

**CodeHarbor:**
[...]

PR #1696, opened 2024-10-02
1320: add example tests for file overwriting

kkoehn:
see problem description in [#1320 ](https://github.com/openHPI/codeharbor/issues/1320#issuecomment-2389765834)

### DIFF
--- a/app/services/proforma_service/convert_proforma_task_to_task.rb
+++ b/app/services/proforma_service/convert_proforma_task_to_task.rb
@@ -19,6 +19,7 @@ module ProformaService
     private
 
     def import_task
+      upsert_files @proforma_task, @task
       @task.assign_attributes(
         user:,
         title: @proforma_task.title,
@@ -37,7 +38,6 @@ module ProformaService
         tests:,
         model_solutions:
       )
-      upsert_files @proforma_task, @task
       delete_removed_files
     end
 
--- a/spec/services/proforma_service/convert_proforma_task_to_task_spec.rb
+++ b/spec/services/proforma_service/convert_proforma_task_to_task_spec.rb
@@ -579,7 +579,7 @@ RSpec.describe ProformaService::ConvertProformaTaskToTask do
 
     context 'when proforma_task has been exported from task' do
       let(:proforma_task) { ProformaService::ConvertTaskToProformaTask.call(task:) }
-      let(:task) { create(:task, files: task_files, tests:, model_solutions:, title: 'title') }
+      let!(:task) { create(:task, files: task_files, tests:, model_solutions:, title: 'title') }
       let(:task_files) { build_list(:task_file, 1, :exportable, internal_description: 'original task file') }
       let(:tests) { build_list(:test, 1, files: test_files) }
       let(:test_files) { build_list(:task_file, 1, :exportable, internal_description: 'original test file') }
@@ -596,11 +596,11 @@ RSpec.describe ProformaService::ConvertProformaTaskToTask do
           ))),
           model_solutions: have(1).item.and(include(have_attributes(
             id: model_solutions.first.id,
-            files: include(have_attributes(id: model_solutions.first.files.first.id))
+            files: have(1).item.and(include(have_attributes(id: model_solutions.first.files.first.id)))
           ))),
           tests: have(1).item.and(include(have_attributes(
             id: tests.first.id,
-            files: include(have_attributes(id: tests.first.files.first.id))
+            files: have(1).item.and(include(have_attributes(id: tests.first.files.first.id)))
           )))
         )
       end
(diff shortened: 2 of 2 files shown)


MrSerth:
I've spent further time debugging and looking for a solution. Finally, I wasn't able to do so with the current architecture and, hence, need to suggest a major architectural change:

Could you try to perform the necessary changes directly by updating and saving the changes (either with the `@task.update` method or by first assigning new values and then using `@task.save`)? 

Of course, this means the service will return an altered object, persisted object already and not a modified version (that needs to be save). However, the big advantage is that we can more fine granular control which object gets saved how and don't need to wrap around the Rails magic. 

Then, you should wrap the `execute` block into a database transaction. This allows you to perform the changes (i.e., on files) and they "look" persisted in this transaction even if they are not. You may then also use something like `@task.reload` to get an updated version of the record and move over to the next parts (i.e., tests, model solutions). If you only want to reload associations, you can do so with `@task.association(:tests).reload` (to force, use `@task.association(:tests).reload(true)`).

I am rather confident that we will be able to navigate around the current issue with this approach, even if it changes the API. Hence, I would suggest to follow with this approach.

MrSerth (commented):
In an effort to ease domain-specific review, I have executed the code locally and performed my well-known test with the previously-shared set of new and old exercise files (for ID 845). It worked flawlessly and as expected, thank you. 

Still, I identified an unnecessary N+1 query caused by the `autosave` option. It might no longer be necessary (at least in my tests, everything worked without the option, too). Hence, we might be able to drop it:

```diff
Subject: [PATCH] Remove autosave causing unnecessary N+1 queries
---
Index: app/models/task_file.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/models/task_file.rb b/app/models/task_file.rb
--- a/app/models/task_file.rb	(revision cc487bdb571f5ce4cda4cf42132a76f85e72ef11)
+++ b/app/models/task_file.rb	(date 1737201720489)
@@ -4,7 +4,7 @@
   include ParentValidation
   include TransferValues
 
-  belongs_to :fileable, polymorphic: true, autosave: true, inverse_of: :files
+  belongs_to :fileable, polymorphic: true, inverse_of: :files
   belongs_to :parent, class_name: 'TaskFile', optional: true
   has_one_attached :attachment
   validates :name, presence: true
Index: app/models/model_solution.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/models/model_solution.rb b/app/models/model_solution.rb
--- a/app/models/model_solution.rb	(revision cc487bdb571f5ce4cda4cf42132a76f85e72ef11)
+++ b/app/models/model_solution.rb	(date 1737201720486)
@@ -5,7 +5,7 @@
   include TransferValues
   include ParentValidation
 
-  belongs_to :task, autosave: true, inverse_of: :model_solutions
+  belongs_to :task, inverse_of: :model_solutions
   belongs_to :parent, class_name: 'ModelSolution', optional: true
   has_many :files, as: :fileable, class_name: 'TaskFile', dependent: :destroy
   accepts_nested_attributes_for :files, allow_destroy: true
Index: app/models/test.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/models/test.rb b/app/models/test.rb
--- a/app/models/test.rb	(revision cc487bdb571f5ce4cda4cf42132a76f85e72ef11)
+++ b/app/models/test.rb	(date 1737201720494)
@@ -5,7 +5,7 @@
   include TransferValues
   include ParentValidation
 
-  belongs_to :task, autosave: true, inverse_of: :tests
+  belongs_to :task, inverse_of: :tests
   belongs_to :testing_framework, optional: true
   belongs_to :parent, class_name: 'Test', optional: true
   validates :title, presence: true
Index: app/models/concerns/file_concern.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/models/concerns/file_concern.rb b/app/models/concerns/file_concern.rb
--- a/app/models/concerns/file_concern.rb	(revision cc487bdb571f5ce4cda4cf42132a76f85e72ef11)
+++ b/app/models/concerns/file_concern.rb	(date 1737201720483)
@@ -4,7 +4,7 @@
   extend ActiveSupport::Concern
 
   included do
-    has_many :files, as: :fileable, class_name: 'TaskFile', dependent: :destroy, autosave: true, inverse_of: :fileable
+    has_many :files, as: :fileable, class_name: 'TaskFile', dependent: :destroy, inverse_of: :fileable
     accepts_nested_attributes_for :files, allow_destroy: true
   end
 ```

Please keep in mind that I haven't reviewed the code 😇.

kkoehn:
Thanks for the testing! I agree, the that we should be able to remove the autosaving.

MrSerth (approved):
Awesome, thank you! ☃️

With these changes applied, I do not have any further comments and hence can approve the PR from my point of view. However, I want to reiterate that I have not performed a code-level review.

PR #1846, opened 2025-01-29
1680: Improve error when exporting through an account_link

kkoehn:
Added a more descriptive error message for the presumably most common error.

### DIFF
--- a/app/services/task_service/push_external.rb
+++ b/app/services/task_service/push_external.rb
@@ -9,20 +9,23 @@ class TaskService
     end
 
     def execute
-      body = @zip.string
-      begin
-        response = self.class.connection.post(@account_link.push_url) {|request| request_parameters(request, body) }
-        if response.success?
-          nil
-        else
-          response.status == 401 ? I18n.t('tasks.export_external_confirm.not_authorized', account_link: @account_link.name) : response.body
-        end
-      rescue StandardError => e
-        e
-      end
+      response = self.class.connection.post(@account_link.push_url) {|request| request_parameters(request, @zip.string) }
+      return nil if response.success?
+      return I18n.t('tasks.export_external_confirm.not_authorized', account_link: @account_link.name) if response.status == 401
+
+      handle_error(message: response.body)
+    rescue Faraday::ServerError => e
+      handle_error(error: e, message: I18n.t('tasks.export_external_confirm.server_error', account_link: @account_link.name))
+    rescue StandardError => e
+      handle_error(error: e)
     end
 
     private
+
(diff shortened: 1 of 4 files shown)


MrSerth:
Thanks! Similar to CodeOcean, there is still some issue with an error case.

Screenshot 2025-02-05 at 10 14 33

We could also set the timeout to five seconds:

```diff
Subject: [PATCH] Add timeout for TaskService::PushExternal
---
Index: app/services/task_service/push_external.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/services/task_service/push_external.rb b/app/services/task_service/push_external.rb
--- a/app/services/task_service/push_external.rb	(revision 1818ac35bd8a8d6a0402961e2ded0513e5159cee)
+++ b/app/services/task_service/push_external.rb	(date 1738746760866)
@@ -34,6 +34,9 @@
 
     def connection
       Faraday.new(url: @account_link.push_url) do |faraday|
+        faraday.options[:open_timeout] = 5
+        faraday.options[:timeout] = 5
+
         faraday.adapter Faraday.default_adapter
       end
     end
```

app/services/task_service/push_external.rb
```diff
@@ -32,6 +39,8 @@ def request_parameters(request, body)
     def connection
       Faraday.new(url: @account_link.push_url) do |faraday|
         faraday.adapter Faraday.default_adapter
+        faraday.options[:open_timeout] = 5
+        faraday.options[:timeout] = 5
       end
```
MrSerth, line 44: Given the similar lines in `check_external.rb`, I would prefer the same order and alignment of these three options:

https://github.com/openHPI/codeharbor/blob/dc64015490a1ebaa364688e484a0ef8ae7cb61c7/app/services/task_service/check_external.rb#L43-L48

MrSerth (commented):
Thanks for the further adjustments. I can confirm that only escaped strings are returned. 👏

💁 Please keep in mind: I have not done a regular code review.

app/services/task_service/push_external.rb
kkoehn: You are right, that makes sense.
Thanks for the (not)-review and sanity checks!

MrSerth: You're very welcome! Thanks for your great ongoing work, I am delighted to see these enhancements landing soon 👍

MrSerth (commented):
Just a few suggestions for the translations 🙂

arkirchner (approved):

PR #1852, opened 2025-02-05
1247 Replace Uniquenes validation for xml_id in ModelSolution and Test

### LINKED ISSUE #1247: Broken XML_ID uniqueness validation for tests
There is a validation for tests to make sure there cannot be two tests with the same XML_ID for the same task. However this validation appears to be broken because I can just create tasks with tests that have the same XML_ID.
I guess Rails performs the validation of both tests before they are saved to the database and each validation check does not find the other test in the database so they just pass and get written to database successfully.
We should check if there are similar issues in other models.

EDIT: You can create a task like that but you cannot edit it afterwards because then the tests are in the database and the validation fails.

kkoehn:
fixes #1247


arkirchner (approved):
Thank you!

PR #1868, opened 2025-02-11
Perform extended status group validation for NBP sign-up

MrSerth:
For all users signing up through the NBP wallet, we want to validate their role. Only those registrations providing a specific role (learner / educator) should be allowed. Any other registrations should not be affected by this change.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -47,6 +47,11 @@ class User < ApplicationRecord
 
   # `Other` is a catch-all for any status that doesn't fit into the other categories and should be *last*.
   enum :status_group, {unknown: 0, learner: 2, educator: 3, other: 1}, default: :unknown, prefix: true
+  # When a user is created through the NBP wallet connection, we want to ensure a valid status group.
+  # We need to check with a string, because any symbol or integer otherwise used is automatically converted.
+  validates :status_group, inclusion: {in: %w[learner educator], message: :unrecognized_role}, on: :create, if: lambda {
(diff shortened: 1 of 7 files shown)


MrSerth:
Sorry for the delayed PR, I got derailed by a local bug triggered when running all specs before pushing: `https://github.com/SeleniumHQ/selenium/issues/15273`.

app/models/user.rb
```diff
@@ -47,6 +47,10 @@ class User < ApplicationRecord
 
   # `Other` is a catch-all for any status that doesn't fit into the other categories and should be *last*.
   enum :status_group, {unknown: 0, learner: 2, educator: 3, other: 1}, default: :unknown, prefix: true
+  # When a user is created through the NBP wallet connection, we want to ensure a valid status group.
+  validates :status_group, inclusion: {in: %w[learner educator], message: :unrecognized_role}, on: :create, if: lambda {
```
nenock, line 51: `Relationship#userdata[:status_group]` returns the two valid values as symbols.
```suggestion
  validates :status_group, inclusion: {in: %s[learner educator], message: :unrecognized_role}, on: :create, if: lambda {
```

nenock (approved):
Looks good!

app/models/user.rb
MrSerth: The correct syntax would be

```suggestion
  validates :status_group, inclusion: {in: %s[learner educator], message: :unrecognized_role}, on: :create, if: lambda {
```

However, an array of symbols doesn't work, since a symbol or integer passed is automatically converted. I extended the specs and kept the original behavior.

PR #2092, opened 2025-07-30
chore: Enable Rubocop for migrations

arkirchner:
Rubocop gives valuable information about issues in migration. New migrations should be checked by the linter.


MrSerth (approved):
Nice, thank you!

Issue #2104, opened 2025-08-08
Allow object storage to work with S3-compatible storage solutions


arkirchner:
ActiveStorge blobs can currently only be saved in a local storage (on disk). We want to support S3-comparable storage to support containerization.

 

Steps:
- [ ] Add optional configuration to use S3 as storage
- [ ] Mirror local blobs to S3 storage
- [ ] Add remaining local data to S3
- [ ] Use S3 as default provider
- [ ] Remove mirroring and only use S3
- [ ] Remove local data

PR #2128, opened 2025-08-15
Allow customization of ActiveStorage storage

arkirchner:
In the production environment, an alternative S3-compatible storage provider can be configured.

Related to #2104


arkirchner:
[memo] Do not merge until the deployment pipeline with the new config is ready.

PR #2057, opened 2025-07-01
add mailer for collection invitations

kkoehn:

PR #2133, opened 2025-08-25
Fix collection (mailer) bugs

kkoehn:
fix bugs introduced in #2057 and ecd1d7bac53c674be7e484146618bf95229b78e1


MrSerth (approved):
Awesome, thank you -- looks good! I just added some space before the URL, and think we can merge this. Also, thanks for catching the form bug; I've also fixed the `id` attribute.

